Issue #13, opened 2017-02-08
tests: prepare local development test cluster


diegodelemos:
In order to have a local working cluster the following things need to be done:
- [x] Make current configuration in https://github.com/reanahub/reana-resources-k8s abstract, so local/production configuration can be automatically generated on demand. (https://github.com/reanahub/reana-resources-k8s/issues/17)
- [x] Deploy current system using [minikube](https://github.com/kubernetes/minikube/releases).
- [x] ~~Create a CephFS cluster.~~ Use host directories to mock distributed file system.
- [x] Add documentation to deploy and test locally. (https://github.com/reanahub/reana/issues/4)
- [x] Create `hello world` example. (https://github.com/reanahub/reana-demo-helloworld/issues/2)

tiborsimko:
Perhaps [Vagrant sync folders](https://www.vagrantup.com/docs/synced-folders/basic_usage.html) can be used? I had a good experience with shared folders in a multi-VM environment for light cache sharing (some R/W, but mostly R/O accesses by VMs). I did not test any heavy R/W race conditions at the time though...

lukasheinrich:
perhaps for super local environment, just a mounted host directory could be used?

diegodelemos:
If we suppose that the local environment is always going to be run in only one VM we could. This is true for the current minikube version, which creates the full cluster in one VM, but we do not know what will happen in the near future or even if we will want more than one VM for some reason.

Shall we stay simple for now and try host directories and let see what happens later?

tiborsimko:
> Shall we stay simple for now and try host directories and let see what happens later?

Sounds good! 👍

tiborsimko:
The local development environment is now available. The docs have been added. Closing the issue.

anton-khodak:
@tiborsimko can you please point to the appropriate docs?

Issue #4, opened 2017-01-19
docs: amend installation


diegodelemos:
Amend current docs to follow the actual naming convention and remove Fibonacci example.

tiborsimko:
BTW note also the "focilo" leftover in the documentation.

tiborsimko:
`closes` or rather `addresses` in the commit log message? (because of one added `FIXME`...)

PR #15, opened 2017-03-22
installation: add `minikube` installation

diegodelemos:
* Adds local installation documentation. (addresses #4)

* Adds `minikube` specific installation option to REANA configuration
  script. (addresses #13)

Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>


docs/installation.rst
```diff
@@ -0,0 +1,66 @@
...
+   yadage-atlas-worker-2909073811-t9qv3   1/1       Running            0          1m
+   yadage-cms-worker-209120003-js5cv      1/1       Running            0          1m
+   yadage-lhcb-worker-4061719987-6gpbn    1/1       Running            0          1m
+   zeromq-msg-proxy-1617754619-68p7v      1/1       Running            0          1m
+
+Now we will be able to check whether the nodes are actually working. Firstly, we get the VM's ip:
+
+.. code-block:: console
+
+   $ MINIKUBE_IP=$(minikube ip)
```
jirikuncar: Did you mean `$ export MINIKUBE_ID=$(minikube ip)`?

diegodelemos: yes... Thanks!

diegodelemos:
> There are hardcoded strings of "Installing minikube for OSX ..." and "Installing latest kubectl for OSX ...". There should be Linux alternatives for them.

Done

> There are two instances of typo CONTOLLER_PORT which should be CONTROLLER_PORT since instructions guide you to open "http://$MINIKUBE_IP:$WORKFLOW_CONTROLLER_PORT" which won't work.

👍 

> open "http://$WORKFLOW_CONTROLLER_IP:$WORKFLOW_CONTROLLER_PORT" gave a 404 even with corrected port (port is not written at all because of the CONTOLLER_PORT typo). Despite this, the cluster might be running fine since both of the other test links worked.

This is foreseen because we just have a `yadage` endpoint which only accepts POST requests. I have opened a ticket and assigned you https://github.com/reanahub/reana-workflow-controller/issues/20 it is quick change, just add an empty `workflows` endpoint!

> Could't install because my Linux-system didn't pass the if ! "darwin linux" | grep -F -q -w -i "$(uname -s)" test and the script prompted that "Your OS is not supported." In order to proceed with the installation I had to comment out the test.

I might be wrong, but I think that this is not currently happening, there was an `echo` missing in the tests statement but not anymore.

> Maybe it should be added that for now you should install with pip install -e .[all] before running reana install-minikube. It might not be clear for all who check out the repo on their own.

I am not sure about this lets chat IRL

> Also same thing that is said in Kubernetes cluster part about Python virtual environments, should probably be said in Minikube part as well: $ mkvirtualenv reana-resources-k8s

Not sure about this neither - IRL

> I had a non-blocking (cluster seemed to work fine) problem with ZeroMQ where $ kubectl get pods reported that zeromq-pod is not running: zeromq-msg-proxy-1617754619-qjvc7 0/1 CrashLoopBackOff 5 4m

Work in progress. I am fixing https://github.com/reanahub/reana-workflow-monitor/pull/13

> Maybe it should be added that what you should expect when opening the links. Either text or small screenshots. E.g. for link "http://$JOB_CONTROLLER_IP:$JOB_CONTROLLER_PORT/jobs" theinstructions would say something like "Displays JSON with a key 'jobs' which is empty since no jobs are running on the minikube" and for link "http://$WORKFLOW_MONITOR_IP:$WORKFLOW_MONITOR_PORT/helloworld" that "Displays yadage-monitor".

Totally agree, thanks for the idea 👍 

p.s. next time could you please comment in line? it is a bit messy to go from the comment to the code back and forth! thanks 🤙

hjhsalo:
Sorry, these comments were unnecessary. I didn't notice that they are already in the instructions. :man_facepalming: 

> * Maybe it should be added that for now you should install with pip install -e .[all] before running reana install-minikube. It might not be clear for all who check out the repo on their own.

> * Also same thing that is said in Kubernetes cluster part about Python virtual environments, should probably be said in Minikube part as well: $ mkvirtualenv reana-resources-k8s

hjhsalo:
Note that I just added a [PR21](https://github.com/reanahub/reana-workflow-controller/pull/21) to reana-workflow-controller that addresses comments about /workflows -endpoint.

docs/installation.rst
```diff
@@ -0,0 +1,188 @@
...
+
+And finally, we open the test URLs:
+
+.. code-block:: console
+
+   $ curl "http://$MINIKUBE_IP:$WORKFLOW_CONTROLLER_PORT/workflows"
+   {
+     "workflows": {}
+   }
+   $ open "http://$MINIKUBE_IP:$WORKFLOW_MONITOR_PORT/helloworld"
```
tiborsimko: `open ...` is macOS specific.  Something like `xdg-open ...` works on GNU/Linux. If it works on macOS too, then please switch. Otherwise let us simply use `firefox ...` as people will know what that means...

docs/installation.rst
```diff
@@ -0,0 +1,188 @@
...
+
+.. code-block:: console
+
+   $ export MINIKUBE_IP=$(minikube ip)
+
+Secondly, we get the current instances ports:
+
+.. code-block:: console
+
+   $ export WORKFLOW_CONTROLLER_PORT=$(kubectl describe service workflow-controller | grep 'NodePort:' | cut -f 4 | sed -e "s@/TCP@@")
```
tiborsimko: For these and other environment variables, it would be good to use the common prefix `REANA_...` so that we don't create a mess in people's environments...

docs/installation.rst
```diff
@@ -0,0 +1,188 @@
...
+   $ git clone https://github.com/reanahub/reana-resources-k8s
+   $ cd reana-resources-k8s
+   $ pip install -e .[all]
+
+Now we should configure ``reana-resources-k8s/templates/config.yaml`` as it fits the better, and once we are done, we should generate the manifests and create the nodes
+
+.. code-block:: console
+
+   $ reana-resources-k8s build-manifests
+   $ kubectl create -Rf configuration-manifests
```
tiborsimko: `-rf` should be enough, let's keep it lowercase.

docs/installation.rst
```diff
@@ -0,0 +1,188 @@
...
+
+Clean manifests if they are not needed anymore
+
+.. code-block:: console
+
+   $ rm -Rf configuration-manifests
```
tiborsimko: `-rf`

scripts/reana
```diff
@@ -0,0 +1,112 @@
...
+    echo 'Building REANA configuration manifests ...'
+    reana-resources-k8s build-manifests
+
+    # Set current cluster secret for `reana-job-controller` component
+    current_cluster_secret=$(kubectl get secrets | grep 'service-account-token' | cut -d ' ' -f 1)
+    sed -i'' -e "s/default-token-8p260/$current_cluster_secret/" configuration-manifests/deployments/job-controller.yaml
+
+    # Deploy system
+    echo 'Deploying REANA on minikube ...'
+    kubectl create -Rf configuration-manifests
```
tiborsimko: `-rf`

docs/installation.rst
```diff
@@ -0,0 +1,188 @@
...
+   $ cd reana
+   $ pip install -e .[all]
+
+Now, for instantiating a local REANA `minikube` based cluster we should run
+
+.. code-block:: console
+
+   $ reana install-minikube
+
+which will do the following:
```
tiborsimko: It would be good to decouple the steps that install kubernetes/minikube itself, which many people would prefer to do outside of REANA using their packaging systems, and install the rest of the REANA component ecosystem.

The getting started overview for people would then look like:

```
$ pip install reana
$ reana install-backend minikube # or skip this step if you have it already
$ reana configure-backend minikube # or other clouds later... just an example
$ reana configure-container docker # or singularity... just an example
$ reana pull-components 
$ reana build-manifests
$ reana initialise-backend
$ reana run-hello-world
...
```

(hypothetical example, let's muse IRL)

docs/installation.rst
diegodelemos: For some reason `kubectl` uses only the capital `r` for the recursive option

scripts/reana
```diff
@@ -0,0 +1,112 @@
...
+    echo "  install-minikube    Deploys local minikube based REANA cluster"
+    echo "  -h                  Help"
+}
+
+if [ "$1" != "" ]
+then
+    case $1 in
+        install-minikube ) install_local_minikube
+                           ;;
+        clean-minikube)    minikube delete
```
hjhsalo: I couldn't get minikube cluster to start after the first run without running ``` minikube delete ``` first.
I'm not sure how many times I started it before using ``` reana install-minikube ``` or did I mess around with Virtualbox, but I would bet that some of the minikube VM's were either in corrupted state or not present.
Maybe something about running ``` reana clean-minikube ``` should be added to the documentation? Maybe just a side note:

_"NOTE: In case you run into error messages from Kubernetes related to local Kubernetes cluster (aka minikube), you can try to run ``` reana clean-minikube ``` to delete VMs created by previous minikube installation / initialization"_

BTW, the error I got was:
``` $ reana install-minikube
Starting local Kubernetes cluster...
Starting VM...
E0403 20:06:00.759374   16086 start.go:113] Error starting host: Error getting state for host: machine does not exist.
```

diegodelemos: Since our discussion last week, we decided to leave backend specific management up to the user. Just a minimal command for installing and running `Kubernetes` is provided. However, if it doesn't work (he have to investigate your problem to see if it is easy to fix) I would suggest to remove it from the CLI and stick to REANA specific operations.

diegodelemos:
As I specify in the code, it is not ready for non local installation (it is somehow, but hacks such as changing `reana-resources-k8s` configuration file by hand is needed 🙅 ) but since we are addressing https://github.com/reanahub/reana/issues/13, local development and quick tests by early adopters, we can leave it for later.

Specially in this case, differing from local to production could be easily specified by running `pip install reana[minikube]` or `pip install reana[kubernetes]` so a Python CLI ([click](http://click.pocoo.org/5/)) would be easier to implement (as discussed IRL with @tiborsimko ). On the other hand, I am not sure if now is the best moment to address it, probably using `sh` CLI for now and focusing in other work packages would be more useful. OFC we have to discuss IRL on Friday

tiborsimko:
Thanks, let's merge now and let's improve CLI language and docs later.

PR #19, opened 2017-04-27
cli: add demo functionality to Reana cli script

hjhsalo:
* add cli command to run reana-demo-helloworld example easily
  `reana run helloworld`

Signed-off-by: Harri Hirvonsalo <harri.hirvonsalo@cern.ch>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,7 @@ target/
 
 # Vagrant
 .vagrant
+
+# IntelliJ project files
+.idea
+*.iml
--- a/scripts/reana
+++ b/scripts/reana
@@ -26,6 +26,11 @@ reana-workflow-monitor
 reana-message-broker
 reana-workflow-engine-yadage"
 
+experiment_list="alice
+lhcb
+atlas
+cms"
+
 install_initialize_minikube () {
     # Check whether the OS is supported
     if ! echo "darwin linux" | grep -F -q -w -i "$(uname -s)"
@@ -93,7 +98,7 @@ deploy () {
 init () {
     # FIXME more logic should be added in order to avoid pulling images locally
     # when deployment is not local.
-    # e.g. if `producton` desired, `pull` should be skipped and
+    # e.g. if `production` desired, `pull` should be skipped and
     # reana-resources-k8s templates/config.yaml should be updated accordingly.
     pull
     prepare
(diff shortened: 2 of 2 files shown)


hjhsalo:
For now, hardcoded to run only helloworld-workflow.

Doesn't follow the run-syntax described in reanahub/reana-demo-helloworld#2 as I feel that this is something we will have to discuss more before implementing it; what is the actual input of `reana run` (file? folder?) and how to submit the workflow to workflow-controller, etc.

scripts/reana
```diff
@@ -135,6 +222,7 @@ usage () {
     echo "  install-minikube    Installs and starts local minikube instance"
     echo "  init                Initialize REANA system"
     echo "  down                Delete REANA components"
+    echo "  demo helloworld     Runs a helloworld demo"
```
tiborsimko: `run helloworld` so that we are ready for the future `run myanalysis` without having to change the CLI.

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+
+                if [ "$flag_devel" != "" ]; then
+                    # Start tailing of job-controller logs
+                    #kubectl logs -f $(kubectl get pods | grep -Po 'job-controller[^ ]+')
+
+                    # Start tailing of yadage-atlas-worker logs
+                    #kubectl logs -f $(kubectl get pods | grep -Po 'yadage-atlas-worker[^ ]+')
+
+                    # Open workflow-monitor web-interface
+                    firefox "http://$controller_url/$workflow_id"
```
tiborsimko: It would be good to take commands like `firefox` out of the script (and use them only in docs).

.gitignore
```diff
@@ -78,3 +78,8 @@ target/
 
 # Vagrant
 .vagrant
+
+# IntelliJ project files
+.idea
+*.iml
+out
```
tiborsimko: `out` needed here?

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
 
+demo () {
+    # Used to run REANA demo workflows.
+    # For now, hardcoded to run only helloworld-workflow.
+    demonstration=$2
+
+    # The code to check for flags is from http://stackoverflow.com/q/2875424
+    # Not sure if this is a good way or not.
+    flag_verbose=$(echo "$*" | grep -e "--verbose" -o)
+    flag_devel=$(echo "$*" | grep -e "--devel" -o)
```
tiborsimko: Maybe simple quotes to prevent command execution:

```
$ echo "$(date)"
Thu 27 Apr 18:40:02 CEST 2017
$ echo '$(date)'
$(date)
```

or rely on the flag being present last in $3 some place, or use environment variable...

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+
+            # Get URL of reana-workflow-controller
+            controller_url=$(reana get reana-workflow-monitor)
+
+            # Submit a helloworld-workflow job to workflow-controller
+            workflow_id=$(curl --fail --silent \
+                            -H "Content-Type: application/json" \
+                            -X POST -d \
+                            "{\"experiment\":\"atlas\",
+                              \"toplevel\":\"github:diegodelemos/reana-demo-helloworld@2-add-helloworld-workflow\",
```
tiborsimko: diegodelemos not needed anymore?

.gitignore
hjhsalo: Have to check if this is copy-paste error.

scripts/reana
hjhsalo: @diegodelemos ?

.gitignore
hjhsalo: Legit but not needed. I'll remove it.
IntelliJ -specific folder for [generated .class files and artifacts](http://stackoverflow.com/a/28231127)

scripts/reana
diegodelemos: see reanahub/reana-demo-helloworld#4. Once it is merged `\"toplevel\":\"github:diegodelemos/reana-demo-helloworld@2-add-helloworld-workflow\"` should be changed for `\"toplevel\":\"github:reanahub/reana-demo-helloworld\"`.

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+            # https://unix.stackexchange.com/a/204790
+            # Note that checking with '$?' might fail on some corner cases
+            # AND when authentication is involved (response codes 401 and 407)
+            submit_status=$?
+
+            if [ "$submit_status" = "0" ]; then
+
+                if [ "$flag_verbose" != "" ]; then
+                    echo "[INFO] Open http://$controller_url/$workflow_id to see visual representation of your workflow and it's progress."
+                    echo "[INFO] You can 'cat' or 'tail' /reana/atlas/$workflow_id/yadage/helloworld/greetings.txt -file"
```
diegodelemos: Because of reanahub/reana-workflow-engine-yadage#20 only ATLAS workflows are currently working, however, it would be nice to use something line `$(experiment)` instead of the string `atlas`. Since in `reana-resources-X` we are following the pattern `/reana/<experiment>` (see [here](https://github.com/reanahub/reana-resources-k8s/blob/master/reana_resources_k8s/templates/config.yaml#L65)) it would match perfectly

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+                    echo "       from shared storage of REANA cluster to see how helloworld workflow progresses."
+                    echo "[INFO] (e.g. minikube ssh \"tail -c+1 -F /reana/atlas/$workflow_id/yadage/helloworld/greetings.txt\")"
```
diegodelemos: `$(experiment)` rather than `atlas`

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+                fi
+
+                echo ""
+                echo "Workflow_id: $workflow_id"
+                echo "Monitor: http://$controller_url/$workflow_id"
+                echo "Output:  /reana/atlas/$workflow_id/yadage/helloworld/greetings.txt"
```
diegodelemos: same about `atlas`

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+                    # Start tailing of yadage-atlas-worker logs
+                    #kubectl logs -f $(kubectl get pods | grep -Po 'yadage-atlas-worker[^ ]+')
+
+                    # Open workflow-monitor web-interface
+                    firefox "http://$controller_url/$workflow_id"
+
+                    # Tail output-file of the workflow
+                    echo "[INFO] Tailing...hit Ctrl+C when done."
+                    echo ""
+                    minikube ssh "tail -c+1 -F /reana/atlas/$workflow_id/yadage/helloworld/greetings.txt"
```
diegodelemos: same about `atlas`

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+                            -H "Content-Type: application/json" \
+                            -X POST -d \
+                            "{\"experiment\":\"atlas\",
+                              \"toplevel\":\"github:diegodelemos/reana-demo-helloworld@2-add-helloworld-workflow\",
+                              \"workflow\": \"hello.yaml\",
+                              \"nparallel\": \"10\",
+                              \"preset_pars\": {\"name\": \"J.Doe\", \"delay\": \"2\"}
+                            }" \
+                            http://"$(reana get reana-workflow-controller)"/yadage \
+                            | grep -oP '(?<="workflow_id": ")[^"]*')
```
diegodelemos: `-P` option is not available in BSD, so something like `grep -oE '"workflow_id"\:[[:space:]]*\"[^\"]*\"'| cut -d ":" -f 2 | tr -d "\""` would support both GNU and BSD

scripts/reana
```diff
@@ -104,6 +106,91 @@ down () {
...
+                fi
+
+                echo ""
+                echo "Workflow_id: $workflow_id"
+                echo "Monitor: http://$controller_url/$workflow_id"
+                echo "Output:  /reana/atlas/$workflow_id/yadage/helloworld/greetings.txt"
+
+                if [ "$flag_devel" != "" ]; then
+                    # Start tailing of job-controller logs
+                    #kubectl logs -f $(kubectl get pods | grep -Po 'job-controller[^ ]+')
```
diegodelemos: I see more practical to have some developer documentation where we explain which logs would be nice to follow as well as which endpoints should be opened. If you use `i3`, `tmux` or something else  you have your own setup for doing this kind of things

scripts/reana
hjhsalo: If I cahnge this to `flag_devel=$(echo $* | grep -e "--devel" -o)`, shellcheck complains:
* SC2048: Use "$@" (with quotes) to prevent whitespace problems.
* SC2086: Double quote to prevent globbing and word splitting.

With `flag_devel=$(echo $@ | grep -e "--devel" -o)` shellcheck complains:
* SC2068: Double quote array expansions, otherwise they're like $* and break on spaces.

If I use 'flag_devel=$(echo "$@" | grep -e "--devel" -o)' shellcheck doesn't complain.

I understand your point though.

I don't know enough about shell scripting to be able to say if any of these is good.

Should we change from `sh` to `bash` to use `if [[ "$@" != *"--devel"* ]]; then` if that would be a better alternative?

scripts/reana
hjhsalo: Changed to `open` and `xdg-open`.

scripts/reana
```diff
@@ -104,6 +106,134 @@ down () {
...
+            echo "[INFO] (e.g. minikube ssh \"tail -c+1 -F /reana/$experiment/$workflow_id/yadage/helloworld/greetings.txt\")"
+        fi
+
+        echo ""
+        echo "Workflow_id: $workflow_id"
+        echo "Monitor: http://$controller_url/$workflow_id"
+        echo "Output:  /reana/$experiment/$workflow_id/yadage/helloworld/greetings.txt"
+        echo ""
+
+        if [ "$flag_devel" != "" ]; then
```
hjhsalo: Should I get rid of the whole --devel part of the code?

scripts/reana
```diff
@@ -104,6 +106,134 @@ down () {
...
+        experiment=$4
+    else
+        experiment=${REANA_EXPERIMENT-}
+    fi
+
+    workflow=$2
+
+    # The code to check for flags is from http://stackoverflow.com/q/2875424
+    # Not sure if this is a good way (POSIX compliant?) or not.
+    flag_verbose=$(echo "$@" | grep -e "--verbose" -o)
```
hjhsalo: If I change this to `flag_devel=$(echo $* | grep -e "--devel" -o)`, shellcheck complains:
* SC2048: Use "$@" (with quotes) to prevent whitespace problems.
* SC2086: Double quote to prevent globbing and word splitting.

With `flag_devel=$(echo $@ | grep -e "--devel" -o)` shellcheck complains:
* SC2068: Double quote array expansions, otherwise they're like $* and break on spaces.

If I use `flag_devel=$(echo "$@" | grep -e "--devel" -o)` shellcheck doesn't complain.

I understand your point though.

I don't know enough about shell scripting to be able to say if any of these is good.

Should we change from `sh` to `bash` to use `if [[ "$@" != *"--devel"* ]]; then` if that would be a better alternative?

scripts/reana
hjhsalo: Didn't update yet. I will look into this next week.

scripts/reana
```diff
@@ -26,6 +26,8 @@ reana-workflow-monitor
 reana-message-broker
 reana-workflow-engine-yadage"
 
+experiment_list=$(find ./configuration-manifests/namespaces/ -name \*.yaml -printf '%f\n' |  sed -e 's/\.yaml$//')
```
hjhsalo: Should I just use hardcoded experiment list?

diegodelemos: It should be ok, in the future we can grab it from configuration

scripts/reana
diegodelemos: `flag_devel=$(echo "$@" | grep -qe "--\<verbose\>")`
`flag_devel=$(echo "$@" | grep -qe "--\<devel\>")` 
 might be better because current one would allow providing `--dev` or `--verbo`. So we could stick to `sh` which is more standard

scripts/reana
diegodelemos: which is what you actually do, sorry about that 🙃

scripts/reana
```diff
@@ -104,6 +106,134 @@ down () {
...
+run () {
+    # Used to start workflows on REANA cluster
+    # Currently hardcoded to run REANA helloworld demo workflow (L210).
+    # (L204) Experiment hardcoded to 'atlas' for now because of
+    # reanahub/reana-workflow-engine-yadage/issues/20
+
+    if [ "$3" = "-e" ]; then
+        experiment=$4
+    else
+        experiment=${REANA_EXPERIMENT-}
```
diegodelemos: Is the trailing `-` necessary? In `-h` it doesn't include it
```
Runs a helloworld demo in namespace of
specified experiment (e.g. '-e atlas').
Specify experiment with '-e' flag or 
$REANA_EXPERIMENT environment variable.'
```

scripts/reana
diegodelemos: We could leave it there if we find it useful for development, all in all it is an option..

scripts/reana
diegodelemos: `-printf` is not supported on BSD, something like:
`find ./configuration-manifests/namespaces/ -name \*.yaml -type f -exec basename {} .yaml ";"`
should be more portable

scripts/reana
diegodelemos: This is not yet update, it prevents successful execution on Mac

scripts/reana
hjhsalo: Quote from some SO-post:
_"${var:-} and ${var-} mean different things. ${var:-} will expand to empty string if var unset or null and ${var-} will expand to empty string only if unset"_
BUT now that I think of it, it might be better with ${REANA_EXPERIMENT:-} so it will default to empty string.

scripts/reana
hjhsalo: I think I got your point. Have to admit that I'm not sure if I'm actually restricting matches with `-o`. It should show only the part which matches pattern.
Tested based on suggestion and I couldn't get `grep -qe "--\<verbose\>"` to work.
How about `flag_verbose=$(echo "$@" | grep -oe "--\<verbose\>")`?

scripts/reana
diegodelemos: Shouldn't we warn the user to either set `-e <experiment>`  or the `REANA_EXPERIMENT` env var? otherwise the request could contain an empty string

PR #29, opened 2017-07-07
config: add reana-server as component

diegodelemos:
* Fix images version to v0.0.1.

Signed-off-by: Diego Rodriguez <diego.rodriguez@cern.ch>

### DIFF
--- a/docs/components.rst
+++ b/docs/components.rst
@@ -7,6 +7,16 @@ REANA system is composed of multiple separated components that are developed
 independently. The components are usually published as Docker images. The
 components collaborate via REST APIs.
 
+reana-server
+------------
+
+REANA component for managing analyses.
+
+- source code: `<https://github.com/reanahub/reana-server>`_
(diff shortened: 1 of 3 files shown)


hjhsalo:
(on behalf of @diegodelemos)

What:
In case there is some trouble related to _reana-server_, when starting the Kubernetes cluster (or minikube) _reana-server_ -pod should be killed with `kubectl`. Reana-server should restart itself automatically.

Why:
_Reana-server_ is accessing "Kubernetes provided metadata" (IP-address of the pod is used as the host) when forming URL for _reana-workflow-controller_ and because Kubernetes starts _reana-server_ before _reana-workflow-controller_, the necessary metadata is not available.
When _reana-server_ -pod is killed, it restarts itself, but this time metadata for _reana-workflow-controller_ -pod is available and _reana-server_ can successfully form a URL that points to _reana-workflow-controller_.

In the future, the way URLs for different components are defined will change and this problem should be solved.

PR #46, opened 2018-02-13
docs: add `wdb` debugging instructions

### LINKED ISSUE #44: docs: full wdb debugging example
Enrich "Developer guide" with a concrete example on how to use `wdb` for debugging REANA cluster components.

Look at old @diegodelemos's hints at https://github.com/reanahub/reana-resources-k8s/pull/28#issuecomment-334534629, update them to v01.0 way of starting the cluster, and expand a little bit the example.

diegodelemos:
* (closes #44).


docs/developerguide.rst
```diff
@@ -9,4 +9,44 @@ understand REANA source code and contribute to it.
...
+
+.. image:: /_static/wdb-debugging-ui.png
+
+
+Limitations
+~~~~~~~~~~~
+
+Live code auto reload (such as Flask auto reload when ``FLASK_DEBUG`` is set)
+is not currently supported due to ``minikube mount`` limitations (it
+continuously updates files ``modify`` field so auto reload is triggered
```
diegodelemos: @hjhsalo could you please confirm this? if it also happens to you then we need to merge https://github.com/reanahub/reana-cluster/pull/48

You can quickly check it by having a [development cluster](http://reana-cluster.readthedocs.io/en/latest/developerguide.html#deploying-latest-master-branch-versions) and running:
```console
$ minikube ssh
> watch 'stat /code/reana-server/setup.py'
  File: '/code/reana-server/setup.py'
  Size: 3286            Blocks: 7          IO Block: 8192   regular file
Device: 25h/37d Inode: 8591305187  Links: 1
Access: (0644/-rw-r--r--)  Uid: (4294967294/ UNKNOWN)   Gid: (4294967294/ UNKNOWN)
Access: 1970-01-01 00:00:00.425043499 +0000
Modify: 2018-01-17 10:08:46.425043499 +0000
Change: 2018-01-17 10:08:46.425043499 +0000
 Birth: -
```

You should see that `Access`, `Change` and `Modify` change. Also if you have an idea on how to prevent this it would be helpful since the impact of this issue means adding an extra step to the debugging process. thanks 🙃

hjhsalo: I couldn't reproduce this behavior of `Access`, `Change` and `Modify` values changing. Initially I thought that the same thing would happen.

Below are the steps I did:
- I build `reana-server` with DEBUG-flag with Docker Client attached to Docker daemon of Minikube VM.
- I cloned all REANA components to `sources`-folder and mounted that folder to Minikube VM on path `/code`.
- I ran `reana-cluster -f ./reana_cluster/configurations/reana-cluster-dev.yaml init` and checked that all the components were deployed and their status is `Running` with `kubectl get pods`
- I double checked that Pod / Container where `reana-server`is running has environment variable `FLASK_DEBUG=1` set (`kubectl exec -it <name-reana-server-pod> bash` + `env | grep DEBUG`).

If I follow logs of `reana-server` Pod / Container I can see the `Restarting with stat` message and I can trigger a restart by e.g. `touch ./reana-server/rest/ping.py`.

I would say that `minikube ssh` isn't touching the files at least on my system (Ubuntu 16.04.XX LTS). Version of `minikube` binary is v0.25.0


I suggest that we check IRL when you are back at the office.

diegodelemos: I have upgraded to `minikube` `0.25.0` and indeed the problem with the files continuously getting updated disappeared 👍. Still need to test if the rest works.

docs/developerguide.rst
```diff
@@ -9,4 +9,44 @@ understand REANA source code and contribute to it.
...
+we are going to debug `reana-server`:
+
+.. code-block:: console
+
+   $ # if we are not connected to the minikube docker daemon we should do it
+   $ eval "$(minikube docker-env)"
+   $ docker build . --build-arg DEBUG=true -t reanahub/reana-server:v0.1.0
+
+Set the breakpoint in your current project code using your preferred editor and
+add `import wdb; wdb.set_trace()` instruction, for instance in the
```
anton-khodak: I would suggest moving this example to a separate block like you did in https://github.com/reanahub/reana-resources-k8s/pull/28#issuecomment-334534629, LGTM in general

diegodelemos: I did this by this time because you had to edit the file inside the container:
```console
$ kubectl exec -ti [pod_name] bash
> emacs|vim /path/to/file/where/to/set/breakpoint.py
# Inside the file set the breakpoint as follows:
# import wdb; wdb.set_trace()
# save the file. Since we have FLASK_DEBUG active Flask will reload.
```
However, now it is possible to do it from your editor thanks to the possibility of mounting your code inside the containers. Because of that I didn't include a code block. What do you think about adding an screenshot like this?
screen shot 2018-02-14 at 3 24 31 pm

docs/developerguide.rst
```diff
@@ -9,4 +9,43 @@ understand REANA source code and contribute to it.
...
+`reana-cluster documentation
+<http://reana-cluster.readthedocs.io/en/latest/developerguide.html#deploying-latest-master-branch-versions>`_,
+to debug a component we need to rebuild its Docker image in debug mode. Let say
+we are going to debug `reana-server`:
+
+.. code-block:: console
+
+   $ # if we are not connected to the minikube docker daemon we should do it
+   $ eval "$(minikube docker-env)"
+   $ docker build . --build-arg DEBUG=true -t reanahub/reana-server:v0.1.0
```
tiborsimko: Hmm, I don't like the possible "overriding" of the `0.1.0` tag here; the tag is supposed to be static.

I'm just using `latest` locally:

```console
$ cd src/reana-cluster
$ cp ./reana_cluster/configurations/reana-cluster-dev.yaml .
$ sed -i -e 's,0.1.0,latest,g' reana-cluster-dev.yaml
$ reana-cluster -f ./reana-cluster-dev.yaml init
```

Shall we change our standard `reana-cluster-dev.yaml` to use `latest` tag by default there?

(and explain in the REANA-Cluster documentation that one can keep versions at some fixed released point, such as `0.1.0`, and change to `latest` only the components you would like to debug)

docs/developerguide.rst
```diff
@@ -9,4 +9,43 @@ understand REANA source code and contribute to it.
 Debugging
 ---------
 
-We use ``wdb``.
+Assuming that REANA is being run in development mode following
```
tiborsimko: Instead of "Assuming...", you can use walk-through mode tone, such as:

"In order to debug a REANA component, you first have to install REANA cluster in the development mode.  You do this as follows...":

docs/developerguide.rst
anton-khodak: Yes, quite illustrative!

docs/developerguide.rst
```diff
@@ -9,4 +9,43 @@ understand REANA source code and contribute to it.
...
+we are going to debug `reana-server`:
+
+.. code-block:: console
+
+   $ # if we are not connected to the minikube docker daemon we should do it
+   $ eval "$(minikube docker-env)"
+   $ docker build . --build-arg DEBUG=true -t reanahub/reana-server:v0.1.0
+
+Set the breakpoint in your current project code using your preferred editor and
+add `import wdb; wdb.set_trace()` instruction, for instance in the
```
tiborsimko: ``import wdb; wdb.set_trace()`` -- in reST, there should be two backquotes

docs/developerguide.rst
```diff
@@ -9,4 +9,43 @@ understand REANA source code and contribute to it.
...
+After doing that we can go to the `wdb` dashboard (you can get ``wdb`` address using `reana-cluster get wdb <http://reana-cluster.readthedocs.io/en/latest/cliapi.html#reana-cluster-get>`_).
+
+.. image:: /_static/wdb-active-sessions.png
+
+And finally select the debugging session.
+
+.. image:: /_static/wdb-debugging-ui.png
+
+
+Limitations
```
tiborsimko: Better not to introduce a new section here to clarify with the other forthcoming tips.  Perhaps **Limitations** in bold is enough.

docs/developerguide.rst
tiborsimko: Adding illustrative screenshot is nice.  You can join it with the using-detailed-walk-through example comment of mine, and use a tone such as:

"Let us introduce the ``wdb`` breakpoint as the first instruction of the ``get_analyses()`` function located in ``reana_server/rest/analyses.py``:"

docs/developerguide.rst
diegodelemos: 👍 I will make PR to use `latest` tag as default for dev

docs/developerguide.rst
```diff
@@ -9,4 +9,60 @@ understand REANA source code and contribute to it.
...
-We use ``wdb``.
+In order to debug a REANA component, you first have to install REANA cluster in
+the development mode (see
+`reana-cluster documentation <http://reana-cluster.readthedocs.io/en/latest/developerguide.html#deploying-latest-master-branch-versions>`_).
+Once you have done this, you have to build the image of the component you are
+working on in development mode and we restart the corresponding pod:
+
+.. code-block:: console
+
+   ~/reana-server $ # if we are not connected to the minikube docker daemon \
```
tiborsimko: Instead of the complex prompt (that different people set up differently):

```
~/reana-server $ foo
~/reana-server $ bar
```

please use `cd` and a simple example prompt, as usual:

```
$ cd src/reana-server
$ foo
$ bar
```

PR #68, opened 2018-07-04
cli: filter components in docker commands

tiborsimko:
* Filters away some components that are not dockerisable (such as
  `reana-cluster`) when attempting to execute docker commands
  (such as `reana docker-build`).

* Adds tests for select_components() as well.

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -47,24 +47,26 @@ REPO_LIST_ALL = [
     'reana-env-aliphysics',
     'reana-env-jupyter',
     'reana-env-root6',
-    'reana.io',
     'reana-job-controller',
     'reana-message-broker',
     'reana-server',
+    'reana-ui',
     'reana-workflow-commons',
     'reana-workflow-controller',
     'reana-workflow-engine-cwl',
     'reana-workflow-engine-serial',
     'reana-workflow-engine-yadage',
     'reana-workflow-monitor',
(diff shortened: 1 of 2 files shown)


reana/cli.py
```diff
@@ -195,6 +196,18 @@ def select_components(components):
     return list(output)
 
 
+def is_component_dockerisable(component):
```
diegodelemos: Would be `is_component_dockerized` more clear? I would expect this function to tell me whether I can make the component Docker compatible or not.

reana/cli.py
```diff
@@ -456,13 +469,14 @@ def docker_build(user, tag, component, no_cache):
...
     components = select_components(component)
     for component in components:
-        if no_cache:
-            cmd = 'docker build --no-cache -t {0}/{1}:{2} .'.format(
-                user, component, tag)
-        else:
-            cmd = 'docker build -t {0}/{1}:{2} .'.format(
-                user, component, tag)
-        run_command(cmd, component)
+        if is_component_dockerisable(component):
```
diegodelemos: Shall we warn the user if the component is not dockerised? (upstairs I wrote dockerized, sorry about that) If it is not, the answer is empty and it might drive people to think that something went wrong.

Something like: `'{0} does not contain a Dockerfile.'.format(component)`.

PR #72, opened 2018-07-31
cli: new install-client and install-cluster commands

tiborsimko:
* Adds two new convenient helper commands to install (and upgrade) latest REANA
  client and cluster scripts.

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -62,6 +62,12 @@ REPO_LIST_ALL = [
 ]
 
 
+REPO_LIST_CLIENT = [
+    'reana-commons',
+    'reana-client',
+]
+
+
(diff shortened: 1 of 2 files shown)


reana/cli.py
```diff
@@ -124,6 +136,22 @@ def cli():  # noqa: D301
         $ reana-cluster -f reana-cluster-latest.yaml init
         $ # we now have REANA cluster running "master" versions of components
 
+    How to install latest ``master`` REANA client CLI script:
+
+    .. code-block:: console
+
+        \b
+        $ workon reana
```
diegodelemos: I would remove this `workon` and the one below since we never do `deactivate` since the `mkvirtualenv`  in L112

PR #103, opened 2018-10-10
cli: easy deletion of pods

### LINKED ISSUE #100: cli: easy deletion of pods
Duringna- REANA development, when one tests some feature branch for e.g. `r-j-controller`, one currently deletes the pods "manually" after compiling the new version, for example:

```console
$ cd reana-job-controller
$ reana git-checkout -b . 72 --fetch
$ reana docker-build -c .
$ kubectl delete pod -l app=job-controller
```

We can simplify this process by providing a new command:

```console
$ reana-dev delete-pod -c .
```

that will basically wrap the ``kubectl delete pod -l app=...`` statements to ease the
developer task.

Note that e.g. `r-w-e-yadage` component creates pods called like ``yadage-default-worker-855d498f7f-nd7fh``, so that we need a configuration table that would be translating component names to the pod prefixes, for example:

```python
COMPONENT_PODS = {
  'reana-workflow-engine-yadage': 'yadage-default-worker',
  'reana-job-controller': 'job-controlller',
   }
```

so that ``kubectl delete pod -l app=...`` statements could be launched.

P.S. Note that `reana delete-pod` would by default delete all the cluster pods since no component is specified...

roksys:
* Adds new command for easier deletion of pods. Closes #100

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -78,6 +78,18 @@ WORKFLOW_ENGINE_LIST_ALL = [
     'yadage'
 ]
 
+COMPONENT_PODS = {
+    'reana-workflow-engine-cwl': 'cwl-default-worker',
+    'reana-db': 'db',
+    'reana-job-controller': 'job-controller',
+    'reana-message-broker': 'message-broker',
+    'reana-workflow-engine-serial': 'serial-default-worker',
+    'reana-server': 'server',
+    'reana-workflow-controller': 'workflow-controller',
+    'reana-workflow-monitor': 'workflow-monitor',
+    'reana-workflow-engine-yadage': 'yadage-default-worker',
+}
+
 
 @click.group()
 def cli():  # noqa: D301
@@ -157,7 +169,7 @@ def cli():  # noqa: D301
         $ cd reana-job-controller
         $ reana-dev git-checkout -b . 72 --fetch
         $ reana-dev docker-build -c .
-        $ kubectl delete pod -l app=job-controller
+        $ reana-dev kubectl-delete-pod -c .
 
     How to test multiple component branches:
 
@@ -168,8 +180,8 @@ def cli():  # noqa: D301
         $ reana-dev git-checkout -b reana-workflow-controller 98
         $ reana-dev git-status
         $ reana-dev docker-build
-        $ kubectl delete pod -l app=job-controller
-        $ kubectl delete pod -l app=workflow-controller
+        $ reana-dev kubectl-delete-pod -c reana-job-controller
+        $ reana-dev kubectl-delete-pod -c reana-workflow-controller
 
     How to release and push cluster component images:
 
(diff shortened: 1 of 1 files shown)


reana/cli.py
```diff
@@ -1047,3 +1059,40 @@ def run_example(component, workflow_engine, output, sleep):  # noqa: D301
...
+    if "ALL" in component:
+        cmd = 'kubectl delete --all pods'
+        run_command(cmd)
+    else:
+        components = select_components(component)
+        for component in components:
+            if component in COMPONENT_PODS:
+                cmd = 'kubectl delete pod -l app={0}'.format(
+                    COMPONENT_PODS[component])
+                run_command(cmd)
```
tiborsimko: Please use ` run_command(cmd, component)` so that the output messages wwould be more informative  (will print the leading component in brackets as usual for other commands)

reana/cli.py
```diff
@@ -157,7 +169,7 @@ def cli():  # noqa: D301
         $ cd reana-job-controller
         $ reana-dev git-checkout -b . 72 --fetch
         $ reana-dev docker-build -c .
-        $ kubectl delete pod -l app=job-controller
+        $ reana-dev kubectl-delete-pod -c reana-job-controller
```
tiborsimko: Can use `-c .` in the help here, so that it mirrors `docker-build -c .` command above.  Basically `.` means current working directory, so that people don't have to type much the component name....

reana/cli.py
```diff
@@ -1047,3 +1059,40 @@ def run_example(component, workflow_engine, output, sleep):  # noqa: D301
...
+
+    """
+    if "ALL" in component:
+        cmd = 'kubectl delete --all pods'
+        run_command(cmd)
+    else:
+        components = select_components(component)
+        for component in components:
+            if component in COMPONENT_PODS:
+                cmd = 'kubectl delete pod -l app={0}'.format(
```
tiborsimko: Please add `--wait=false` to the command options for faster action.

Otherwise this command waits for pod deletion. Keeping our developer-oriented use case, I think we may want to return as soon as possible.  This would permit to plug the command into your editing pipeline, e.g. edit some file in `r-j-c`, save it, which would automatically trigger pod deletion via this command... and we wouldn't want our editors to wait. We could use asynchronous OS process for this, but adding `--wait=false` seems just as good and simpler.

```console
$ time kubectl delete pod -l app=job-controller              
                                                                                  
pod "job-controller-5fcb4c5c9c-djscz" deleted
kubectl delete pod -l app=job-controller  0.09s user 0.01s system 0% cpu 32.204 total
$ time kubectl delete pod --wait=false -l app=job-controller  
                                                               
pod "job-controller-5fcb4c5c9c-fhhqk" deleted
kubectl delete pod --wait=false -l app=job-controller  0.15s user 0.03s system 111% cpu 0.159 total
```

PR #110, opened 2018-11-07
release: v0.4.0

tiborsimko:

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,15 @@
 Changes
 =======
 
+Version 0.4.0 (2018-11-07)
+--------------------------
+
+- Uses common OpenAPI client in communications between workflow engines and job
+  controller.
+- Improves AMQP re-connection handling.
+- Enhances test suite and increases code coverage.
+- Changes license to MIT.
+
 Version 0.3.0 (2018-09-27)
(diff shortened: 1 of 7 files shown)


CHANGES.rst
```diff
@@ -1,6 +1,13 @@
 Changes
 =======
 
+Version 0.4.0 (2018-11-07)
+--------------------------
+
+- Improves AMQP re-connection handling.
+- Enhances test suite and increases code coverage.
+- Changes license to MIT.
+
```
dinosk: We could optionally mention here that yadage and cwl engines switched from using the `requests` library, to a common openapi client for their calls to RJC

tiborsimko: OK, the phrase could go like:

- Workflow engine workers now use common OpenAPI client to connect to job controller.

It may still be a bit too "speaking towards insiders" rather than "speaking towards end-users", but it may be interesting to REANA admins, so perhaps worth mentioning in this way?

dinosk: Indeed it might be more towards developers, just thought of adding it as now all REST API clients in the cluster are communicating via openapi clients, and improves the consistency between the workflow engines. We mention message queue reconnection handling so I think it makes sense to include it as well.

PR #123, opened 2019-01-10
cli: allow parameters for run-example

### LINKED ISSUE #121: cli: run-example -p myparam=myparamval -o myopt=myoptval
Now that `reana-client start` and `reana-client run` commands support overriding workflow parameters (via `-p`) and operational options (via `-o`):

```console
$ cd src/reana-demo-worldpopulation
```

using:

```console
$ reana-client create -f reana.yaml -n worldpopulation.serial
$ reana-client upload ./code ./data -w worldpopulation.serial
$ reana-client start -w worldpopulation.serial -p region=Asia -o CACHE=off
``` 

or simply:

```console
$ reana-client run -p region=Asia -o CACHE=off 
```

it would be good to introduce the same parameter/option selection possibility for the `reana-dev run-example` helper command as well, so that one can simply write:

```console
$ reana-dev run-example -c . -w serial -p region=Asia -o CACHE=off
```

This will be useful for the CI-like functionality.

roksys:
* Allows to pass input parameters for run-example command.
  Closes #121

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -1105,12 +1105,21 @@ def setup_environment():  # noqa: D301
 @click.option('--workflow_engine', '-w', multiple=True,
               default=['cwl', 'serial', 'yadage'],
               help='Which workflow engine to run? [cwl,serial,yadage]')
-@click.option('--output', '-o', multiple=True,
+@click.option('--file', '-f', multiple=True,
               help='Expected output file?')
 @click.option('--sleep', '-s', default=60,
               help='How much seconds to wait for results? [60]')
+@click.option('--parameter', '-p', 'parameters', multiple=True,
+              help='Additional input parameters to override '
+                   'original ones from reana.yaml. '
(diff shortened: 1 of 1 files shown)


reana/cli.py
```diff
@@ -1154,9 +1168,13 @@ def run_example(component, workflow_engine, output, sleep):  # noqa: D301
                         inputdir, workflow_name)
                     run_command(cmd, component)
             # run workflow
+            input_parameters = ' '.join(
+                ['-p ' + parameter for parameter in parameters])
+            operational_options = ' '.join(
+                ['-p ' + option for option in options])
```
tiborsimko: Note the typo here, should be `-o` instead of `-p` for options, otherwise LGTM

roksys: Also `-o` flag is already taken - https://github.com/reanahub/reana/blob/6d921eebb301531e67aed684eb35bceb350973b4/reana/cli.py#L1108

PR #125, opened 2019-01-23
cli: build images with autogenerated dev tag

diegodelemos:
* Dev tag is autogenerated with `git describe --dirty`.

Connects https://github.com/reanahub/reana-cluster/issues/142.


reana/cli.py
```diff
@@ -877,8 +883,12 @@ def git_push(full, component):  # noqa: D301
 @click.option('--build-arg', '-b', default='', multiple=True,
               help='Any build arguments? (e.g. `-b DEBUG=true`)')
 @click.option('--no-cache', is_flag=True)
+@click.option('--dev-tag', is_flag=True,
```
tiborsimko: Hmm, `-t` has the default value `latest`, what about using a special value like `-t dirty` that would generate proper dirty tag `v0.4.0-9-gd68e970`? It may be more easy than having an extra forcing flag `--dev-tag` since people can misunderstand what would be the result of various combinations of `-t foo --dev-tag`.  If we go for a sinle option `-t`, we could simply explain in the docstring all the various options that `-t` understands, a bit like `-c CLUSTER` and `-c .` and `-c ALL` which people may already know so that could "feel at home" when they see similar behaviour in other options too.

tiborsimko: :+1: and perhaps we can consider using dynamic `git describe ---dirty` value as default, unless `-t` is used in which case static `-t` value would take over?  (This would mean we'd have to use `reana-dev docker-build -t latest` locally when developing, but this is easy.)  The dynamic default value would probably make more sense than `latest`...

reana/cli.py
```diff
@@ -906,9 +914,15 @@ def docker_build(user, tag, component, build_arg, no_cache):  # noqa: D301
     :type no_cache: bool
     """
     components = select_components(component)
+    static_tag = tag
     for component in components:
         if is_component_dockerised(component):
             cmd = 'docker build'
+            if not static_tag:
+                tag = run_command('git describe --dirty', component,
```
tiborsimko: Hmm, perhaps take out the function that generates version into something like:

```
def get_current_version(srcdir, dirty=True)
```

that would behave like `get_current_branch()`?  

Could be reusable in other contexts.

(In this case you could even use `subprocess.getoutput()` so that `run_command()` would not even need the newly introduced `return_output`; though this can be handy!)

PR #135, opened 2019-02-24
make: initial release of CI helper scripts

tiborsimko:
* Provides initial simple wrappers useful for (fast) local testing and (slow)
  nightly building purposes.

* Removes the mandatory presence of ``REANA_SRCDIR`` and ``REANA_GITHUB_USER``
  environment variables. The first one is detected from the path and the second
  one can be passed as an option.


Makefile
```diff
@@ -0,0 +1,115 @@
...
+	$(error "Please install Kubectl v1.13.3 or higher")
+endif
+ifndef HAS_MINIKUBE
+	$(error "Please install Minikube v0.34 or higher")
+endif
+	minikube status --profile ${MINIKUBE_PROFILE} || minikube start --profile ${MINIKUBE_PROFILE} --vm-driver ${MINIKUBE_DRIVER} --feature-gates="TTLAfterFinished=true"
+	test -e ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate || virtualenv ${HOME}/.virtualenvs/${VENV_NAME}
+
+clone: # Clone REANA source code repositories locally.
+	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
```
dinosk: I'm using pyenv so the virtualenv directory is not at this location. Not a problem but could be mentioned that it's meant to be used with virtualenv in the docs.

tiborsimko: Yeah, I've used vanilla `virtualenv` as per harmony with the other docs, since we use it as an example everywhere...

PR #139, opened 2019-03-13
make: configurable CPU, RAM, and disk size for minikube

tiborsimko:
* In order to test bigger Docker images such as CMSSW, it is useful to allow
  Minikube to have bigger RAM and disk size.


tiborsimko:
@diegodelemos @okraskaj @roksys These are the defaults that I'm using locally on a i5/8GB laptop.  Works for you?

PR #143, opened 2019-04-05
docs: amend wdb URL retrieval process

diegodelemos:
* Provides one line command to open REANA debugging session
  (closes reanahub/reana-cluster#173).

### DIFF
--- a/docs/developerguide.rst
+++ b/docs/developerguide.rst
@@ -70,8 +70,12 @@ component:
     * Debugger PIN: 221-564-335
    $ curl $REANA_SERVER_URL/api/workflows?access_token=$REANA_ACCESS_TOKEN
 
-After doing that we can go to the `wdb` dashboard (you can get ``wdb`` address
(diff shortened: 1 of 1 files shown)


docs/developerguide.rst
```diff
@@ -70,8 +70,12 @@ component:
     * Debugger PIN: 221-564-335
    $ curl $REANA_SERVER_URL/api/workflows?access_token=$REANA_ACCESS_TOKEN
 
-After doing that we can go to the `wdb` dashboard (you can get ``wdb`` address
-using `reana-cluster get wdb <http://reana-cluster.readthedocs.io/en/latest/cliapi.html#reana-cluster-get>`_).
+After doing that we can go to the `wdb` dashboard:
+
+.. code-block:: console
+
+   $ firefox http://`minikube ip`:31984
```
roksys: What do you think adding a note after code-block something similar to - ``If you don't know Minikube ip and want to open WDB dashboard, you could use $ reana-cluster get wdb.``

diegodelemos: The `minikube ip` command already gives the IP, in fact you can copy paste this to your terminal and it should work (if you don't have `firefox` you can use `open`).

PR #144, opened 2019-04-09
global: kubectl/minikube version upgrade

roksys:


docs/administratorguide.rst
```diff
@@ -193,7 +193,7 @@ script. The typical usage scenario goes as follows:
    $ # install reana-cluster utility
    $ pip install reana-cluster
    $ # deploy new cluster and check progress
```
diegodelemos: We are missing `helm` in this guide. Either we specify it as a requirement or we just add the `helm init`, I would be more for the second option.

PR #153, opened 2019-04-11
make: helm/traefik updates

### LINKED ISSUE #146: make: update reana-cluster installation to use traefik
Makefile should be updated so traefik is installed inside the cluster, needed for interactive sessions to work, see https://github.com/reanahub/reana/blob/master/Makefile#L85.

tiborsimko:
* Updates installation procedures regarding helm and traefik. (closes #146)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ ifndef HAS_MINIKUBE
 	$(error "Please install Minikube v0.35.0 or higher")
 endif
 	minikube status --profile ${MINIKUBE_PROFILE} || minikube start --profile ${MINIKUBE_PROFILE} --vm-driver ${MINIKUBE_DRIVER} --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --disk-size ${MINIKUBE_DISKSIZE} --feature-gates="TTLAfterFinished=true"
+	helm init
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -83,10 +84,11 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE}) && \
 	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-latest.yaml down && \
+	if [ $$(helm ls reana-traefik | grep -q reana-traefik) ]; then helm delete --purge reana-traefik; fi && \
```
diegodelemos: It wasn't working? [`reana-cluster down` should do it](https://github.com/reanahub/reana-cluster/blob/master/reana_cluster/backends/kubernetes/k8s.py#L628) if traefik exists.

tiborsimko: I see, it could be that my cluster had `helm init` but `traefik was not deployed so it was failing....

I can delete it from here (so that Makefile would be nicer) but check whether the logic add if-helm-ls-then-helm-delete shouldn't be added to `down()` to cover for above scenarios...

PR #155, opened 2019-04-11
make: include all example when running CI

### LINKED ISSUE #147: make: include all examples when using `ci` option
When releasing new REANA versions, we test that all our working examples continue to work after the release.

To ease this task, we can add an option to run all working examples with the [`ci` target](https://github.com/reanahub/reana/blob/master/Makefile#L96).

### LINKED ISSUE #150: cli: add possibility to run all examples
Stemmed from https://github.com/reanahub/reana/issues/147#issuecomment-481662367.

Implement:
```console
$ reana-dev run-example -c ALL-EXAMPLES
```
Which would run all working REANA examples and fail if there is a problem (focused on CI).

diegodelemos:
* Adds new option to `reana-dev run-example` so all examples are tested
  (closes #150).

* `make ci` runs all examples (closes #147).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,7 @@ MINIKUBE_MEMORY ?= 3072
 MINIKUBE_DISKSIZE ?= 40g
 TIME_SLEEP ?= 40
 VENV_NAME ?= reana
+DEMO ?= DEMO
 
 # let's detect where we are and whether minikube and kubectl are available:
 HAS_KUBECTL := $(shell command -v kubectl 2> /dev/null)
@@ -45,6 +46,11 @@ help:
 	@echo '  $$ GITHUB_USER=johndoe MINIKUBE_DRIVER=virtualbox make setup clone'
 	@echo '  # how to build latest checked-out sources and run an example:'
 	@echo '  $$ make ci'
+	@echo
+	@echo '  # how to run a specific example:'
+	@echo '  $$ DEMO=reana-demo-helloworld make example'
+	@echo '  # how to run all REANA examples:'
+	@echo '  $$ make example'
 	@echo
 	@echo '  # how to perform an independent automated CI test run:'
 	@echo '  $$ mkdir /tmp/nightlybuild && cd /tmp/nightlybuild'
@@ -89,10 +95,10 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 	docker images | grep '<none>' | awk '{print $$3;}' | xargs -r docker rmi && \
 	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-latest.yaml init --traefik
 
-example: # Run a simple "hello world" analysis example.
+example: # Run a REANA example. By default all REANA examples are executed.
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	eval $$(reana-dev setup-environment) && \
-	reana-dev run-example -c r-d-helloworld -s ${TIME_SLEEP}
+	reana-dev run-example -c ${DEMO} -s ${TIME_SLEEP}
 
 ci: # Perform full Continuous Integration build and test cycle. [main function]
 	make setup
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -14,6 +14,14 @@ import sys
 
 import click
 
+REPO_LIST_DEMO = [
+    'reana-demo-atlas-recast',
+    'reana-demo-bsm-search',
+    'reana-demo-helloworld',
+    'reana-demo-root6-roofit',
+    'reana-demo-worldpopulation',
+]
+
 REPO_LIST_ALL = [
     'reana',
     'reana-client',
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -46,6 +46,11 @@ help:
 	@echo '  # how to build latest checked-out sources and run an example:'
 	@echo '  $$ make ci'
 	@echo
+	@echo '  # how to run a specific example:'
+	@echo '  $$ make run-example EXAMPLE=reana-demo-helloworld'
```
tiborsimko: Usually people use `VAR=val make foo` with Makefiles.

Makefile
```diff
@@ -46,6 +46,11 @@ help:
...
+	@echo '  # how to run all REANA examples:'
+	@echo '  $$ make run-example'
```
tiborsimko: `run-example` would be the only two-word target amongst the rest... That's why I used `example` as a single word.

(It may not be pretty, but `test` was already taken for the unit test... unless we think of renaming the two...)

reana/cli.py
```diff
@@ -375,6 +378,8 @@ def select_components(components):
                                 cover all REANA client components;
                           * (6) special value 'ALL' that will expand to include
                                 all REANA repositories.
+                          * (7) special value 'ALL-EXAMPLES' that will exapand
```
tiborsimko: Perhaps the special value could be just `EXAMPLES`, or `DEMO`, again to have nice analogy with already-existing  `CLUSTER` (and `ALL`) special values... Using again the optics of one single word without dashes.  If `CLUSTER` denotes all repositories concerned with cluster builfing, then `DEMO` could denote all `reana-demo...` examples repositories.  That kind of naming logic.

diegodelemos: I would go for `DEMO` then, I am amending this and previous comments, thanks!

Makefile
```diff
@@ -89,10 +94,10 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 	docker images | grep '<none>' | awk '{print $$3;}' | xargs -r docker rmi && \
 	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-latest.yaml init --traefik
 
-example: # Run a simple "hello world" analysis example.
+example: # Run a REANA example. By default all REANA examples are executed.
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	eval $$(reana-dev setup-environment) && \
-	reana-dev run-example -c r-d-helloworld -s ${TIME_SLEEP}
+	reana-dev run-example -c $${EXAMPLE:-DEMO} -s ${TIME_SLEEP}
```
tiborsimko: Maybe then call it `DEMO=r-d-helloworld make example` so that we use DEMO consistently everywhere?

reana/cli.py
```diff
@@ -42,7 +45,7 @@
     'reana-workflow-engine-yadage',
     'reana-workflow-monitor',
     'reana.io',
-]
+] + REPO_LIST_DEMO
```
tiborsimko: BTW one issue is that some of the demos are not runnable yet (e.g. lhcb one).  Also, other demos may require huuge memory (e.g. the alice one).   So if people would run default `make example`, they would see the command failing.... What about using `REPO_LIST_DEMO` only for the repositories which are known to work well locally?  That's what I meant by referring to "all our working examples"... Mostly just r-d-helloworld, roofit, worldpopulation, recast, bsm... and the others sitting under ALL still until we make them runnable.

Makefile
```diff
@@ -89,10 +94,10 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 	docker images | grep '<none>' | awk '{print $$3;}' | xargs -r docker rmi && \
 	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-latest.yaml init --traefik
 
-example: # Run a simple "hello world" analysis example.
+example: # Run a REANA example. By default all REANA examples are executed.
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	eval $$(reana-dev setup-environment) && \
-	reana-dev run-example -c r-d-helloworld -s ${TIME_SLEEP}
+	echo "reana-dev run-example -c $${DEMO:-DEMO} -s ${TIME_SLEEP}"
```
tiborsimko: One last thing, but you need to rebase anyway: :wink: 

You could use:

```
DEMO ?=  DEMO
```

in the beginning of the script, like the other options (e.g. `VENV_NAME` and friends), for consistency... In this way all the input parameters are localised in one part of the script

reana/cli.py
tiborsimko: Running all examples currently gives:

```console
[reana-demo-bsm-search] reana-client create -f reana-cwl.yaml -n bsm-search.cwl
Usage: reana-client create [OPTIONS]
Try "reana-client create --help" for help.

Error: Invalid value for "-f" / "--file": Path "reana-cwl.yaml" does not exist.
[reana-demo-bsm-search] Command 'reana-client create -f reana-cwl.yaml -n bsm-search.cwl' returned non-zero exit status 2.
make: *** [Makefile:99: example] Error 2
```

This is expected because not all demo examples have all engines, e.g. BSM is only Yadage, etc.

Can you please add some logic `if file-exists reana-cwl.yaml then run-given-example-with-cwl else silently-skip-this-workflow-engine fi` to `reana-dev`?

tiborsimko: Making change and pushing it to your branch...

```diff
@@ -1306,6 +1314,9 @@ def run_example(component, workflow_engine,
     for component in components:
         for workflow_engine in workflow_engines:
             workflow_name = construct_workflow_name(component, workflow_engine)
+            # check whether example contains recipe for given engine:
+            if not os.path.exists(reana_yaml[workflow_engine]):
+                continue
             # create workflow:
             for cmd in [
                 'reana-client create -f {0} -n {1}'.format(
```

PR #158, opened 2019-04-15
docs: supporting non-root users

### LINKED ISSUE #149: docs: document how environment should be set up permission wise
In order to allow applications external to the user job (workflow engines, notebooks...) to manipulate files we are setting the users on those applications to belong to the root group (`groups=0(root)`). Because of this, users should be careful when creating their evironments and set the user they use for the analysis to belong to the root group.

This should be documented in the [Containerise your environment](https://reana.readthedocs.io/en/latest/userguide.html#containerise-your-environment) section.

diegodelemos:
* Since the introduction of Jupyter Notebooks the necessity of accessing
  files by non-root users became evident, now we add all non-root REANA
  users to root group so we can access files and directories
  (closes #149).

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -315,6 +315,34 @@ and push the created image to the DockerHub image registry:
 
     $ docker push johndoe/myenv
 
+**Supporting arbitrary user IDs**
+
+In the Docker container ecosystem, the processes run in the containers by
+default use the ``root`` user identity. However, this may not be secure. If
+you want to improve the security in your environment you can set up your own
+user under which identity the processes will run.
+
+In order for processes to run under any user identity and still be able to
+write to shared workspaces, we use a GID=0 technique
+`as used by OpenShift <https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines>`_:
+
+- UID: you can use any user ID you want;
+- GID: your should add your user to group with GID=0 (the root group)
+
+This will ensure the writable access to workspace directories managed by the
+REANA platform.
+
+For example, you can create the user ``johndoe`` with UID=501 and add the user
+to GID=0 by adding the following commands at the end of the previous
+``Dockerfile``:
+
+.. code-block:: console
+
+    # Setup user and permissions
+    RUN adduser johndoe -u 501 --disabled-password --gecos ""
+    RUN usermod -a -G 0 johndoe
+    USER johndoe
+
 **Testing the environment**
 
 We now have a containerised image representing our computational environment


docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
...
 
+**Supporting non-root users**
+
+Running containers with root user is not secure and therefore dicouraged. If
+you want to improve the security in your environment you can setup your own
+user.
+
+In order to be compatible with the REANA platform you should do the following:
+
+- Add your user to the root group, so anything created by this user will be
```
roksys: ``so anything created by this user will be``. It is more common to use ``everything`` in a positive sentence.

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
...
+  accessible by the REANA user, who also belongs to the root group.
+
+- Any directory or files used in your analysis should have the correct
+  permissions, so the REANA user can access them.
+
+If we put everything together, in orther to create an environment with
```
roksys: probably typo ``in orther ``?

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
 
     $ docker push johndoe/myenv
 
+**Supporting non-root users**
+
+Running containers with root user is not secure and therefore dicouraged. If
```
tiborsimko: -  You can say something like "In the Docker container ecosystem, the processes run in the containers by default using the `root` user identity.  However, this may not be secure..."
 
- typo in `discouraged`

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
...
+
+- Add your user to the root group, so anything created by this user will be
+  accessible by the REANA user, who also belongs to the root group.
+
+- Any directory or files used in your analysis should have the correct
+  permissions, so the REANA user can access them.
+
+If we put everything together, in orther to create an environment with
+previous ``Dockerfile`` supporting a custom user, we would only need to add
+the following to the end of the file:
```
roksys: I would rephrase for example to ``the following code to the end...`` or ``the following commands to the end...``

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
 
     $ docker push johndoe/myenv
 
+**Supporting non-root users**
+
+Running containers with root user is not secure and therefore dicouraged. If
+you want to improve the security in your environment you can setup your own
```
tiborsimko: you can set up your own user under which identity the processes will run.

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
...
+user.
+
+In order to be compatible with the REANA platform you should do the following:
```
tiborsimko: In order for processesqo to run under any user identities and still be able to write to shared workspaces, we use a GID=0 technique as used by OpenShift  https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html#openshift-specific-guidelines  :

- UID: you can use any user ID you want;
- GID: your should add your user to group with GID=0 (the `root` group)

This will ensure the writable access to workspace directories managed by the REANA platform.

Fer example, you can create the user `johndoe` with UID=501 and add the user to GID=0 by following:

...

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
...
+
+- Add your user to the root group, so anything created by this user will be
+  accessible by the REANA user, who also belongs to the root group.
+
+- Any directory or files used in your analysis should have the correct
```
tiborsimko: I think people won't understand what they need to do for "should have correct permissions"... We can leave this bit out, since usually they don't need to do anything and/or if they want to force a certain UID then they will most probably set the permissions right naturally...

docs/userguide.rst
tiborsimko: typo `in order` typo, but see also text suggestion above with a specific UID, so this part may be reformulated...

docs/userguide.rst
```diff
@@ -315,6 +315,33 @@ and push the created image to the DockerHub image registry:
 
     $ docker push johndoe/myenv
 
+**Supporting non-root users**
```
tiborsimko: ... or "Supporting arbitrary user IDs" as in the OpenShift docs, may appear psychologically more positive than "non-root" due to "non" :wink:

diegodelemos:
Thanks for the input, comments addressed :)

docs/userguide.rst
```diff
@@ -315,6 +315,34 @@ and push the created image to the DockerHub image registry:
 
     $ docker push johndoe/myenv
 
+**Supporting arbitrary user IDs**
+
+In the Docker container ecosystem, the processes run in the containers by
+default use the root user identity. However, this may not be secure. If you
```
tiborsimko: In the Docker container ecosystem, the processes in the containers run under the default ``root`` user identity.  

Otherwise it reads well I think :+1:

PR #160, opened 2019-04-15
docs: mention interactive session in user guide

### LINKED ISSUE #148: docs: add interactive sessions to user guide
We should add a new section about Interactive Sessions under the user guide. Options are:

- Split the `Run your analysis on REANA cloud` section, so we can have a part to run the workflows and another to show the scenario of adding a new step and prototyping using the Jupyter Notebook (showing usage as we thought it)
- Create a complete new section about Notebooks (leaving user to imagine how they would use it in their use cases)

cc @tiborsimko

diegodelemos:
* Closes #148.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -460,6 +460,9 @@ follows:
 
 We are done! Our outputs plot should be located in the ``results`` directory.
 
+Note that you can inspect your analysis workspace by opening `Jupyter notebook interactive sessions
+<https://reana-client.readthedocs.io/en/latest/userguide.html#opening-interactive-sessions>`_.
+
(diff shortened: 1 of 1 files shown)


docs/userguide.rst
```diff
@@ -460,6 +460,9 @@ follows:
 
 We are done! Our outputs plot should be located in the ``results`` directory.
 
+Note that you can inspect your analysis workspace by using `Jupyter Nobebooks interactive sessions
```
tiborsimko: - by launching... or by opening, to make reference to the "open" command 
- Jupyter notebook interactive sessions

PR #166, opened 2019-06-17
make: remove macOS incompatible xargs parameter

diegodelemos:


Makefile
```diff
@@ -92,7 +92,7 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-latest.yaml down && \
 	sleep ${TIME_SLEEP} && \
 	minikube ssh --profile ${MINIKUBE_PROFILE} 'sudo rm -rf /var/reana' && \
-	docker images | grep '<none>' | awk '{print $$3;}' | xargs -r docker rmi && \
```
tiborsimko: This would stop the script execution if you have none hits... so it wouldn't work.

Note that I fixed the BSD-xargs argument issue in #165 already by introducing an if-else statement. This PR is can therefore be closed.

PR #174, opened 2019-07-15
make: create reana-db secret on prod mode

diegodelemos:


roksys:
LGTM. Could you also update configuration file name in https://github.com/reanahub/reana/blob/87f5982508360cbebe53d40390d4c2a8a343e8aa/reana/cli.py#L187 ?

PR #186, opened 2019-08-08
cli: add reana-ui as cluster component

diegodelemos:


diegodelemos:
This can be tested together with https://github.com/reanahub/reana-ui/pull/30 and https://github.com/reanahub/reana-cluster/pull/217, using make for everything but for `make deploy`:

```console
$ reana-cluster -f /Users/rodrigdi/code/reanahub/reana/../reana-cluster/reana_cluster/configurations/reana-cluster-minikube.yaml \
                --ui init --traefik --generate-db-secrets
```

I didn't add an option in `make` for UI because for now it will only work at CERN cc @tiborsimko .

PR #198, opened 2019-09-11
makefile: create RS egg-info generated when DEBUG

diegodelemos:
* Add additional pip-check to make sure people won't have dependency
  conflicts after running make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,6 @@ help:
 	@echo
 	@echo '  # Example 3: build and deploy REANA in development mode (with live code changes and debugging)'
 	@echo '  $$ minikube mount $$(pwd)/..:/code'
-	@echo '  $$ cd reana-server && pip install -e . && cd ..  # workaround for reana-workflow-controller#64'
 	@echo '  $$ CLUSTER_CONFIG=dev make build deploy'
 	@echo
 	@echo '  # Example 4: run one small demo example to verify the build'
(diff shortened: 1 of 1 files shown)


diegodelemos:
To test run `reana-dev git-clean` and then `make ci`. REANA-Server shouldn't have the command not found error anymore.

Makefile
```diff
@@ -107,6 +107,12 @@ build: # Build REANA client and cluster components.
 	pip uninstall -y reana-commons reana-client reana-cluster reana-db pytest-reana && \
 	reana-dev install-client && \
 	reana-dev install-cluster && \
+	if [ "${DEBUG}" -gt 0 ]; then \
+		cd ${PWD}/../reana-server/ && \
+		python setup.py bdist_egg && \
```
tiborsimko: Note that we have a line in "Example 3" that tells people they have to run pip install in reana-server.  If we do it for them, we can delete that line above :wink:

Makefile
```diff
@@ -107,6 +107,12 @@ build: # Build REANA client and cluster components.
 	pip uninstall -y reana-commons reana-client reana-cluster reana-db pytest-reana && \
 	reana-dev install-client && \
 	reana-dev install-cluster && \
+	if [ "${DEBUG}" -gt 0 ]; then \
```
tiborsimko: If DEBUG is guaranteed to be numerical, we could leave out the double quotes, here and in the other statement, I guess... Although this works well too.

PR #199, opened 2019-09-12
docs: adds krb5s example

roksys:

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -410,9 +410,14 @@ Declare necessary resources
 ---------------------------
 
 You can declare other additional runtime dependencies that your workflow needs
-for successful operation, such as access to `CVMFS
-<https://cernvm.cern.ch/portal/filesystem>`_. This is achieved by means of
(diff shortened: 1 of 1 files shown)


docs/userguide.rst
```diff
@@ -410,9 +410,13 @@ Declare necessary resources
...
 
 You can declare other additional runtime dependencies that your workflow needs
-for successful operation, such as access to `CVMFS
-<https://cernvm.cern.ch/portal/filesystem>`_. This is achieved by means of
-providing a ``resources`` clause in ``reana.yaml``. For example:
+for successful operation.
+
+**CVMFS**
+
+If your workflow needs to access `CVMFS <https://cernvm.cern.ch/portal/filesystem>`_.
```
tiborsimko: If your workflow needs to access CVMFS filesystem, you should provide a `cvmfs` sub-clause of the `resources` clause that would list all the CVMFS volumes that would be mounted for the workflow execution. For example:

PR #204, opened 2019-09-30
make: fix helm installation

diegodelemos:
- Helm is still creating the Tiller deployment as
  `apiVersion: extensions/v1beta1`, which has been deprecated in
  Kubernetes 1.16. See more here
  https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

  The problem has not been yet fixed in Helm upstream, once the issue
  (https://github.com/helm/helm/issues/6374) is solved we can revert
  this commit.


diegodelemos:
Closing as we won't use this unless we upgrade to Kubernetes 1.16.

Issue #205, opened 2019-09-30
global: upgrade to Kubernetes 1.16


diegodelemos:
Kubernetes [1.16 has been released](https://kubernetes.io/blog/2019/09/18/kubernetes-1-16-release-announcement/). REANA has one issue which depends on this new version: https://github.com/reanahub/reana-workflow-controller/issues/215.

In order to upgrade, we should deal with the [1.16 deprecations](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/).

Actions:
- [ ] ~~Fix existing Helm installation problem with 1.6 (https://github.com/reanahub/reana/pull/204)~~ No need to be fixed as Helm 3.0 doesn't use Tiller anymore
- [x] Deployments change to `apiVersion: apps/v1`
- [ ] ~~Ingresses change to `apiVersion: networking.k8s.io/v1beta1 `~~ This deprecation has been moved to version 1.20
```console
$ rg "extensions/v1beta1"
traefik-web-ui.yaml
15:apiVersion: extensions/v1beta1

reana/Makefile
92:     helm init --override spec.selector.matchLabels.'name'='tiller',spec.selector.matchLabels.'app'='helm' --output yaml | sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' | kubectl apply -f -

reana/docs/gettingstarted.rst
64:               sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' | \

reana/docs/administratorguide.rst
198:               sed 's@apiVersion: extensions/v1beta1@apiVersion: apps/v1@' | \

reana-commons/reana_commons/k8s/api_client.py
28:    if api == 'extensions/v1beta1':
43:                                                    api='extensions/v1beta1'))

reana-workflow-controller/reana_workflow_controller/k8s.py
76:            api_version="extensions/v1beta1",
121:            api_version="extensions/v1beta1",

reana-cluster/reana_cluster/backends/kubernetes/templates/ingress/reana-ingress.yaml
2:apiVersion: extensions/v1beta1

reana-cluster/reana_cluster/backends/kubernetes/templates/deployments/db-template.yaml
2:apiVersion: extensions/v1beta1

reana-cluster/reana_cluster/backends/kubernetes/templates/deployments/cache-template.yaml
2:apiVersion: extensions/v1beta1

reana-cluster/reana_cluster/backends/kubernetes/templates/deployments/server-template.yaml
2:apiVersion: extensions/v1beta1

[...]

PR #206, opened 2019-09-30
make: pin Kubernetes version to 1.15.4

diegodelemos:
* Kubernetes 1.16 comes with a few deprecations which will need
  to be addressed before upgrading, see
  more here https://github.com/reanahub/reana/issues/205.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ MINIKUBE_DISKSIZE ?= 40g
 MINIKUBE_DRIVER ?= kvm2
 MINIKUBE_MEMORY ?= 3072
 MINIKUBE_PROFILE ?= minikube
+MINIKUBE_KUBERNETES ?= v1.15.4
 TIMECHECK ?= 5
 TIMEOUT ?= 300
 VENV_NAME ?= reana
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -88,7 +88,7 @@ endif
 ifndef HAS_MINIKUBE
 	$(error "Please install Minikube v1.0.0 or higher")
 endif
-	minikube status --profile ${MINIKUBE_PROFILE} || minikube start --profile ${MINIKUBE_PROFILE} --vm-driver ${MINIKUBE_DRIVER} --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --disk-size ${MINIKUBE_DISKSIZE} --feature-gates="TTLAfterFinished=true"
+	minikube status --profile ${MINIKUBE_PROFILE} || minikube start --kubernetes-version='v1.15.4' --profile ${MINIKUBE_PROFILE} --vm-driver ${MINIKUBE_DRIVER} --cpus ${MINIKUBE_CPUS} --memory ${MINIKUBE_MEMORY} --disk-size ${MINIKUBE_DISKSIZE} --feature-gates="TTLAfterFinished=true"
```
tiborsimko: Perhaps make it a configuration variable, so that people on older minikube (e.g. 1.0 should still work!) can run the script still easily.  E.g. `MINIKUBE_KUBERNETES` like the other `MINIKUBE_DRIVER` and friends.

diegodelemos: Updated

Makefile
```diff
@@ -12,6 +12,7 @@ MINIKUBE_DISKSIZE ?= 40g
 MINIKUBE_DRIVER ?= kvm2
 MINIKUBE_MEMORY ?= 3072
 MINIKUBE_PROFILE ?= minikube
+MINIKUBE_KUBERNETES ?= v1.15.4
```
tiborsimko: Please add also the doc line, going like:

```
@echo '  MINIKUBE_KUBERNETES Which Kubernetes version to use with Minikube? [default=v1.15.4]'
```

so that the option would be documented when people type naked `make`.

PR #207, opened 2019-10-03
make: fail if cannot connect to Minikube

diegodelemos:
* Eval doesn't fail if the command inside fails, it always returns
  0 exit. If someone runs steps such as prefetch or build, the
  step of connecting to Minikube's Docker daemon will fail silently
  and the Docker client will connect to the local Docker daemon (if
  it exists) or fail if there is no Docker daemon to connect to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ clone: # Clone REANA source code repositories locally.
 build: # Build REANA client and cluster components.
 	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make build\033[0m"
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
-	eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE})  && \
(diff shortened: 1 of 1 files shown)


diegodelemos:
Rebased on top of #206

Makefile
```diff
@@ -101,7 +103,7 @@ clone: # Clone REANA source code repositories locally.
 build: # Build REANA client and cluster components.
 	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make build\033[0m"
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
-	eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE})  && \
+	minikube docker-env --profile ${MINIKUBE_PROFILE} 2&>1 /dev/null && eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE}) && \
```
tiborsimko: This creates files `1` in the current working directory, I'll push a quick fix before merging...

tiborsimko: ... oops, I don't have rights to push onto your feature branch.  Please fix simply by using ` ... > /dev/null` instead of `... 2&>1 /dev/null` and repush yourself.

PR #226, opened 2019-12-03
makefile: use autogenerate secrets

diegodelemos:
* Autogenerate secrets will now generate all needed secrets. Secrets
  such as CERN SSO and GitLab integration ones
  (reanahub/reana-cluster#242).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ ifeq ($(SHOULD_MINIKUBE_MOUNT),1)
 endif
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	minikube docker-env --profile ${MINIKUBE_PROFILE} > /dev/null && eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE}) && \
-	reana-cluster -f ${PWD}/../reana-cluster/reana_cluster/configurations/reana-cluster-minikube$(addprefix -, ${CLUSTER_CONFIG}).yaml $(CLUSTER_FLAGS) down && \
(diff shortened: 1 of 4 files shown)


diegodelemos:
Initially, https://github.com/reanahub/reana-cluster/issues/242 mentions changing the REANA-UI documentation. However, using the current solution is not necessary since the [secrets will only be used in production mode](https://github.com/reanahub/reana-cluster/pull/248/files#diff-553f301b92242889471252a2c1462906R49-R85).

To test the secrets
```
$ make deploy
```
Or with the UI, following the [documentation to create SSO and GitLab-test applications](https://reana-ui.readthedocs.io/en/latest/developerguide.html#authentication):
```
$ CLUSTER_FLAGS=--ui SERVER_URL=`hostname` make deploy
```

diegodelemos:
I forgot to mention that in order to fall into the interactive mode one has to run the `reana-cluster` CLI by itself such as we do in production (we do not use the Makefile for that).

```console
$ reana-cluster -f $REANA_SRCDIR/reana-cluster/reana_cluster/configurations/reana-cluster-minikube.yaml \
                --ui init --interactive
```

tiborsimko:
> `reana-cluster -f FILE --ui init --interactive`

The position of `ui` between `--ui` and `--interactive` may be prone to misunderstandings.. It also lacks click help `reana-cluster init --help` as the other general flags such as `--debug` and friends.  We could also improve it later...

diegodelemos:
We already have a ticket for this here https://github.com/reanahub/reana-cluster/issues/222 I will add these musings there.

PR #229, opened 2019-12-17
make: check for Flask|uWSGI to be initialised

diegodelemos:
* Because of reanahub/reana-server#193, where we want to silence
  start-up logs on R-Server we are losing DB events logs in prod
  like setup. Now we need should check for REANA to be ready to
  either Flask or uWSGI to be ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ endif
 		waited=$$(($$waited+${TIMECHECK})); \
 		if [ $$waited -gt ${TIMEOUT} ];then \
 			break; \
-		elif [ $$(kubectl logs -l app=server -c server --tail=500 | grep -c '^Invenio database created.') -eq 1 ]; then \
+		elif [ $$(kubectl logs -l app=server -c server --tail=500 | grep -ce 'spawned uWSGI master process\|Serving Flask app') -eq 1 ]; then \
 			break; \
 		else \
 			sleep ${TIMECHECK}; \


Makefile
```diff
@@ -157,7 +157,7 @@ endif
 		waited=$$(($$waited+${TIMECHECK})); \
 		if [ $$waited -gt ${TIMEOUT} ];then \
 			break; \
-		elif [ $$(kubectl logs -l app=server -c server --tail=500 | grep -c '^Invenio database created.') -eq 1 ]; then \
```
tiborsimko: IIRC I had been catching `REANA database created` event previously to speed up things, but @roksys mentioned during #224 review that it was too early sometimes on his laptop, so we went for catching later `Invenio database created` event. 

If we switch to Flask/WSGI check here, wouldn't it cause the same too-early problems on some busy or slower laptops?

diegodelemos: All [DB operations](https://github.com/reanahub/reana-server/blob/master/scripts/setup) are done and committed by the time [Flask app](https://github.com/reanahub/reana-cluster/blob/f818028694b403b4f58a1442d10eb040ec9995d4/reana_cluster/backends/kubernetes/templates/deployments/server-template.yaml#L22) or [uWSGI](https://github.com/reanahub/reana-server/blob/ca50d793c94ae36e7c8434191fc457f375b6b62c/Dockerfile#L45-L50) are running so I would say this is the safest point.

PR #239, opened 2020-01-10
helm: cleanup

diegodelemos:
* Use `reana-dev` as Chart name since it is a development one.

* Hides error when `reana-dev` chart is not installed and we try to
  uninstall.

* Adds environment variables from GitLab integration and CERN SSO
  on dev mode, they were missing.

* Changes installation message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ help:
 	@echo '  $$ CLUSTER_FLAGS=debug.enabled=true make build deploy'
 	@echo
 	@echo '  # Example 4: build and deploy REANA with a custom hostname including REANA-UI'
-	@echo '  $$ CLUSTER_FLAGS=ui.enabled=true SERVER_URL=https://reana-local.cern.ch make build deploy'
+	@echo '  $$ CLUSTER_FLAGS=ui.enabled=true SERVER_URL=https://example.org make build deploy'
 	@echo
 	@echo '  # Example 5: run one small demo example to verify the build'
 	@echo '  $$ DEMO=reana-demo-helloworld make example'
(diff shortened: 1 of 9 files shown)


helm/reana-dev/templates/NOTES.txt
```diff
@@ -0,0 +1 @@
+Thanks for flying REANA
```
tiborsimko: ... with a rocket emoticon?!

diegodelemos: yesss! I was a bit sad when removing it 😃

helm/reana-dev/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: reana
+name: reana-dev
```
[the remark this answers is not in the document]
diegodelemos: Chart name and directory have to be the same:
```console
$ helm lint helm/reana-dev
==> Linting helm/reana-dev
[ERROR] Chart.yaml: directory name (reana-dev) and chart name (reana) must be the same

Error: 1 chart(s) linted, 1 chart(s) failed
```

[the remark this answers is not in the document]
tiborsimko: Yeah, I would keep `reana` everywhere, including directory name. As we have it in the `master` right now. (...another option could be to call it something like `minikube` to express the what it does, but it wouldn't be so great either...)

Makefile
```diff
@@ -134,7 +134,7 @@ ifeq ($(SHOULD_MINIKUBE_MOUNT),1)
 endif
 	source ${HOME}/.virtualenvs/${VENV_NAME}/bin/activate && \
 	minikube docker-env --profile ${MINIKUBE_PROFILE} > /dev/null && eval $$(minikube docker-env --profile ${MINIKUBE_PROFILE}) && \
-	helm dep update helm/reana && helm uninstall reana || \
+	helm dep update helm/reana && helm ls | grep -q reana-dev && helm uninstall reana || \
```
mvidalgarcia: Should be `reana`

helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -144,6 +144,8 @@ spec:
           value: ""
         - name: GIT_SSL_NO_VERIFY
           value: "true"
+        - name: REANA_GITLAB_HOST
+          value: {{ .Values.secrets.gitlab.REANA_GITLAB_HOST | default "gitlab.cern.ch" }}
```
diegodelemos: Use secret `reana-cern-gitlab-secrets: REANA_GITLAB_HOST`.

PR #242, opened 2020-01-14
cli: command to create eggs in all python comp

diegodelemos:
* This is needed for code mounted in the REANA cluster to work
  (scenario when running REANA in dev mode).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,9 +111,7 @@ build: # Build REANA client and cluster components.
 	pip uninstall -y reana-commons reana-client reana-db pytest-reana && \
 	reana-dev install-client && \
 	if [ "${DEBUG}" -gt 0 ]; then \
-		cd ${PWD}/../reana-server/ && \
-		python setup.py bdist_egg && \
-		cd -; \
+		reana-dev python-install-eggs; \
 	fi && \
 	pip check && \
 	reana-dev git-submodule --update && \
--- a/Makefile
+++ b/Makefile
@@ -24,8 +24,8 @@ SHELL = /usr/bin/env bash
 HAS_KUBECTL := $(shell command -v kubectl 2> /dev/null)
 HAS_MINIKUBE := $(shell command -v minikube 2> /dev/null)
 DEBUG := $(shell grep -q 'debug.enabled=true' <(echo ${CLUSTER_FLAGS}) && echo 1 || echo 0)
-SHOULD_MINIKUBE_MOUNT := $(shell [ "${DEBUG}" -gt 0 ] && [ -z "`ps -ef | grep -i '[m]inikube mount' 2> /dev/null`" ] && echo 1 || echo 0)
 PWD := $(shell pwd)
+REANA_CODE_DIR=$(shell cd ..; pwd)
 
 all: help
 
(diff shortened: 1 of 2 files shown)


reana/cli.py
```diff
@@ -1574,3 +1574,14 @@ def kubectl_delete_pod(component):  # noqa: D301
                 cmd = 'kubectl delete pod --wait=false -l app={0}'.format(
                     COMPONENT_PODS[component])
                 run_command(cmd, component)
+
+
+@cli.command(name='git-install-eggs')
+def git_install_eggs():
+    """Create eggs-info/ in all REANA infrastructure and runtime components."""
+    reana_ui_index = REPO_LIST_CLUSTER.index('reana-ui')
+    python_cluster_components = (REPO_LIST_CLUSTER[0:reana_ui_index] +
```
mvidalgarcia: What about the following and you can get rid of `reana_ui_index`?
```python
python_cluster_components = [c for c in REPO_LIST_CLUSTER if c != 'reana-ui']
```

reana/cli.py
```diff
@@ -1574,3 +1574,14 @@ def kubectl_delete_pod(component):  # noqa: D301
                 cmd = 'kubectl delete pod --wait=false -l app={0}'.format(
                     COMPONENT_PODS[component])
                 run_command(cmd, component)
+
+
+@cli.command(name='git-install-eggs')
```
tiborsimko: The `reana-dev` commands concerning Git, Docker or Kubectl are usually called fairly exactly after their CLI counterparts, as user type them... so it would be fairly misleading to use "git" as the leading prefix for the eggs commands. Rather, we could use "python" which would be more appropriate. This would give something like  `python-setup-bdist-egg` so perhaps shortened to `python-create-eggs` or some such, using leading "python" prefix.

reana/cli.py
diegodelemos: I was unconsciously trying to avoid iterating, but it is such a small list that this way pays off because of readability. Thanks for the input :)

reana/cli.py
diegodelemos: Agreed, thanks!

reana/cli.py
mvidalgarcia: Yep, that was the main reason, otherwise it was a bit cryptic.

diegodelemos:
Comments addressed, I've also added the VirtualBox mount since we are all in VirtualBox already.

Note: make setup takes double of time now... if someone has a better idea on how to mount the code without having to restart please let me know :)

mvidalgarcia:
> Comments addressed, I've also added the VirtualBox mount since we are all in VirtualBox already.
> 
> Note: make setup takes double of time now... if someone has a better idea on how to mount the code without having to restart please let me know :)

Is it necessary to do `minikube start` then `minikube stop` and then `minikube start` again? Can't we avoid the first start, then check somehow if the sharedfolder exists and only in that case run the stop?

Makefile
```diff
@@ -111,27 +121,14 @@ build: # Build REANA client and cluster components.
...
 	reana-dev install-client && \
 	if [ "${DEBUG}" -gt 0 ]; then \
-		cd ${PWD}/../reana-server/ && \
-		python setup.py bdist_egg && \
-		cd -; \
+		reana-dev python-install-eggs; \
 	fi && \
 	pip check && \
 	reana-dev git-submodule --update && \
 	reana-dev docker-build -b DEBUG=${DEBUG} && \
```
mvidalgarcia: You need to remove `&& \`.

PR #267, opened 2020-02-14
helm: adds watch permission for core kubernetes api client

roksys:
* Addresses reanahub/reana-job-controller/issues/3


diegodelemos:
Travis is not passing for Python2.7 because I've pinned  `pydocstyle>5.0.0` in `pytest-reana` however, `5.0.0` it is not available in Python2.7, can you pin it in here to `3.0.0`?

PR #281, opened 2020-03-25
make: add option to exclude components from build

### LINKED ISSUE #194: make: add option to build/deploy with the UI component
Add a [new configuration](https://github.com/reanahub/reana/blob/245559c72206ec0414dd2fe025e68ca9cde25447/Makefile#L43-L53) to choose if the REANA UI should be built and deployed. By default, it shouldn't do it since it is only ready for CERN.

Currently, it is built by default but never deployed.

diegodelemos:
* Closes #194.

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -314,12 +314,16 @@ def shorten_component_name(component):
     return short_name
 
 
-def find_standard_component_name(short_component_name):
-    """Return standard component name corresponding to the short name.
+def find_standard_component_name(component_name):
+    """Return standard component name corresponding to the component name.
+
+    Note this is an idempotent operation, if ``component_name`` is already
+    standard it will return it as it is.
 
     Example: r-w-controller -> reana-workflow-controller
+             reana-ui       -> reana-ui
 
-    :param short_component_name: short component name
+    :param component_name: component name
     :type component: str
 
     :return: standard component name
@@ -327,15 +331,26 @@ def find_standard_component_name(short_component_name):
 
     :raise: exception in case more than one is found
     """
-    output = []
-    for component in REPO_LIST_ALL:
-        component_short_name = shorten_component_name(component)
-        if component_short_name == short_component_name:
-            output.append(component)
-    if len(output) == 1:
-        return output[0]
-    raise Exception('Component name {0} cannot be uniquely mapped.'.format(
-        'short_component_name'))
+    def _is_standard_name(component_name):
+        """Detect whether the provided component name is already standard."""
+        prefixes = component_name.split('-')[:-1]
+        return all([len(n) > 1 for n in prefixes])
+
+    if _is_standard_name(component_name):
+        standard_component_name = component_name
+    else:
+        output = []
+        for component in REPO_LIST_ALL:
+            component_short_name = shorten_component_name(component)
+            if component_short_name == component_name:
+                output.append(component)
+        if len(output) == 1:
+            standard_component_name = output[0]
+        else:
+            raise Exception('Component name {0} cannot be uniquely '
+                            'mapped.'.format(component_name))
+
+    return standard_component_name
 
 
 def find_reana_srcdir():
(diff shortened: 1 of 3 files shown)


reana/cli.py
```diff
@@ -476,9 +478,60 @@ def select_components(components):
...
+        Instead, we shorten all component names (idempotent operation)
+        and then find the standard component name.
+
+        :param components: List of components
+        :type components: list
+
+        :return: List with all components in standard (long) naming form.
+        :rtype: List
+        """
+        short_components = \
```
tiborsimko: Hmm, perhaps we could take this normalising function outside, as it might be useful elsewhere?

Or perhaps we should simply enrich `find_standard_component_name()` so that it would accept any input, even mixed? (It was originally written as short name -> standard name converter.)

P.S. There are test cases, we can enrich them as well, to make sure about the exclusion behaviour.

Wondering because the comment *"It is not possible..."* seems like a wrong reason... Something we can easily solve outside the function and it would simplify the logic here...

diegodelemos: I was not sure we wanted the behaviour to be changed there, much better by making `find_standard_component_name` idempotent, also tests added, sorry about that 😊, it should be now ready.

reana/cli.py
```diff
@@ -327,15 +327,26 @@ def find_standard_component_name(short_component_name):
...
+        prefixes = component_name.split('-')[:-1]
+        return all([len(n) > 1 for n in prefixes])
+
+    if _is_standard_name(short_component_name):
+        standard_component_name = short_component_name
+    else:
+        output = []
+        for component in REPO_LIST_ALL:
+            component_short_name = shorten_component_name(component)
+            if component_short_name == short_component_name:
```
mvidalgarcia: `component_short_name` vs `short_component_name`? What's the difference? I think the naming can be improved.

Also, correct me if I'm wrong, the argument that this function is receiving is called `short_component_name` but I see that it's first checked if this is already a standard name (`is_standard_name`), in case this is true then the name of the variable is confusing. I think the description of the function can be amended too as now it can receive a non-short component name.

reana/cli.py
```diff
@@ -477,9 +490,40 @@ def select_components(components):
...
+    :return: Set of selected components without ``exclude_components``, all in
+        standard naming form.
+    :rtype: set
+    """
+    st_exclude_components = \
+        [find_standard_component_name(c) for c in exclude_components]
+    non_existing_exclude_components = \
+        set(st_exclude_components).difference(selection)
+    if non_existing_exclude_components:
+        display_message('Unkown component(s) to exclude: {}'.format(
```
mvidalgarcia:
```suggestion
        display_message('Unknown component(s) to exclude: {}'.format(
```

reana/cli.py
```diff
@@ -477,9 +490,40 @@ def select_components(components):
...
+    :param exclude_components: List of components to exclude, either in short
+        or standard naming form.
+    :type selection: set
+    :type exclude_components: list
+
+    :return: Set of selected components without ``exclude_components``, all in
+        standard naming form.
+    :rtype: set
+    """
+    st_exclude_components = \
```
mvidalgarcia: What is `st`?

tests/test_cli.py
```diff
@@ -115,6 +121,7 @@ def test_find_standard_component_name():
             ('reana', 'reana'),
             ('r-server', 'reana-server'),
             ('r-j-controller', 'reana-job-controller'),
+            ('reana-ui', 'reana-ui')
```
mvidalgarcia: Shouldn't it be `'r-ui'`?

reana/cli.py
diegodelemos: Clarified.

tests/test_cli.py
diegodelemos: It has to be precisely `reana-ui` to test that `find_standard_component_name` is an idempotent operation.

reana/cli.py
tiborsimko: `st` is my terminal :wink: https://st.suckless.org/

PR #283, opened 2020-03-31
cli: new `python-unit-tests` command

### LINKED ISSUE #282: reana-dev: introduce easy Python unit testing for multiple repositories
Introduce easy Python unit testing for multiple repositories.

Useful when shared modules (e.g. `r-commons`) change and we want to trigger unit test suite build for some component that did not change (e.g. `r-j-controller`) since this would not be run as part of the usual CI processes until some code is being changed inside `r-j-controller`.

Helps to spot troubles early.

Example interface:

```console
$ reana-dev python-unit-tests -c r-j-controller
$ reana-dev python-unit-tests -c CLUSTER
$ reana-dev python-unit-tests -c ALL
```

The implementation should start new venv for each tested component, kill pyc and eggs, and run unit test suite with latest shared packages (`r-commons`, `r-db` and friends.)

tiborsimko:
* Adds new `python-unit-tests` command that allows to launch all the
  Python unit tests for desired components using independent virtual
  environments for each REANA component. (Takes about ten minutes.)
  (closes #282)

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -11,6 +11,7 @@
 import datetime
 import logging
 import os
+import platform
 import subprocess
 import sys
 import time
@@ -234,6 +235,15 @@ def cli():  # noqa: D301
 
         \b
         $ reana-dev run-example -c reana-demo-root6-roofit -w serial
+
+    How to run Python unit tests in independent virtual environments:
+
+    .. code-block:: console
+
+        \b
+        $ reana-dev python-unit-tests -c r-server -c r-w-controller
+        $ reana-dev python-unit-tests -c CLUSTER
+        $ reana-dev python-unit-tests -c ALL
 
     How to test one component pull request:
 
@@ -547,6 +557,23 @@ def select_workflow_engines(workflow_engines):
             display_message('Ignoring unknown workflow engine {0}.'.format(
                 workflow_engine))
     return list(output)
+
+
+def is_component_python_package(component):
+    """Return whether the component is a Python package.
+
+    Useful to skip running wide unit test commands for those components that
+    are not concerned.
+
+    :param component: standard component name
+    :type component: str
+
+    :return: True/False whether the component is a Python package
+    :rtype: bool
+    """
+    if os.path.exists(get_srcdir(component) + os.sep + 'setup.py'):
+        return True
+    return False
 
 
 def is_component_dockerised(component):
@@ -1464,7 +1491,7 @@ def docker_pull(user, tag, component):  # noqa: D301
 
 @cli.command(name='install-client')
 def install_client():  # noqa: D301
-    """Install latest REANA client Python package and its dependencies.
+    """Install latest REANA client and its dependencies.
 
     \b
     :param upgrade: Should we upgrade? [default=True]
@@ -1671,7 +1698,6 @@ def kubectl_delete_pod(component):  # noqa: D301
                          * (7) special value 'ALL' that will expand to include
                                all REANA repositories.
     :type component: str
-
     """
     if "ALL" in component:
         cmd = 'kubectl delete --all pods --wait=false'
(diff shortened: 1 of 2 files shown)


reana/cli.py
```diff
@@ -1695,6 +1720,61 @@ def python_install_eggs():
...
+                        '{} && cd ../pytest-reana && '
+                        ' pip install . --upgrade'.format(cmd_activate_venv),
+                        '{} && cd ../reana-commons && '
+                        ' pip install . --upgrade'.format(cmd_activate_venv),
+                        '{} && cd ../reana-db && '
+                        ' pip install . --upgrade'.format(cmd_activate_venv),
+                        'git clean -d -ff -x',
+                        '{} && pip install ".[tests]" --upgrade'.format(
+                            cmd_activate_venv),
+                        '{} && python setup.py test'.format(cmd_activate_venv),
```
diegodelemos, line 1768: Why we don't call `./run-tests.sh` here? In https://github.com/reanahub/reana-job-controller/pull/244#issuecomment-608346000 there is work related to moving the invocation of tests inside the RJC image because we cannot run RJC tests locally on Mac (no HTCondor Python package on Mac).

tiborsimko: Ouch, macOS.

I did not want to run `run-tests.sh` because of speed concerns. E.g. we are running `docker build...` inside `run-tests.sh`, which would considerably lengthen the execution, and could cause either mixes with running REANA system or else would eat another disk space.  E.g. checking Sphinx docs, manifest etc is good, but we can perhaps leave this onto Travis CI build?

The command was conceived as a quick "run reasonable Python tests quickly" locally without  messing much with my REANA instance... That kind of feeling.

We can chat IRL whether that's the most common use case for you as well?  We could introduce a flag to distinguish the two perhaps.

diegodelemos: I was thinking of the problem regarding the long execution time because of the builds, and all solutions are just adding complexity. Therefore, being pragmatic, I would be in favor of a flag to exclude components, so on macOS we could run everything except RJC (even though excluding code from a test command sounds wrong, but [even pytest does it](http://doc.pytest.org/en/latest/skipping.html)).

tiborsimko: Another option, if you can *never* run `r-j-controller`'s unit tests on macOS, would be to simply detect the underlying OS and to skip the `r-j-controller` tests and emit a warning to user that either this whole component is skipped or that at least some tests are skipped (once we patch `r-j-controller` tests to separate htc ones from non-htc ones, so that at least *some* unit tests would run on macOS. In this way people won't even have to type `--exclude-components`. WDYT?

tiborsimko: Added detection of `macOS`.

reana/cli.py
```diff
@@ -549,6 +559,23 @@ def select_workflow_engines(workflow_engines):
...
+
+    :param component: standard component name
+    :type component: str
+
+    :return: True/False whether the component is a Python package
+    :rtype: bool
+    """
+    if os.path.exists(get_srcdir(component) + os.sep + 'setup.py'):
+        return True
+    return False
```
mvidalgarcia, line 576: You could just return the value of the condition.

reana/cli.py
```diff
@@ -1695,6 +1721,66 @@ def python_install_eggs():
...
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
```
mvidalgarcia, line 1744: On line 1725 I see that default is `'ALL'`

tiborsimko: Good catch, taken note, now that the branch is merged, let's push a fix for both comments separately

PR #287, opened 2020-04-07
release: 0.7.0-dev20200407

diegodelemos:
* Closes reanahub/docs.reana.io#18.


mvidalgarcia:
Since you're at it... https://github.com/reanahub/reana/blame/master/PUBLISHING.md#L3

PUBLISHING.md
```diff
@@ -37,15 +37,16 @@ Before proceeding we need to generate a new personal token with access to the re
 In the following examples, `cr` is the alias name of the command line tool `chart-releaser`.
```
tiborsimko, line 37: - Please link to chart-releaser web site so that people can install it if they don't have it
- It is better to spell commands in full in docs, i.e. use `chart-releaser` instead of `cr`... it is not that long and people may have different personal alias for `cr`.
- What about moving this to `docs.reana.io` as I commented elsewhere?

index.yaml
```diff
@@ -0,0 +1,36 @@
...
+    home: www.reana.io
+    icon: http://www.reana.io/static/img/logo-reana.svg
+    keywords:
+    - research-data
+    - reproducible-analyses
+    - reproducible-workflows
+    - kubernetes
+    - containers
+    - cwl
+    - reusable-science
```
tiborsimko, line 17: Haven't looked at a keyword list (is there one?) to see what people typically use, but some comments:

- `research-data` better be said `research-data-analyses` or even better just `data-analysis`
- `reusable-science` better be said `reproducible-science` or also `reproducibility`
- another popular term is `scientific-workflows` that we could add
- dunno about another popular term "pipelines", perhaps `scientific-pipelines` or `analysis-pipelines`
- please check also what keywords are used by other systems like Pachyderm, Nextflow, RENKU, etc

PUBLISHING.md
diegodelemos: 1. There is a link in the beginning to chart-releaser on GitHub
2. `cr` is the real name of the command-line tool, I will make it clear
3. I am moving it to docs.reana.io now

tiborsimko: WRT (2), Yes, I just saw separately... Whops! 

Suggested text:

```
In the following examples, `cr` is the [Chart Releaser](...) command-line utility 
allowing to create GitHub releases from Helm charts.
```

Something like that.

diegodelemos:
Releasing documentation moved to docs.reana.io https://github.com/reanahub/docs.reana.io/pull/32/commits/6ac80e1d0a11185a9b8782fb9971cbd54df6d70c

index.yaml
diegodelemos: I've picked the keywords from our GitHub repo:
Screenshot 2020-04-07 at 13 40 11
I guess that if we want to update them we should do it everywhere.

Issue #288, opened 2020-04-10
helm: differentiate between dev values and release values


diegodelemos:
Currently the `values.yaml` stored in the repository point to `latest` tag, which doesn't exist, but we use in development mode (see Makefile).

We shall reflect in the content of the `helm/reana` chart directory the real content of the latest Chart release.

PR #289, opened 2020-04-10
helm: chart directory to reflect latest release

diegodelemos:
Closes https://github.com/reanahub/reana/issues/288

---

This PR includes automatic Helm Chart releases with [Helm Chart Releaser Action](https://github.com/marketplace/actions/helm-chart-releaser?version=v1.0.0-rc.2), leaving the following actions to be taken:
- [x] Set up an access token to be stored in the repository (see more [here](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets)). Probably we want this token to be a [machine user token](https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users), for example `reana`. (should be done by repo owner)
- [x] Create a new branch `gh-pages` for the CR Action to work (it expects a `gh-pages` branch always) 
- [x] Remove [releasing charts docs](http://docs.reana.io/development/releasing/) as it will be automated.

### DIFF
--- a/.github/workflows/helm-releaser.yaml
+++ b/.github/workflows/helm-releaser.yaml
@@ -14,8 +14,8 @@ jobs:
 
       - name: Configure Git
         run: |
-          git config user.name "REANA"
-          git config user.email "reana-team@cern.ch"
+          git config user.name "REANA team"
+          git config user.email "team@reana.io"
 
       - name: Run chart-releaser
         uses: helm/chart-releaser-action@v1.0.0-rc.1
--- a/.github/workflows/helm-releaser.yaml
+++ b/.github/workflows/helm-releaser.yaml
@@ -1,0 +1,25 @@
+name: Helm Chart Releaser
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Configure Git
+        run: |
+          git config user.name "REANA"
+          git config user.email "reana-team@cern.ch"
+
+      - name: Run chart-releaser
+        uses: helm/chart-releaser-action@v1.0.0-rc.1
+        with:
+          charts_dir: helm/reana
+        env:
+          CR_TOKEN: "${{ secrets.CR_TOKEN }}"
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@
 # under the terms of the MIT License; see LICENSE file for more details.
 
 # configuration options that may be passed as environment variables:
+BUILD_TYPE ?= dev
 DEMO ?= DEMO
 GITHUB_USER ?= anonymous
 INSTANCE_NAME ?= reana
@@ -26,6 +27,7 @@ DEBUG := $(shell grep -q 'debug.enabled=true' <(echo ${CLUSTER_FLAGS}) && echo 1
 PWD := $(shell pwd)
 REANA_CODE_DIR=$(shell cd ..; pwd)
 TRUNC_INSTANCE_NAME=$(shell echo ${INSTANCE_NAME} | head -c 10 | xargs echo)
+VALUES_YAML_PATH := $(shell test "${BUILD_TYPE}" = "dev" && echo helm/configurations/values-dev.yaml || echo helm/reana/values.yaml)
 
 all: help
 
@@ -41,6 +43,8 @@ help:
 	@echo
 	@echo 'Configuration options:'
 	@echo
+	@echo '  BUILD_TYPE              Is it a dev build or a release build? [default=dev]'
+	@echo '  BUILD_ARGUMENTS         Space separated list of build arguments. [e.g. "COMPUTE_BACKENDS=htcondorcern" no build arguments are passed by default]'
 	@echo '  CLUSTER_FLAGS           Which values need to be passed to Helm? [e.g. "debug.enabled=true,ui.enabled=true"; no flags are passed by default]'
 	@echo '  DEMO                    Which demo example to run? [e.g. "reana-demo-helloworld"; default is several]'
 	@echo '  EXCLUDE_COMPONENTS      Which REANA components should be excluded from the build? [e.g. reana-ui,reana-message-broker]'
@@ -81,7 +85,12 @@ help:
 	@echo '  # Example 8: perform full CI build-and-test cycle'
 	@echo '  $$ make ci'
 	@echo
-	@echo '  # Example 8: perform full CI build-and-test cycle in an independent cluster'
+	@echo '  # Example 9: perform full CI build-and-test cycle for a given release before publishing'
+	@echo '  $$ GITHUB_USER=reanahub BUILD_TYPE=release make ci'
+	@echo '  $$ # If everything goes well you can publish all the images to DockerHub'
+	@echo '  $$ reana-dev docker-push -t auto -u reanahub'
+	@echo
+	@echo '  # Example 10: perform full CI build-and-test cycle in an independent cluster'
 	@echo '  $$ mkdir /tmp/nightly && cd /tmp/nightly'
 	@echo '  $$ git clone https://github.com/reanahub/reana && cd reana'
 	@echo '  $$ INSTANCE_NAME=nightly make ci'
@@ -118,6 +127,16 @@ clone: # Clone REANA source code repositories locally.
 
 build: # Build REANA client and cluster components.
 	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make build\033[0m"
+ifeq ($(BUILD_TYPE),release)
+	make release-build
+else ifeq ($(BUILD_TYPE),dev)
+	make dev-build
+else
+	echo "Unknown build type $$BUILD_TYPE"
+endif
+
+dev-build:
+	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make dev-build\033[0m"
 	source ${HOME}/.virtualenvs/${INSTANCE_NAME}/bin/activate && \
 	minikube docker-env --profile ${INSTANCE_NAME} > /dev/null && eval $$(minikube docker-env --profile ${INSTANCE_NAME}) && \
 	pip install . --upgrade &&  \
@@ -129,6 +148,30 @@ build: # Build REANA client and cluster components.
 	pip check && \
 	reana-dev git-submodule --update && \
 	reana-dev docker-build -b DEBUG=${DEBUG} $(addprefix --exclude-components , ${EXCLUDE_COMPONENTS})
+
+release-build:
+	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make release-build\033[0m"
+	source ${HOME}/.virtualenvs/${INSTANCE_NAME}/bin/activate && \
+	minikube docker-env --profile ${INSTANCE_NAME} > /dev/null && eval $$(minikube docker-env --profile ${INSTANCE_NAME}) && \
+	echo "Upgrading REANA to latest master ..." && \
+	reana-dev git-upgrade -c CLUSTER && \
+	echo "Cleaning components directories ..." && \
+	reana-dev git-clean -c CLUSTER && \
+	BUILT_IMAGES_FILE_PATH=$(shell echo /tmp/reana-images-`git describe --dirty`.txt) && \
+	echo "Building images with no cache ..." && \
+	reana-dev docker-build -c CLUSTER --no-cache \
+		$(foreach argument, ${BUILD_ARGUMENTS}, -b ${argument}) \
+		-u "${GITHUB_USER}" -t auto \
+		--output-component-versions $$BUILT_IMAGES_FILE_PATH \
+		$(addprefix --exclude-components , ${EXCLUDE_COMPONENTS}) && \
+	echo "Building reanahub/krb5 image" && \
+	cd ../reana-job-controller && \
+	REANA_KRB5_IMAGE_NAME="reanahub/krb5:latest" && \
+	docker build --file krb5/Dockerfile --tag $$REANA_KRB5_IMAGE_NAME . && \
+	echo $$REANA_KRB5_IMAGE_NAME >> $$BUILT_IMAGES_FILE_PATH && \
+	echo "The following images have been built:" && \
+	./scripts/update_images.sh $$BUILT_IMAGES_FILE_PATH && \
+	rm $$BUILT_IMAGES_FILE_PATH
 
 deploy: # Deploy/redeploy previously built REANA cluster.
 	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make deploy\033[0m"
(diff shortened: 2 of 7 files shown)


diegodelemos:
> Note: the naming convention stated in this comment is the used inside the PR, so it is in prototype status and can be modified.

Inside the Makefile's recipe to build production images, I am passing all compute backends as this is what we do for production... but if we do so RJC won't start:
```console
$ kubectl logs reana-run-batch-826a4748-27c8-4b6d-9988-28aa811e60c5-z86mp job-controller                                                                                                [10:45:05]
 * Serving Flask app "reana_job_controller/app.py"
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
Configuration Error Line 0 while reading config source /usr/bin/cernbatchsubmit|
not a valid command, errno=2 : No such file or directory
```

This points in the direction of having different images, which leads at least to two questions right now:
1. How should the images be separated?
     1. release VS CERN specific
     2. by backend flavour.
     3. any other ideas?

      Regardless the decision, the outcome can be very well encapsulated in our deployment repository, where the `values.yaml` will point to (1. i.) `reanahub/reana-job-controller-cern:v0.7.0` or (1. ii.) `reanahub/reana-job-controller-condor:v0.7.0` and the image in the public release will go just with Kubernetes and be `reanahub/reana-job-controller:v0.7.0`.
2. How is the logic to build them going to be split across our code base? Currently, there is `PUBLISH=1 make prod-build`
     1. We extend by parameters: `COMPUTE_BACKENDS=kubernetes,htcondorcern,slurmcern make prod-build`, or even `BUILD_ARGUMENTS="COMPUTE_BACKENDS=kubernetes,htcondorcern,slurmcern" make prod-build` (the latter smells like a hell to be implemented in Make)
     2. We decouple by targets: `make release-build`, `make cern-build` (this hints that we should leave the build logic in `gitlab.cern.ch/reana/deployment`, so this Makefile will be having only `make release-build`). Note that this is the approach I initially took, otherwise I should have used the already existing `make build`.
     3. any other ideas?


@tiborsimko @mvidalgarcia @alintulu

tiborsimko:
> 1. How should the images be separated?

I like option 1ii, by backend flavour. Helps to reuse the same image for many deployments, e.g. we could have CERN general instance of REANA, CERN-ATLAS instance, etc.

Kubernetes is the default backend, so it should be included always, and additional backends should be included in alphabetical order, for example:

```
reanahub/reana-job-controller-htcondorcern-htcondornyu-slurmcern:0.7.0
```

> 2. How is the logic to build them going to be split across our code base?

I like option 2i, extending build arguments.  It is nicely consistent with option 1ii.  People specify which backends they want to build, and the images are named accordingly, and pushed accordingly.

> `make prod-build`

Haven't look at the code, but how would `prod-build` play with `production/values.yaml`? Have you thought about `dev-build` and `development/values.yaml` private builds as well? Something one could also streamline by extending build parameters, for example:

```
$ COMPUTE_BACKENDS=foo,bar CLUSTER_FLAGS="ui.enabled=true" BUILD_TYPE="production" make build push
...
reanahub/reana-job-controller-bar-foo:0.6.0-28-g080898f

$ COMPUTE_BACKENDS=foo CLUSTER_FLAGS="ui.enabled=true" BUILD_TYPE="development" make build push
...
reanahub/reana-job-controller-foo:0.6.0-28-g080898f_development
```

i.e. "kubernetes" always included (and hence omitted in the published image name), and build type (prod/dev) included after version number (with "production" omitted) after underscore... Just an example of possible image naming conventions.

BTW I'm assuming here that e.g. NYU deployment does not need anything special built-in inside images, that any specifics can be expressed as compute backend types, and any other further things (e.g. server URL for the reana-ui! e.g. SSO!) can be passed via other build arguments if need be.

Speaking of which, the default URL in REANA-UI should probably by localhost (for people trying out system locally), and we could have:

```
reanahub/reana-ui-cern:0.6.0-28-g080898f_development
reanahub/reana-ui-cern:0.6.0-28-g080898f_qa
reanahub/reana-ui-cern:0.6.0-28-g080898f
reanahub/reana-ui-nyu:0.6.0-28-g080898f_development
reanahub/reana-ui-nyu:0.6.0-28-g080898f_qa
reanahub/reana-ui-nyu:0.6.0-28-g080898f
```

similarly as for r-j-controller.

Ideally, the URL should be an environment variable or something, that we don't need to publish separate images, obviously... So the "flavour" will handle only different authentication techniques, say local accounts vs CERN SSO vs NYU SSO etc.

Talk IRL?

diegodelemos:
After chatting IRL, we decide to go for `1. ii.` and `2. i.`. For `2. i.` to happen we will merge all build-related actions into the `build` target and use sub-targets to point to `dev-build` and `release-build`.

.github/workflows/helm-releaser.yaml
```diff
@@ -0,0 +1,25 @@
...
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+
+      - name: Configure Git
+        run: |
+          git config user.name "REANA"
+          git config user.email "reana-team@cern.ch"
```
tiborsimko, line 18: - user name "REANA team"  (or "REANAHub team"?)
- email: `team@reanahub.io`

Makefile
```diff
@@ -41,6 +44,8 @@ help:
 	@echo
 	@echo 'Configuration options:'
 	@echo
+	@echo '  BUILD_TYPE              Is it a release build or a development build? [default=development]'
+	@echo '  BUILD_ARGUMENTS         Space separated list of build arguments. [e.g. "COMPUTE_BACKENDS=htcondorcern" no build arguments are passed by default]'
```
tiborsimko, line 48: As discussed IRL a few days ago, aren't we going to separate arguments into their own respective environment variables?

Instead of:

```console
$ BUILD_ARGUMENTS="COMPUTE_BACKENDS=htcondorcern,slurmcern VERBOSITY_LEVEL=info" make foo
```

going for separating them:

```
$ BUILD_COMPUTE_BACKENDS=htcondorcern,slurmcern BUILD_VERBOSITY_LEVEL=info make ci
```

This is the usual envvar usage paradigm, and it could also help us to distinguish target components one day, for example:

```console
$ export REANA_JOB_CONTROLLER_COMPUTE_BACKENDS=htcondorcern,slurmcern
$ export REANA_SERVER_VERBOSITY=info
$ make ci
```

Just some examples.

P.S. We can do this also after merge.

Makefile
```diff
@@ -51,6 +56,7 @@ help:
 	@echo '  MINIKUBE_DRIVER         Which vm driver to use for Minikube? [default=virtualbox]'
 	@echo '  MINIKUBE_MEMORY         How much memory to allocate for Minikube? [default=3072]'
 	@echo '  MINIKUBE_KUBERNETES     Which Kubernetes version to use with Minikube? [default=v1.16.3]'
+	@echo '  PUBLISH_IMAGES          Should the production images be pushed to the registry? [0|1, default=0]'
```
tiborsimko, line 59: I think `PUBLISH_IMAGES` should be dropped, it is better to do this manually, as presented in Example 9.

Makefile
```diff
@@ -118,6 +129,16 @@ clone: # Clone REANA source code repositories locally.
...
 	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make build\033[0m"
+ifeq ($(BUILD_TYPE),release)
+	make release-build
+else ifeq ($(BUILD_TYPE),development)
+	make dev-build
+else
+	echo "Unknown build type $$BUILD_TYPE"
+endif
+
+dev-build: # Build REANA client and cluster components.
```
tiborsimko, line 140: Do not document `dev-build` target so that it would not appear in the help when people type `make`.

Makefile
```diff
@@ -130,6 +151,30 @@ build: # Build REANA client and cluster components.
 	reana-dev git-submodule --update && \
 	reana-dev docker-build -b DEBUG=${DEBUG} $(addprefix --exclude-components , ${EXCLUDE_COMPONENTS})
 
+release-build:
+	@echo -e "\033[1;32m[$$(date +%Y-%m-%dT%H:%M:%S)]\033[1;33m reana:\033[0m\033[1m make release-build\033[0m"
+	source ${HOME}/.virtualenvs/${INSTANCE_NAME}/bin/activate && \
+	minikube docker-env --profile ${INSTANCE_NAME} > /dev/null && eval $$(minikube docker-env --profile ${INSTANCE_NAME}) && \
+	echo "Upgrading REANA to lastest master ..." && \
```
tiborsimko, line 158: `s/lastest/latest/g`

Makefile
```diff
@@ -130,6 +151,30 @@ build: # Build REANA client and cluster components.
...
+	reana-dev git-upgrade -c CLUSTER && \
+	echo "Cleaning components directories ..." && \
+	reana-dev git-clean -c CLUSTER && \
+	BUILT_IMAGES_FILE_PATH=$(shell echo `pwd`/reana-images-`git describe --dirty`.txt) && \
+	echo "Building images with no cache ..." && \
+	reana-dev docker-build -c CLUSTER --no-cache \
+		$(foreach argument, ${BUILD_ARGUMENTS}, -b ${argument}) \
+		-u "${GITHUB_USER}" -t auto \
+		--output-component-versions $$BUILT_IMAGES_FILE_PATH \
```
tiborsimko, line 167: Exclude the image file from the repo via `.gitignore`, or create it in tmpdir?

helm/reana-dev/values.yaml
```diff
@@ -0,0 +1,18 @@
+# REANA components pointing to `latest`, locally built master branch
```
tiborsimko, line 1: `reana-dev` subdirectory only holds `values.yaml`.  Couldn't we simply have `values-dev.yaml`?  Thinking of future extensions, we won't create a new directory for each configuration... It would scale better if we could have something like:

```
helm/reana/charts/...
helm/configurations/minikube-production.yaml
helm/configurations/minikube-development.yaml
helm/configurations/minikube-develelopment-vc3.yaml
...
```

(... also, we have `dev-build` in Makefile, `BUILD_TYPE` uses "development"... I used longer version "development" here, but we could also standardise into "dev"...)

scripts/update_images.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+# -*- coding: utf-8 -*-
```
tiborsimko, line 3: - `coding` is not necessary, a reminder from Python I guess?
- There are quite a few warnings if you run `shellcheck`, but we can fix this later.

helm/reana/values.yaml
```diff
@@ -57,31 +57,31 @@ db_env_config:
 components:
   reana_server:
     imagePullPolicy: IfNotPresent
-    image: reanahub/reana-server
+    image: reanahub/reana-server:0.6.0-35-gafd0a7f
```
tiborsimko, line 60: Hmm, this would break local production-like build workflows:

```console
$ DEMO=r-d-r-roofit EXCLUDE_COMPONENTS=r-ui,r-m-broker make ci
```

as these would not use `latest` anymore, but the released versions, isn't it?

If so, we really need those several configuration files, as I mentioned in
https://github.com/reanahub/reana/pull/289#pullrequestreview-394489878, for example:

```
configurations/minikube-dev-latest.yaml
configurations/minikube-prod-latest.yaml
configurations/minikube-prod-released.yaml
...
```

Or else some `sed` tricks in `Makefile` to bring back the `latest` for these cases.

Makefile
diegodelemos: Yes, it is actually a left over, thanks!

helm/reana/values.yaml
diegodelemos: It shouldn't break the local production-like build as the [`values.yaml` path](https://github.com/reanahub/reana/pull/289/files#diff-b67911656ef5d18c4ae36cb6741b7965R31) depends on whether the BUILD is a release or development.
```console
$ BUILD_TYPE=release make ci # will run CI pointing to release tags
$ make ci # implicitly being BUILD_TYPE=development, will always point to latest
```

Maybe what could be improved is the naming? We have `BUILD_TYPE` being one of `['release', 'development']`, that's where the confusion comes I guess? Something like `['release', 'local']` or any other suggestion you have.

scripts/update_images.sh
diegodelemos: Yes, it was a reminder. Regarding spellcheck... I have the linter activated in the editor, but since the last upgrade, I have no spellcheck anymore... I've installed it again, thanks for the heads-up!

helm/reana/values.yaml
tiborsimko: `BUILD_TYPE` is equal to `development` by default, but if you launch the `DEMO=r-d-r-roofit EXCLUDE_COMPONENTS=r-ui,r-m-broker make ci` command I mentioned above, the deployment procedure tries to download tagged images on my system instead of using `latest`... Have you tried to run it in your "reana" production-like environment?

Ah, I see where the error is:

```console
$  make deploy -n
bash: line 0: test: development: integer expression expected
...
helm install reana helm/reana  -fhelm/reana/values.yaml --wait && \
```

A wrong `test` statement when setting `VALUES_YAML_PATH`.  

Here's the fix:

```diff
diff --git a/Makefile b/Makefile
index c230276..909bab4 100644
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ DEBUG := $(shell grep -q 'debug.enabled=true' <(echo ${CLUSTER_FLAGS}) && echo 1
 PWD := $(shell pwd)
 REANA_CODE_DIR=$(shell cd ..; pwd)
 TRUNC_INSTANCE_NAME=$(shell echo ${INSTANCE_NAME} | head -c 10 | xargs echo)
-VALUES_YAML_PATH := $(shell test ${BUILD_TYPE} -eq development && echo helm/reana-dev/values.yaml || echo helm/reana/values.yaml)
+VALUES_YAML_PATH := $(shell test ${BUILD_TYPE} = "development" && echo helm/reana-dev/values.yaml || echo helm/reana/values.yaml)

 all: help
```

Makefile
```diff
@@ -81,6 +87,11 @@ help:
 	@echo '  # Example 8: perform full CI build-and-test cycle'
 	@echo '  $$ make ci'
 	@echo
+	@echo '  # Example 9: perform full CI build-and-test cycle for a given release before publishing'
+	@echo '  $$ GITHUB_USER=reanahub BUILD_TYPE=release make ci'
+	@echo '  $$ # If everything goes well you can publish all the images to DockerHub'
+	@echo '  $$ reana-dev docker-push -t auto -u reanahub'
+	@echo
 	@echo '  # Example 8: perform full CI build-and-test cycle in an independent cluster'
```
mvidalgarcia, line 93:
```suggestion
	@echo '  # Example 10: perform full CI build-and-test cycle in an independent cluster'
```

Makefile
```diff
@@ -130,6 +149,30 @@ build: # Build REANA client and cluster components.
...
+	reana-dev git-clean -c CLUSTER && \
+	BUILT_IMAGES_FILE_PATH=$(shell echo /tmp/reana-images-`git describe --dirty`.txt) && \
+	echo "Building images with no cache ..." && \
+	reana-dev docker-build -c CLUSTER --no-cache \
+		$(foreach argument, ${BUILD_ARGUMENTS}, -b ${argument}) \
+		-u "${GITHUB_USER}" -t auto \
+		--output-component-versions $$BUILT_IMAGES_FILE_PATH \
+		$(addprefix --exclude-components , ${EXCLUDE_COMPONENTS}) && \
+	echo "Building reanahub/krb5 image" && \
+	REANA_KRB5_IMAGE_NAME=$(shell cd ../reana-job-controller && echo reanahub/krb5:latest) && \
```
mvidalgarcia, line 168: Why is it needed here `cd ../reana-job-controller`? You're doing it in the next step.

diegodelemos: This is actually a leftover, before I was building the tag with `git describe dirty` but we only point to latest, thanks!

PR #290, opened 2020-04-13
helm: load uwsgi config as configmap

mvidalgarcia:
closes reanahub/reana-server#228

## metrics-server

Many performance tests were done using the new [`metrics-server`](https://kubernetes.io/docs/tasks/debug-application-cluster/resource-metrics-pipeline/)

```console
$ kubectl top pod reana-server-xyx
NAME.                CPU(cores)         MEMORY(bytes)   
reana-server-xyz.    0m                 312Mi


$ kubectl get --raw /apis/metrics.k8s.io/v1beta1/namespaces/default/pods/reana-server-xyz  | jq
{
  "kind": "PodMetrics",
  "apiVersion": "metrics.k8s.io/v1beta1",
  "metadata": {
    "name": "reana-server-xyz",
    "namespace": "default",
    "selfLink": "/apis/metrics.k8s.io/v1beta1/namespaces/default/pods/reana-server-xtz",
    "creationTimestamp": "2020-04-13T14:58:14Z"
  },
  "timestamp": "2020-04-13T14:58:00Z",
  "window": "1m0s",
  "containers": [
    {
      "name": "rest-api",
      "usage": {
        "cpu": "0",
        "memory": "220124Ki"
      }
    },
    {
      "name": "scheduler",
      "usage": {
        "cpu": "0",
        "memory": "99628Ki"
      }
    }
  ]
}


```

Memory usage doesn't seem to skyrocket at any moment, the maximum I got was 360Mi (2 processes and 2 threads in uwsgi conf), so in that regard, it seems to be fine. Let's see some more advanced test using [locust](https://locust.io/) to determine the best value for [processes](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#processes) and [threads](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#threads).

##  locust

Conditions: 2 minutes, 500 users, hatch rate 10.

### a) 1 thread, 1 process

image

image

image

(same for all)
image

**Memory usage:** ~230Mi

### b) 2 threads, 2 processes

image

image

**Memory usage:** ~360Mi

### c) 4 threads, 6 process (invenio-rdm)

image

image

image

**Memory usage:** ~735Mi

## Some conclusions
- When increasing the number of processes and threads memory usage (230Mi vs 735Mi) increases but the server is much more performant (27.9 vs 61.5 requests per second).
- The number of failures is minimum for a) and c) whereas b) seems to struggle more.
- We should find a balance between memory usage and performance. If using ~700Mi of memory is not a problem we should definitely go for something similar to c).

### DIFF
--- a/helm/reana/templates/reana-server.yaml
+++ b/helm/reana/templates/reana-server.yaml
@@ -40,6 +40,10 @@ spec:
         args: ["set -e; ./scripts/setup; invenio run -h 0.0.0.0 -p 5000"]
         tty: true
         stdin: true
+        {{- else }}
+        command: ["/bin/sh", "-c"]
+        args: ["./scripts/setup > /var/log/reana-server-init-output.log 2>&1 &&
+                uwsgi --ini /var/reana/uwsgi/uwsgi.ini"]
         {{- end }}
         volumeMounts:
           {{- if .Values.debug.enabled }}
@@ -48,6 +52,8 @@ spec:
           {{- end }}
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
+          - name: uwsgi-config
+            mountPath: '/var/reana/uwsgi'
         env:
           - name: REANA_COMPONENT_PREFIX
             value: {{ include "reana.prefix" . }}
@@ -185,3 +191,7 @@ spec:
         hostPath:
           path: /code/reana-server
       {{- end }}
+      - name: uwsgi-config
+        configMap:
+          defaultMode: 420
+          name: uwsgi-config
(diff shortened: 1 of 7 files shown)


helm/reana/templates/reana-server.yaml
```diff
@@ -40,6 +40,10 @@ spec:
         args: ["set -e; ./scripts/setup; invenio run -h 0.0.0.0 -p 5000"]
         tty: true
         stdin: true
+        {{- else }}
+        command: ["/bin/sh", "-c"]
+        args: ["./scripts/setup > /var/log/reana-server-init-output.log 2>&1 &&
+                uwsgi --wsgi-disable-file-wrapper --ini /var/reana/uwsgi/uwsgi.ini"]
```
diegodelemos, line 46: Can we also add to the configuration file the `--wsgi-disable-file-wrapper` option?

mvidalgarcia: Are you sure that is going to work? I couldn't find that option [on the docs](https://uwsgi-docs.readthedocs.io/en/latest/Options.html), that's why I left it in the command.

Issue #292, opened 2020-04-17
make: respect minikube profile


tiborsimko:
The CI build fails after recent changes for me:

```
...
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Thanks for flying REANA 🚀
Environment variables could not be generated:
Command '['minikube', 'ip', '--profile', 'default']' returned non-zero exit status 65.
make[1]: Leaving directory '/home/simko/private/project/reana/src/reana'
```

We seem to be missing `--profile ${INSTANCE_NAME}` (note "default" vs "reana") at some place... No time to look at it right now, hence ticketising.

PR #294, opened 2020-04-17
cli: remove namespace from setup-environment

diegodelemos:
* This allows to setup-environment for multiple local REANA
  instances and fixes the problem of always using `default` as
  instance name (#292).

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -1511,11 +1511,7 @@ def install_client():  # noqa: D301
 @click.option(
     '--insecure-url', is_flag=True,
     help='REANA Server URL with HTTP.')
-@click.option(
-    '--namespace',
-    default='default',
-    help='Namespace of the components which configuration should be produced.')
-def setup_environment(server_hostname, insecure_url, namespace):  # noqa: D301
+def setup_environment(server_hostname, insecure_url):  # noqa: D301
     """Display commands to set up shell environment for local cluster.
 
     Display commands how to set up REANA_SERVER_URL and REANA_ACCESS_TOKEN
@@ -1527,14 +1523,13 @@ def setup_environment(server_hostname, insecure_url, namespace):  # noqa: D301
         component_export_line = 'export {env_var_name}={env_var_value}'
         export_lines.append(component_export_line.format(
             env_var_name='REANA_SERVER_URL',
-            env_var_value=server_hostname or get_external_url(insecure_url,
-                                                              namespace)))
+            env_var_value=server_hostname or get_external_url(insecure_url)))
 
         get_admin_token_sql_query_cmd = [
             'psql', '-U', 'reana', 'reana', '-c',
             'SELECT access_token FROM user_']
         sql_query_result = exec_into_component(
-            get_prefixed_component_name('db', namespace),
+            get_prefixed_component_name('db'),
             get_admin_token_sql_query_cmd)
         # We get the token from the SQL query result
         admin_access_token = sql_query_result.splitlines()[2].strip()
--- a/reana/k8s_utils.py
+++ b/reana/k8s_utils.py
@@ -40,7 +40,7 @@ def exec_into_component(component_name, command):
     return output.decode('UTF-8')
 
 
-def get_external_url(insecure=False, namespace=None):
+def get_external_url(insecure=False):
     """Get external IP and port to access REANA API.
 
     :param insecure: Whether the URL should be insecure (http) or secure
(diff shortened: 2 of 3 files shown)


tiborsimko:
Seeing the same problem:

```console
$ DEMO=r-d-r-roofit EXCLUDE_COMPONENTS=r-ui,r-m-broker make ci
...
LAST DEPLOYED: Fri Apr 17 11:05:12 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Thanks for flying REANA 🚀
Environment variables could not be generated:
Command '['minikube', 'ip', '--profile', 'default']' returned non-zero exit status 65.
make[1]: Leaving directory '/home/simko/private/project/reana/src/reana'
```

If the motivation for the 95ebddb change was to easily test given instance with py27 client, what I've been doing is running `setup-environment` in the usual py38 venv environment, saving the output, and sourcing the export commands in the py27 venv.  Since we test py27 only occasionally, it seems to be sufficient.

reana/cli.py
```diff
@@ -1512,29 +1512,29 @@ def install_client():  # noqa: D301
...
+    help='Instance name which configuration should be produced.')
+def setup_environment(server_hostname, insecure_url, instance_name):  # noqa: D301
     """Display commands to set up shell environment for local cluster.
 
-    Display commands how to set up REANA_SERVER_URL and REANA_ACCESS_TOKEN
-    suitable for current local REANA cluster deployment. The output should be
-    passed to eval.
+    Display commands on how to set up REANA_SERVER_URL and REANA_ACCESS_TOKEN,
+    suitable for the default local REANA instance deployment (determined by the
+    virtuel environment name) or to the REANA instance specified with
```
tiborsimko, line 1522: `s/virtuel/virtual/g`

reana/cli.py
```diff
@@ -1512,29 +1512,29 @@ def install_client():  # noqa: D301
     '--insecure-url', is_flag=True,
     help='REANA Server URL with HTTP.')
 @click.option(
-    '--namespace',
-    default='default',
-    help='Namespace of the components which configuration should be produced.')
-def setup_environment(server_hostname, insecure_url, namespace):  # noqa: D301
+    '--instance-name',
+    help='Instance name which configuration should be produced.')
```
tiborsimko, line 1516: The help message is not clear, perhaps something like:  "REANA instance name" even though we might add a word about minikube profile name...

tiborsimko:
> Seeing the same problem:

Works now after you repushed :+1: 

I'm still not sure whether we need this in the codebase at all. If we have a general developer convention that `python_venv_name == minikube_profile_name == REANA_INSTANCE_NAME` (and also `== k8s_namespace_name` in the future), then we could test easily py27 clients like this:

```console
$ workon reana
$ reana-dev setup-environment  > /tmp/z.sh
$ workon reana-client-py27
$ source /tmp/z.sh
$ reana-client run -w roofit
```

For some occasional py27 work this should be sufficient?! FWIW that's how I've been debugging some py27 issues in the past. 

Generally speaking, having leaner code base and leaving extra options out is a good thing; the less code there is, the less chance for errors. WDYT?

diegodelemos:
I was also thinking of the following while I am developer and integrator:
```console
$ # `reana` instance is running a make ci with a set of PRs
$ # `reana-dev` I am developing a feature
$ reana-dev setup-environment reana # something went wrong, so I want to
$                                   # see the error and share with my teammate
$ reana-client logs -w helloworld-serial
$ reana-dev setup-environment reana-dev # I want to get back to my `reana-dev` and
$                                       # continue developing
$ reana-client run -w my-test-for-my-dev-change
```
If we don't want it we can remove it totally. In my opinion, it is useful to be able to change "from which instance" to get config from easily.

mvidalgarcia:
I don't have a strong opinion, I'd be fine if we remove it completely (including the click option that was there before! otherwise I wouldn't have bother in _fixing_ it). I'm fine with the formula that @tiborsimko  proposes about switching venv temporarily. Plus, 95% of the time I'll be working with `reana` venv.
I wonder if this can be more annoying for the use case that @diegodelemos described. It's not my case as I'm not using two minikube instances at the moment, so his opinion matters more.

TL;DR I'm fine with both options.

tiborsimko:
> I was also thinking of the following while I am developer and integrator:

For such context switching, (1) working in DEV on a new feature and (2) working in PROD on integrating many PRs, why not to use two different `tmux` sessions with two different Python virtual environments already setup? This allows to cleanly separate DEV context work from the PROD context work, which is a good thing, both for the mind and the technology. 

E.g. imagine that your DEV work touches `reana-client` or `reana-commons`; then you have to also run `workon reana` and `workon reana-dev` commands in your CLI scenario in order to switch DEV and PROD context fully so that you would have latest-DEV-client for DEV work and latest-PROD-client for PROD work.  So it starts to be suddenly more typing... Switching to a separate `tmux` session should be cleaner and quicker in general?

diegodelemos:
Agreed.

PR #303, opened 2020-04-20
helm: use default secrets values

diegodelemos:


diegodelemos:
Since this pattern will be reproduced often, we might profit of changing the GitHub action to trigger the release only when we push a tag? Currently, we do it with a merge to master and if there are changes under `helm/reana`... The [list of releases](https://github.com/reanahub/reana/releases) is rapidly growing.

PR #307, opened 2020-04-24
cli: adapt to new UserToken table

mvidalgarcia:
closes reanahub/reana-ui#93


mvidalgarcia:
Side-effect: https://travis-ci.org/github/reanahub/reana/jobs/680147358
`reana` installs `pytest-reana` which installs `reana-db` which has an f-string not compatible with python2

tiborsimko:
We still claim support for Python-2.7 in `reana`'s `setup.py`, but I think we can safely remove it. 

As you mentioned, it is only the user-facing modules such as `r-client`, `r-commons` that need to be Python-2 friendly for some time still.  Everything else can be Python-3 only.

PR #309, opened 2020-05-01
helm: add optional notifications

### LINKED ISSUE #308: cronjobs: regular cluster status summaries by email
Using REANA-Server's `flask reana-adimin report-status` collect system status and send them by email to the configured notification receiver in the time intervals defined by the administrator.

diegodelemos:
* Useful to program system wide status reports with cronjobs
  (closes #308).

### DIFF
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -8,7 +8,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 
 | Parameter                                                | Description                                                                          | Default value                                   |
 |----------------------------------------------------------|--------------------------------------------------------------------------------------|-------------------------------------------------|
-| `components.reana_db.enabled`                            | Instantiate a PostgreSQL database inside the cluster                                 | true                                        |
+| `components.reana_db.enabled`                            | Instantiate a PostgreSQL database inside the cluster                                 | true                                            |
 | `components.reana_job_controller.image`                  | [REANA-Job-Controller image](https://hub.docker.com/repository/docker/reanahub/reana-job-controller) to use  | `reanahub/reana-job-controller:<chart-release-verion>` |
 | `components.reana_message_broker.image`                  | [REANA-Meessage-Broker image](https://hub.docker.com/repository/docker/reanahub/reana-message-broker) to use | `reanahub/reana-message-broker:<chart-release-verion>` |
 | `components.reana_message_broker.imagePullPolicy`        | REANA-Message-Broker image pull policy                                               | IfNotPresent                                    |
@@ -33,6 +33,14 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `ingress.annotations.kubernetes.io/ingress.class`        | Type of ingress controller                                                           | traefik                                         |
 | `ingress.annotations.traefik.frontend.entryPoints`       | Entrypoints allowed by the ingress controller                                        | "http,https"                                    |
 | `ingress.enabled`                                        | Create an ingress resource to access the REANA instance from outside the cluster     | true                                            |
+| `notifications.email_config.login`                       | Login for the sender email address                                                   | None                                            |
+| `notifications.email_config.password`                    | Password for the sender email address                                                | None                                            |
+| `notifications.email_config.receiver`                    | Email address which will be receiving the notifications                              | None                                            |
+| `notifications.email_config.sender`                      | Email address which will be sending the notifications                                | None                                            |
+| `notifications.email_config.smtp_server`                 | SMTP email server host                                                               | None                                            |
+| `notifications.email_config.smtp_port`                   | SMTP email server port                                                               | None                                            |
+| `notifications.enabled`                                  | Enable REANA system events notifications                                             | false                                           |
+| `notifications.system_status`                            | Cronjob pattern representing how often the system status notification should be sent. Leave it empty to deactivate it | "0 0 * * *"                                     |
 | `secrets.cern.sso.CERN_CONSUMER_KEY`                     | CERN SSO consumer key                                                                | None                                            |
 | `secrets.cern.sso.CERN_CONSUMER_SECRET`                  | **[Do not use in production, use secrets instead]** CERN SSO consumer secret         | None                                            |
 | `secrets.database.pasword`                               | **[Do not use in production, use secrets instead]** PostgreSQL database password     | None                                            |
@@ -40,7 +48,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `secrets.gitlab.REANA_GITLAB_HOST`                       | Hostname of the GitLab instance                                                      | None                                            |
 | `secrets.gitlab.REANA_GITLAB_OAUTH_APP_ID`               | GitLab OAuth application id                                                          | None                                            |
 | `secrets.gitlab.REANA_GITLAB_OAUTH_APP_SECRET`           | **[Do not use in production, use secrets instead]** GitLab OAuth application secret  | None                                            |
-| `secrets.reana.REANA_SECRET_KEY`                         | **[Do not use in production, use secrets instead]** REANA encrypytion secret key     | None                                            |
+| `secrets.reana.REANA_SECRET_KEY`                         | **[Do not use in production, use secrets instead]** REANA encryption secret key     | None                                            |
 | `serviceAccount.create`                                  | Create a service account for the REANA system user                                   | true                                            |
 | `serviceAccount.name`                                    | Service account name                                                                 | reana                                           |
 | `serviceAccount.namespace`                               | Service account namespace                                                            | default                                         |
--- a/helm/reana/templates/NOTES.txt
+++ b/helm/reana/templates/NOTES.txt
@@ -1,1 +1,22 @@
+
+The REANA system has been installed:
+
+There are a few steps left to finalise its configuration
+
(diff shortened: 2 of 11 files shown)


helm/reana/README.md
```diff
@@ -33,6 +33,14 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `ingress.annotations.kubernetes.io/ingress.class`        | Type of ingress controller                                                           | traefik                                         |
 | `ingress.annotations.traefik.frontend.entryPoints`       | Entrypoints allowed by the ingress controller                                        | "http,https"                                    |
 | `ingress.enabled`                                        | Create an ingress resource to access the REANA instance from outside the cluster     | true                                            |
+| `notifications.email_config.login`                       | Password for the sender email address                                                | None                                            |
+| `notifications.email_config.password`                    | Password for the sender email address                                                | None                                            |
+| `notifications.email_config.receiver`                    | Email address which will be receiving the nofications                                | None                                            |
```
mvidalgarcia, line 38:
```suggestion
| `notifications.email_config.receiver`                    | Email address which will be receiving the notifications                                | None                                            |
```

helm/reana/README.md
```diff
@@ -33,6 +33,14 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `ingress.annotations.kubernetes.io/ingress.class`        | Type of ingress controller                                                           | traefik                                         |
 | `ingress.annotations.traefik.frontend.entryPoints`       | Entrypoints allowed by the ingress controller                                        | "http,https"                                    |
 | `ingress.enabled`                                        | Create an ingress resource to access the REANA instance from outside the cluster     | true                                            |
+| `notifications.email_config.login`                       | Password for the sender email address                                                | None                                            |
```
mvidalgarcia, line 36:
```suggestion
| `notifications.email_config.login`                       | Sender email address                                                | None                                            |
```

helm/reana/templates/reana-server.yaml
```diff
@@ -111,7 +111,25 @@ spec:
             value: ""
           - name: GIT_SSL_NO_VERIFY
             value: "true"
+          - name: REANA_EMAIL_SMTP_SERVER
```
mvidalgarcia, line 114: Missing, otherwise it uses default k8s env var.

```diff
+            {{- range $key, $value := .Values.db_env_config }}
+            - name: {{ $key }}
+              value: {{ $value | quote }}
+            {{- end }}
```

helm/reana/templates/cronjobs.yaml
```diff
@@ -0,0 +1,85 @@
...
+                  key: REANA_EMAIL_PASSWORD
+            {{- end }}
+            - name: REANA_COMPONENT_PREFIX
+              value: {{ include "reana.prefix" . }}
+            - name: REANA_EMAIL_SENDER
+              value: {{ .Values.notifications.email_config.sender }}
+            - name: REANA_ADMIN_ACCESS_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "reana.prefix" . }}-admin-access-token
```
mvidalgarcia, line 60: I think you've removed this secret, where is it supposed to get this value from?

diegodelemos: I think that it should be manually created, there is no way I can see we can create the secret with the correct value from the beginning, so this will have to go into docs or the message displayed just after installing REANA. The interaction could look like:
1. User installs REANA
2. Creates admin user and/or retrieves the admin token
3. Then creates secret manually

Or in commands:
```
$ helm install reana reanahub/reana --wait
NAME: reana
LAST DEPLOYED: Wed Mar 18 10:27:06 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:

Create your admin user with:
> kubectl exec -ti reana-server-xxxx-yyy -- flask reana-admin users-create-default user@my.org
<reana-admin-access-token>
> read -s REANA_ADMIN_ACCESS_TOKEN
> kubectl create secret generic <prefix>-admin-access-token --from-literal=ADMIN_ACCESS_TOKEN='$REANA_ADMIN_ACCESS_TOKEN'
$ kubectl exec -ti reana-server-xxxx-yyy -- flask reana-admin users-create-default
$ kubectl create secret generic reana-admin-access-token --from-literal=ADMIN_ACCESS_TOKEN='$REANA_ADMIN_ACCESS_TOKEN'
Thanks for flying REANA 🚀
```

Note: this will would, of course, automatise for developers.


What do you think about this? Otherwise, we create a secret with a value that makes no sense.

mvidalgarcia: I agree, I think it makes sense to create the admin user manually instead of imposing it in `/scripts/setup`

https://github.com/reanahub/reana-server/blob/a22b50263764467ebe8a8ec28284431b62205c12/scripts/setup#L14

The only problem I see is how to retrieve this generated token later, which it's essentially the same problem we have for `setup-environment`.

tiborsimko: :+1:  for allowing people to set their admin account details semi-manually instead of hard-coding values such as `info@reana.io` as in the old days.

We could have a helper script `reana/scripts/create-admin-user.sh` that would do all the steps in an unassisted way in order to simplify procedure.

- One option would to read the wanted email address and token, as you described.  However we should then make validation of the token values passed by the user, to make sure the values is compliant.

- Another option would be to use `uuidgen` or equivalent in the script to generate token automatically, to avoid any problems.  I'd like this option better. This would also simplify the interaction in that new admins would have to provide basically only their email, for example: 

```console
$ helm install reana reanahub/reana --wait
[INFO] ...
[INFO] ...
$ ./scripts/create-admin-user.sh john.doe@example.org
[INFO] Admin user john.doe@example.org created.
[INFO] Your REANA_ACCESS_TOKEN is aaaa-bbbb-1111-2222.
```

In the future, for systems with SSO login, this post-installation admin-creation part could be modified to e.g. require logging in via web, etc.

diegodelemos: > The only problem I see is how to retrieve this generated token later, which it's essentially the same problem we have for setup-environment.

I thought of this precisely to tackle the problem in this PR and the one you mention (https://github.com/reanahub/reana/pull/311#issue-419645624) as if this is done then we can instead of going to DB, instruct `reana-dev` to read it from the Kubernetes secrets store:
```
$ kubectl get secret -o json reana-dev-secrets | \
  jq -r .data.REANA_SECRET_KEY | base64 -D
secret_key
```
And that would solve it.

However, we shouldn't ignore that we would end up with the REANA admin access token secret in two places (DB and Kubernetes)... a duplication and all the problems it can potentially bring.

mvidalgarcia: This is practically what `reana-admin users-create-default` does:

```console
$ flask reana-admin users-create-default test@test.es
Created 1st user with access_token: fcFL7atgKmdD5LaP5wX1sln0GRJ97-MTfLHz2hz69ME
```

I would rename this command to something like `create-admin-user` and modify its output.

diegodelemos: I will do the fastest following what we discussed to get this going and perhaps we can ticketise https://github.com/reanahub/reana/pull/309#discussion_r427854190 as some things I believe would take more time to implement and test.

tiborsimko: Yes, `reana-server/script/setup` does that, and takes care of creating DB as well... I'm not sure it should be done there though. Following "do-one-thing-and-do-it-well" principle, let's reserve `reana-server` for handing REST API only, and let's finish post-installation-related setup steps outside, right after Helm finishes? This would better split responsibilities, and would allow for easy customisation and scriptability of various local installations. For example, we shall have Alembic upgrade recipes soon; where would we like to run those? This should be an interactive CLI process to warn admins of failures, etc. The `r-server` component is not a place for that, I think such scripts should rather live in the `reana` package, where admins are already helming and such. I guess one day soon we shall probably take both the admin generation and the DB creation out  of `r-server`.

helm/reana/templates/NOTES.txt
```diff
@@ -1 +1,22 @@
-Thanks for flying REANA 🚀
\ No newline at end of file
+
+The REANA system has been installed:
+
+There are a few steps left to finalise it's configuration
```
mvidalgarcia, line 4:
```suggestion
There are a few steps left to finalise its configuration
```

reana/cli.py
```diff
@@ -1524,15 +1526,12 @@ def setup_environment(server_hostname, insecure_url):  # noqa: D301
...
-        get_admin_token_sql_query_cmd = [
-            'psql', '-U', 'reana', 'reana', '-c',
-            'SELECT token FROM user_token']
-        sql_query_result = exec_into_component(
-            get_prefixed_component_name('db'),
-            get_admin_token_sql_query_cmd)
-        # We get the token from the SQL query result
-        admin_access_token = sql_query_result.splitlines()[2].strip()
+        get_access_token_cmd = \
+            'kubectl get secret -o json reana-dev-admin-access-token'
```
mvidalgarcia, line 1530: I think it should be `get_prefixed_component_name('admin-access-token')` or similar.

diegodelemos: Absolutely, thanks a lot for spotting it :)

PR #311, opened 2020-05-18
helm: add reana secret key 

mvidalgarcia:
closes reanahub/reana-db#66

⚠️ The [`reana-dev setup-environment` problem](https://github.com/reanahub/reana/blob/master/reana/cli.py#L1528-L1530) (encrypted token output) still need to be addressed or ticketized.

### DIFF
--- a/helm/reana/templates/reana-server.yaml
+++ b/helm/reana/templates/reana-server.yaml
@@ -93,6 +93,11 @@ spec:
               secretKeyRef:
                 name: {{ include "reana.prefix" . }}-cern-gitlab-secrets
                 key: REANA_GITLAB_HOST
+          - name: REANA_SECRET_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ include "reana.prefix" . }}-secrets
+                key: REANA_SECRET_KEY
           {{- if .Values.debug.enabled }}
           # Disable CORS in development environment, for example
           # to connect from an external React application.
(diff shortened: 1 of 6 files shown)


helm/reana/templates/reana-server.yaml
```diff
@@ -93,6 +93,11 @@ spec:
               secretKeyRef:
                 name: {{ include "reana.prefix" . }}-cern-gitlab-secrets
                 key: REANA_GITLAB_HOST
+          - name: REANA_SECRET_KEY
```
mvidalgarcia, line 96: Do I need to add this env also in `reana-workflow-controller.yaml`? I see there are other cases in both templates such us `REANA_DB_USERNAME`.

diegodelemos: In principle, we do not access `user.access_token` anywhere else than in RS, if SQLAlchemy doesn't need it (does lazy loading) we can skip adding it there (the fewer places we have this key loaded the best from a security point of view). I would say it needs testing.

helm/reana/values.yaml
```diff
@@ -32,6 +32,7 @@ secrets:
   gitlab: {}
   cern:
     sso: {}
+  reana: {}
```
diegodelemos, line 35: Can you add this new config to the [README](https://github.com/reanahub/reana/blob/master/helm/reana/README.md)?

helm/reana/templates/reana-server.yaml
mvidalgarcia: How would you test it?

Issue #304, opened 2020-04-21
make: move to shell scripts


tiborsimko:
**Problem**

Context:

- `Makefile` is only a tiny layer on top of `reana-dev` to automate the execution of some frequent commands to build, deploy, and develop REANA in Minikube.

Observations:

- Using `Makefile` can be sometime tedious. Example: when adding recent Helm features, an external script `scripts/update_images.sh` was created to update image tags.

- We don't have a lot of dependency between various steps, so using `Makefile` is not really necessary.

**Proposal**

Convert `Makefile` into a set of shell scripts that are easier to work with.  Using the same guiding principles as before:

- Use `reana-dev` to automatise as much as possible the hard logic. The build/deployment scripts should be only a tiny layer on top if it, basically mirroring what would execute in shell.
- Create several independent shell scripts, each having quite a good correspondence to current `Makefile` targets. For example `make build` can become `build-images.sh`, `make ci` can become `run-ci.sh`, etc.
- Use the usual shell environment variables, such as `DEMO=r-d-r-roofit EXCLUDE_COMPONENTS=r-ui,r-m-broker make ci` that can be directly reused.

Possible enhancements:

- Harmonise environment variable names to start by `REANA_...` so that one can easily find all local customisations via `env | grep REANA_` which ones are set.
- Use `REANA_INSTANCE_NAME` to select between DEV and PROD instances, where we keep convention `reana_instance_name == python_venv_name == minikube_profile_name`, as is the case now.  We can automatise setting variables via `~/.virtualenvs/${REANA_INSTANCE_NAME}/bin/preactivate` helper scripts if need be.
- Possible improvement: once #274 is implemented, we can also use `REANA_INSTANCE_NAME == k8s_namespace_name`. This would allow us to use one single Minikube instance to run both DEV and PROD clusters there.  (This would also allow to kill all the various `MINIKUBE_CPUS` custom variables, see https://github.com/reanahub/reana-cluster/issues/230.) Pros: using usual Minikube profile is familiar to people, and having one Docker image registry saves pulling etc time for base images.  Cons: both DEV and PROD instances would run on the same K8s version, so cannot easily test 1.15 and 1.18, say. (One would have to resort to two different Minikubes, based on different Kubernetes versions.)
- Scripts used to deploy REANA locally can be perhaps reused to deploy REANA on CERN cloud for "private" development or CI deployments, so having a set of shell scripts that are reusable via environment variables may be profitable for the future as well.

Example:

[...]

PR #319, opened 2020-05-27
Fix mail config

diegodelemos:


Makefile
```diff
@@ -172,7 +172,8 @@ release-build:
 	cd ../reana && \
 	echo "The following images have been built:" && \
 	./scripts/update_images.sh $$BUILT_IMAGES_FILE_PATH && \
-	rm $$BUILT_IMAGES_FILE_PATH
+	rm $$BUILT_IMAGES_FILE_PATH && \
+	pip install ../reana-client --upgrade
```
tiborsimko, line 176: Hmm, is single `pip` call enough? Do we need this? If so, we should probably pip uninstall a bunch, as we do in the dev build?  Anyway, we could also address this as part of #304.

helm/reana/templates/secrets.yaml
```diff
@@ -52,6 +52,6 @@ metadata:
     "helm.sh/resource-policy": keep
 type: Opaque
 data:
-  REANA_EMAIL_LOGIN: {{ .Values.notifications.email_config.login | b64enc }}
-  REANA_EMAIL_PASSWORD: {{ .Values.notifications.email_config.password | b64enc }}
+  REANA_EMAIL_LOGIN: {{ .Values.notifications.email_config.login | default "team@reana.io" | b64enc }}
+  REANA_EMAIL_PASSWORD: {{ .Values.notifications.email_config.password | default "123456" | b64enc }}
```
tiborsimko, line 56: Hmm, `123456` may evoke souvenirs in people. What about using `CHANGEME` or something as we discussed before?

PR #344, opened 2020-06-24
Kind multinode opt mount

diegodelemos:


helm/configurations/values-dev.yaml
```diff
@@ -1,4 +1,7 @@
 # REANA components pointing to `latest`, locally built master branch
+debug:
+  enabled: true
+
```
tiborsimko, line 4: This would not allow to test production-like deployment before release builds, isn't it? Things can behave differently when code is mounted, so it would be good to retain that option. I mean we have been distinguishing three "local" modes:

- production, for people testing out REANA, using tagged docker images from docker hub
- prod-like development, using `latest` images, but without code mounting
- debug-like development, using `latest` images, with code mounting

The second option allows to run CI tests before releasing images for example. (Or were you thinking of switching to  tagged builds for that mode?) 

We could solve this by introducing several `values.yaml` files:

- production, default `helm/reana/values.yaml`, interesting for people testing out REANA
- prod-like development, using `helm/configurations/kind-latest.yaml`
- debug-like development with live code mounts, using `helm/configurations/kind-latest-debug.yaml`

This would allow to have in the future more configurations, say `kind-latest-multi-node.yaml` and whatnot, or `-vc3.yaml` and whatnot...

The `reana-dev cluster-deploy` would then optionally take which configuration to deploy.

P.S. Can use "local-" instead of leading "kind-..." or other such suitable name...

Recall also old musings https://github.com/reanahub/reana/pull/289#discussion_r409441312

diegodelemos: Since `debug`/`dev` is a very clear and recurrent thing to do, I have added a `--debug` flag, just as in `reana-dev cluster-create` where `--debug` will mount the code transparently. This flag would mean in this context to simply add the `debug.enabled=true` to Helm install as a `--set key=value` option:
> production, for people testing out REANA, using tagged docker images from docker hub
1. Would be installing latest stable version from [our Helm repo](https://reanahub.github.io/reana/) because it wouldn't be safe to use the `values.yaml` as many times we update the templates but not the docker images, so we would break this mode.
> prod-like development, using latest images, but without code mounting
2. Running `reana-dev cluster-deploy`
> debug-like development, using latest images, with code mounting
3. Running `reana-dev cluster-deploy --debug`
> We could solve this by introducing several values.yaml files:
4. Additionally to the "basic" setup, we could add a `--file` or `--values` option so we have room for growing towards `*-vc3.yaml` etc....

Would it be of interest for this tool to encapsulate (1) as well?

reana/cli.py
```diff
@@ -1647,12 +1668,35 @@ def cluster_build():  # noqa: D301
...
-        "helm install reana helm/reana -f helm/configurations/values-dev.yaml -n default --wait",
+    dev_values_yaml = "helm/configurations/values-dev.yaml"
+    debug_flags = "--set debug.enabled=true" if debug else ""
+    helm_install = "helm install reana helm/reana -n {namespace} -f {values} --create-namespace {debug_flags} --wait".format(
+        namespace=namespace, values=dev_values_yaml, debug_flags=debug_flags
+    )
+    deploy_cmds = [
+        helm_install,
+        "kubectl config set-context --current --namespace={}".format(namespace),
+        "sleep 5",  # Let the database initialise.
```
diegodelemos, line 1691: @tiborsimko I had to add this as sometimes the `create-admin-user.sh` would fail with an OperationalError trying to connect to Postgres which was still being initialised, did you see this problem? The correct/long-term solution for this would be to add `livenessProbes` to PosgreSQL, or even better, use the [PostgreSQL Helm Chart](https://hub.helm.sh/charts/bitnami/postgresql/3.18.3) so we don't maintain the template.

tiborsimko: Haven't seen this with the "old kind" times; using `--wait` was always enough there.  We could make a mini sleep loop as in Makefile times until we go full pg helm...

helm/configurations/values-dev.yaml
tiborsimko: > Would it be of interest for this tool to encapsulate (1) as well?

We can leave it out indeed, you are right about template vs docker image updates... We could have another (simpler, non-reana-dev) CI sequence to test latest releases from the end user point of view. In this way `reana-dev` will remain for local developments only

reana/cli.py
```diff
@@ -1905,38 +1946,77 @@ def run_example(
 @click.option(
     "-m",
     "--mount",
+    "mounts",
+    multiple=True,
+    callback=volume_mounts_to_list,
+    help="Which local directories whould be mounted in the Kind nodes? local_path=reana_system_path",
```
tiborsimko, line 1952: Can you propagate the same `-m` new behaviour to the `run_ci()` command as well?  It would be practical for CI runs.

tiborsimko: (so that `-m` behaves the same for all commands it exists for...)

reana/cli.py
```diff
@@ -1905,38 +1946,77 @@ def run_example(
...
+)
+@click.option(
+    "--debug",
     is_flag=True,
     default=False,
-    help="Should we mount /var/reana from host?"
+    help="Should we mount the REANA source code for live code updates?",
+)
+@click.option(
+    "-n", "--worker-nodes", default=0, help="How many worker nodes? default 0"
```
tiborsimko, line 1961: Please change to `-w`, `--worker-nodes` (or for `-c`, `--count-worker-nodes`) so that `-n` will be reserved for `--namespace` everwhere.

tiborsimko: (or we just don't use short one-letter flag for this option?  Long is enough; the flag won't be used every day. In this way we don't have to invent too fantasy flag name just so that the leading letter would be free)

reana/cli.py
diegodelemos: Added a new tiny script `wait-for-db.sh`. I am totally flexible, it can be renamed or even, this logic could be added inside `create-admin-user.sh` as we do not need it anywhere else.

reana/cli.py
diegodelemos: Added. Note that due to https://github.com/helm/helm/issues/5618 we have to make the implementation a bit more convoluted. However, the final result is more streamlined with the way we do the `cluster-create`, where we print the config that is being applied, in this case the modified `values.yaml` content, so it turns out to be good in my opinion. 
<details><summary>Example output</summary>
<p>

```
$ reana-dev cluster-deploy -j /code:/var/code                                                                                                      [11:17:18] 
[2020-06-25T11:17:18] reana: cat <<EOF | helm install reana helm/reana -n default --create-namespace --wait -f -                                                                                                                      
components:                                                                                                        
  reana_job_controller:                                                                                                                                                                                                               
    image: reanahub/reana-job-controller                                                                           
  reana_message_broker:                                                                                                                                                                                                               
    image: reanahub/reana-message-broker                 
  reana_server:                                                                                                                                                                                                                       
    image: reanahub/reana-server                                                                                   
  reana_ui:                                                                                                        
    image: reanahub/reana-ui                                                                                       
  reana_workflow_controller:                                                                                       
    environment:                                                                                                   
      REANA_JOB_HOSTPATH_MOUNTS: '[{"name": "var-code", "hostPath": "/code", "mountPath":                          
        "/var/code"}]'                                                                                             
    image: reanahub/reana-workflow-controller                                                                      
  reana_workflow_engine_cwl:                                                                                       
    image: reanahub/reana-workflow-engine-cwl                                                                      
  reana_workflow_engine_serial:                                                                                                                                                                                                       
    image: reanahub/reana-workflow-engine-serial                                                                                                                                                                                      
  reana_workflow_engine_yadage:                                                                                    
    image: reanahub/reana-workflow-engine-yadage                                                                   
notifications:                                                                                                                                                                                                                        
  email_config:                                                                                                    
    receiver: team@reana.io                                                                                        
    sender: notifications@reana.io                                                                                                                                                                                                    
  enabled: true                                                                                                                                                                                                                       
  system_status: '*/5 * * * *'    

NAME: reana                                                                                                        
LAST DEPLOYED: Thu Jun 25 11:17:19 2020                                                                                                                                                                                               
NAMESPACE: default                                                                                                                                                                                                                    
STATUS: deployed                                                                                                   
REVISION: 1                                                                                                                                                                                                                           
TEST SUITE: None                                                                                                                                                                                                                      
NOTES:                                                                                                             
The REANA system has been installed:                                                                               
                                                                                                                   
If you are installing REANA for the first time, there are a few steps left to finalise its configuration.          
                                                                                                                   
1. Get the REANA-Server pod name:                                                                                  
                                                                                                                   
    $ REANA_SERVER=$(kubectl get pod -l "app=reana-server" -o name -o jsonpath='{.items[0].metadata.name}')                                                                                                                           
                                                                                                                   
2. Initialise the database:                                                                                        
                                                                                                                   
    $ kubectl exec $REANA_SERVER -- ./scripts/setup                                                                                                                                                                                   
                                                                                                                                                                                                                                      
3. Create your administrator user and store the token:                                                                                                                                                                                

    $ kubectl exec $REANA_SERVER -- flask reana-admin create-admin-user john.doe@example.org
    <reana-admin-access-token-value>
    $ read -s REANA_ADMIN_ACCESS_TOKEN  # paste the secret here
    $ kubectl create secret generic reana-admin-access-token \
      --from-literal=ADMIN_ACCESS_TOKEN="$REANA_ADMIN_ACCESS_TOKEN"

Thanks for flying REANA 🚀
[2020-06-25T11:17:33] reana: kubectl config set-context --current --namespace=default
Context "kind-kind" modified.
[2020-06-25T11:17:33] reana: /Users/rodrigdi/code/reanahub/reana/scripts/wait-for-db.sh
reana-db-f5494cb59-rbg77
Waiting for REANA-DB to be ready.
Waiting for REANA-DB to be ready.
Waiting for REANA-DB to be ready.
REANA-DB ready
[2020-06-25T11:17:49] reana: /Users/rodrigdi/code/reanahub/reana/scripts/create-admin-user.sh
/Users/rodrigdi/code/reanahub/reana/scripts/create-admin-user.sh: line 18: not: command not found
Defaulting container name to rest-api.
Use 'kubectl describe pod/reana-server-85bd5d6448-x6mjm -n default' to see all of the containers in this pod.
Creating REANA database...
DB Created.
REANA database created.
Creating Invenio database...
/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '

/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '
/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '

Creating database postgresql+psycopg2://reana:reana@reana-db.default.svc.cluster.local:5432/reana
/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '

/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '

Creating all tables!
INFO  [alembic.runtime.migration] Context impl PostgresqlImpl.
INFO  [alembic.runtime.migration] Will assume transactional DDL.
Created all tables!
Invenio database created.
Defaulting container name to rest-api.
Use 'kubectl describe pod/reana-server-85bd5d6448-x6mjm -n default' to see all of the containers in this pod.
/usr/local/lib/python3.6/site-packages/flask_sqlalchemy/__init__.py:834: FSADeprecationWarning: SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default in the future.  Set it to True or False to su
ppress this warning.
  'SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and '
secret/reana-admin-access-token created                                                                                                                                                                                                    
```


</p>
</details>

reana/cli.py
tiborsimko: Hmm, do we need an "external" shell script?  Perhaps doing the loop inside `reana-dev` Python code in this very place may be enough? 

Yeah, adding it to `create-admin-user.sh` is also a good option... As a kind of first checking whether the DB pod is up.  (Note that I had to add the Running pod test earlier there.)

reana/cli.py
```diff
@@ -1675,9 +1768,9 @@ def cluster_undeploy():  # noqa: D301
 @click.option(
     "-m",
     "--mount",
-    is_flag=True,
-    default=False,
-    help="Should we mount /var/reana from host?"
+    "mounts",
+    multiple=True,
+    help="Which local directories would be mounted in the Kind nodes? local_path=reana_system_path",
```
tiborsimko, line 1773: The syntax is actually `local_path:cluster_node_path` using `:`, not `=`.

(probably ditto for job mounts)

reana/cli.py
```diff
@@ -1636,23 +1657,95 @@ def client_uninstall():  # noqa: D301
...
+            job_mount_config.append(
+                {
+                    "name": mount["containerPath"].replace("/", "-")[1:],
+                    "hostPath": mount["hostPath"],
+                    "mountPath": mount["containerPath"],
+                }
+            )
+
+        if job_mount_config:
+            job_mount_config_string = json.dumps(job_mount_config)
```
tiborsimko, line 1715: `job_mount_config_string` is assigned, but never used
￼

reana/cli.py
```diff
@@ -1636,23 +1657,95 @@ def client_uninstall():  # noqa: D301
...
+        else:
+            job_mount_config = ""
```
tiborsimko, line 1717: `job_mount_config` is initialised as an empty list, here it is empty string...

reana/cli.py
```diff
@@ -1636,23 +1657,95 @@ def client_uninstall():  # noqa: D301
...
-def cluster_deploy():  # noqa: D301
+@click.option(
+    "--namespace", "-n", default="default", help="Kubernetes namespace [default]"
+)
+@click.option(
+    "-j",
+    "--job-mounts",
+    multiple=True,
+    callback=volume_mounts_to_list,
+    help="Which directories from the Kubernetes nodes to mount inside the job pods? kubernetes_node_path=inside_job_mountpath",
```
tiborsimko, line 1691: - `cluster_node_path:job_pod_mountpath`

- perhaps we should give concrete examples using ` /var/reana/mydata:/mydata` as in reana-commons?

reana/cli.py
```diff
@@ -1636,23 +1657,95 @@ def client_uninstall():  # noqa: D301
...
+
+    job_mount_config = job_mounts_to_config(job_mounts)
+    if job_mount_config:
+        values_dict["components"]["reana_workflow_controller"]["environment"] = {
+            "REANA_JOB_HOSTPATH_MOUNTS": json.dumps(job_mount_config)
+        }
+    if debug:
+        values_dict["debug"]["enabled"] = True
+
+    helm_install = "cat <<EOF | helm install reana helm/reana -n {namespace} --create-namespace --wait -f -\n{values}".format(
```
tiborsimko, line 1734: FWIW seeing sometimes a message about EOF:

```
notifications:
  email_config:
    receiver: team@reana.io
    sender: notifications@reana.io
  enabled: true
  system_status: '*/5 * * * *'

/bin/sh: line 26: warning: here-document at line 0 delimited by end-of-file (wanted `EOF')

NAME: reana
LAST DEPLOYED: Thu Jun 25 14:06:13 2020
NAMESPACE: default
```

but the deployment works fine

reana/cli.py
```diff
@@ -1687,15 +1780,16 @@ def cluster_undeploy():  # noqa: D301
     help="Destroy and recreate cluster from scratch?",
 )
 @cli.command(name="run-ci")
-def run_ci(mount, recreate):  # noqa: D301
+def run_ci(mounts, recreate):  # noqa: D301
```
tiborsimko, line 1783: Since the mount options are only used during cluster creation time, one cannot easily change mount? do-not-mount? decision once the cluster is created.  So perhaps we can move the recreate command to cluster-create so that we don't have to duplicate all these arguments?

This would mean people will have to do:

```console
$ reana-dev cluster-create
$ reana-dev run-ci
$ reana-dev cluster-delete
``` 

manually.

If it seems OK, then we should do:

- enrich `cluster-create` so that it does docker-pulling and kind-load-docker-image steps after cluster creation;
- leave out cluster-delete step, people will have to do it manually
- remove mount, recreate option from run-ci, because there were consumed already
- enrich run-ci options with --debug flag so that docker images would be built with DEBUG flag
```

File: reana/cli.py [hunk 11]
```diff
@@ -1636,23 +1657,98 @@ def client_uninstall():  # noqa: D301
...
+        for mount in job_mounts:
+            job_mount_list.append(
+                {
+                    "name": mount["containerPath"].replace("/", "-")[1:],
+                    "hostPath": mount["hostPath"],
+                    "mountPath": mount["containerPath"],
+                }
+            )
+
+        if job_mount_config:
```
tiborsimko, line 1715: if job_mount_list:

tiborsimko: (...this condition will never be true...)

PR #345, opened 2020-06-29
kind: full CI

tiborsimko:


reana/cli.py
```diff
@@ -1984,18 +2129,25 @@ def run_example(
...
+    help="Which local directories to mount in the cluster nodes? [local_path:cluster_node_path]",
 )
 @click.option(
     "--debug",
     is_flag=True,
     default=False,
     help="Should we mount the REANA source code for live code updates?",
 )
-@click.option("--worker-nodes", default=0, help="How many worker nodes? default 0")
+@click.option("--worker-nodes", default=0, help="How many worker nodes? [default=0]")
```
diegodelemos, line 2140: I am wondering whether it would make sense to check whether `--mount` was passed whenever the user selects 2 or more nodes, otherwise REANA won't work. What do you think?

tiborsimko: OK, I can add it.  (And also delete the dangling image commit...)

diegodelemos:
The hardcoded `return True` in REANA-Commons shouldn't be necessary anymore using https://github.com/reanahub/reana-commons/pull/198.

PR #347, opened 2020-07-03
helm: add option to split load by node affinity

diegodelemos:


helm/reana/README.md
```diff
@@ -49,7 +49,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `notifications.system_status`                            | Cronjob pattern representing how often the system status notification should be sent. Leave it empty to deactivate it | "0 0 * * *"                                     |
 | `reana_url`                                              | REANA URL host                                                                       | None                                            |
-| `default_runtime_namespace`                              | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | None                                            |
+| `runtime_namespace`                                      | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | `.Release.Namespace`                            |
+| `dedicated_nodes`                                        | Should the infrastructure and runtime pods be splitted by node? (all nodes in the cluster should be labeled either as `reana.io/system=infrastructure` or `reana.io/system=runtime`)   | False |
```
mvidalgarcia, line 53:
```suggestion
| `dedicated_nodes`                                        | Should the infrastructure and runtime pods be split by node? (all nodes in the cluster should be labeled either as `reana.io/system=infrastructure` or `reana.io/system=runtime`)   | False |
```

helm/reana/templates/reana-cache.yaml
```diff
@@ -33,3 +33,5 @@ spec:
         image: redis:5.0.5
         ports:
         - containerPort: 6379
+      nodeSelector:
+        reana.io/system: infrastructure
```
mvidalgarcia, line 37: Endline missing

helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -203,3 +203,5 @@ spec:
         hostPath:
           path: /code/reana-workflow-controller
       {{- end }}
+      nodeSelector:
+        reana.io/system: infrastructure
```
mvidalgarcia, line 207: Endline missing

helm/reana/templates/roles.yaml
```diff
@@ -32,8 +32,8 @@ subjects:
 - kind: ServiceAccount
   name: {{ include "reana.prefixed_infrastructure_svaccount_name" . }}
   namespace: {{ .Release.Namespace }}
-{{- if .Values.default_runtime_namespace }}
+{{- if .Values.runtime_namespace }}
 - kind: ServiceAccount
   name: {{ include "reana.prefixed_runtime_svaccount_name" . }}
-  namespace: {{ .Values.default_runtime_namespace }}
+  namespace: {{ .Values.runtime_namespace }}
 {{- end }}
```
mvidalgarcia, line 39: Same

helm/reana/README.md
```diff
@@ -49,7 +49,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `notifications.system_status`                            | Cronjob pattern representing how often the system status notification should be sent. Leave it empty to deactivate it | "0 0 * * *"                                     |
 | `reana_url`                                              | REANA URL host                                                                       | None                                            |
-| `default_runtime_namespace`                              | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | None                                            |
+| `runtime_namespace`                                      | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | `.Release.Namespace`                            |
+| `dedicated_nodes`                                        | Should the infrastructure and runtime pods be split by node? (all nodes in the cluster should be labeled either as `reana.io/system=infrastructure` or `reana.io/system=runtime`)   | False |
```
tiborsimko, line 53: Tested:

- option off, no node labelled, expected pass -> OK

- option on, no node labelled, expect no deployment -> OK

- option on, one node labelled `reana.io/system=infrastructure`, expected deployment -> OK

- option on, one node labelled `reana.io/system=infrastructure`, no node labelled `reana.io/system=runtime`, expected workflow not to run -> but it did

- option on, one node labelled  `reana.io/system=infrastructure`, another node labelled ` `reana.io/system=runtime`, expected workflow to run on node two -> OK

So the feature work well overall, except for two comments:

(1) Unexpected success in the fourth case. Shouldn't we require some nodes to be labelled `reana.io/system=runtime` as per the docs? 

(2) BTW, instead of setting `dedicated_nodes: true` as a boolean variable to govern this behaviour, what about rather using two variables, something like:

```
node_label_infrastructure = reana.io/system=infrastructure
node_label_runtime = reana.io/system=runtime
```

so that people can use any labels they wish? E.g. they could be different, e.g. they could be equal, e.g. they could be "none" by default, which would switch this feature off. WDYT? Talk IRL.

helm/reana/README.md
```diff
@@ -40,6 +40,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `ingress.annotations.kubernetes.io/ingress.class`        | Type of ingress controller                                                           | traefik                                         |
 | `ingress.annotations.traefik.frontend.entryPoints`       | Entrypoints allowed by the ingress controller                                        | "http,https"                                    |
 | `ingress.enabled`                                        | Create an ingress resource to access the REANA instance from outside the cluster     | true                                            |
+| `node_label_infrastructure`                              | Define the label which identifies the nodes where the infrastructure pods should run, e.g. `reana.io/system=infrastructure` | None |
+| `node_label_runtime`                                     | Define the label which identifies the nodes where the runtime pods should run, e.g. `reana.io/system=runtime` (note that if `runtime_namespace` is not set, this option will automatically enable it with the value `reana-runtime`) | None |
```
tiborsimko, line 44: Hmm, do we need to use different namespaces if we go for separate infrastructure labels and runtime labels? 

- If not, it would be good to keep the concepts separate, to allow people to use the default namespace (or the reana namespace) and still have an option to specify two labels to mark various nodes for various purposes;

- If the options are intertwined, perhaps we can advertise only one settings, e.g. please select namespace_infrastructure and namespace_runtime, and if they have different values, please label nodes etc etc.

Just trying to prevent any surprises if people would do `helm install -n mynamespace reana -f myvalues.yaml` manually...

helm/reana/templates/serviceaccount.yaml
```diff
@@ -3,19 +3,21 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "reana.prefixed_infrastructure_svaccount_name" . }}
-  namespace: {{ .Release.Namespace }}
-{{- if .Values.default_runtime_namespace }}
```
diegodelemos, line 7: `namespace_runtime`

helm/reana/README.md
```diff
@@ -40,6 +40,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `ingress.annotations.kubernetes.io/ingress.class`        | Type of ingress controller                                                           | traefik                                         |
 | `ingress.annotations.traefik.frontend.entryPoints`       | Entrypoints allowed by the ingress controller                                        | "http,https"                                    |
 | `ingress.enabled`                                        | Create an ingress resource to access the REANA instance from outside the cluster     | true                                            |
+| `node_label_infrastructure`                              | Define the label which identifies the nodes where the infrastructure pods should run, e.g. `reana.io/system=infrastructure`. By default infrastructure pods can be scheduled to any availbable node in the cluster. | None |
+| `node_label_runtime`                                     | Define the label which identifies the nodes where the runtime pods should run, e.g. `reana.io/system=runtime`. By default runtime pods can be scheduled to any availbable node in the cluster. | None |
```
tiborsimko, line 44: `s/availbable/available/g`

helm/reana/README.md
diegodelemos: Solved as node labels and namespaces are not related anymore (the previous approach used [PodNodeSelector](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#podnodeselector) assigning the node selector to a Namespace, but this assumption was wrong because: (1) normally computer centres the creation of namespaces is not allowed (2) this feature wasn't working on Kind which means we cannot test it locally.

helm/reana/README.md
diegodelemos: Applied, thanks for the suggestion.

PR #349, opened 2020-07-08
cli: adapt unit tests to Postgres db container

### LINKED ISSUE #348: tests: move db tests to postgres container
**Current behavior**
All tests are run using an SQLite DB. Due to the [inclusion of DB schemas](https://github.com/reanahub/reana-db/pull/73) in our models, we need to switch to a different DB engine.

**Desired behavior**
Start using a Postgres docker container (since is the one used in production) as a DB for the tests. Check [cookiecutter-invenio-instance/run-tests.sh](https://github.com/inveniosoftware/cookiecutter-invenio-instance/blob/master/run-tests.sh) for reference.

mvidalgarcia:
closes #348

### DIFF
--- a/reana/cli.py
+++ b/reana/cli.py
@@ -627,6 +627,20 @@ def is_component_runnable_example(component):
     return False
 
 
+def does_component_need_db(component):
+    """Return whether the component needs DB to run tests.
+
+    Useful to determine which components need a Postgres DB container to run the tests.
+
+    :param component: standard component name
+    :type component: str
+
+    :return: True/False whether the component needs DB
+    :rtype: bool
+    """
+    return component in (COMPONENTS_USING_SHARED_MODULE_DB + ["reana-db"])
+
+
 def construct_workflow_name(example, workflow_engine):
     """Construct suitable workflow name for given REANA example.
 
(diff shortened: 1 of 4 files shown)


reana/cli.py
```diff
@@ -1923,6 +1937,13 @@ def python_unit_tests(component):  # noqa: D301
             cmd_activate_venv = "source  ~/.virtualenvs/_{}/bin/activate".format(
                 component
             )
+            if does_component_need_db(component):
+                run_command(
+                    "docker stop reana-postgres\n"
```
tiborsimko, line 1942: In case people run several commands:

```console
$ reana-dev python-unit-tests -c r-j-controller
$ reana-dev python-unit-tests -c r-server
```

and in case some of these processes hang, it'll good to use different DB containers.

What about giving them the same name as for Python venv? (like `__reana-job-controller`)

reana/cli.py
```diff
@@ -1934,10 +1955,21 @@ def python_unit_tests(component):  # noqa: D301
...
+                    if does_component_need_db(component)
+                    else "",
+                ),
                 "rm -rf ~/.virtualenvs/_{}".format(component),
             ]:
                 run_command(cmd, component)
+
+            if does_component_need_db(component):
+                run_command(
+                    "echo '==> [INFO] Stopping DB container...' && "
```
tiborsimko, line 1970: The `echo` statements should not be needed, the `reana-dev` will print each command that is run via `run_command()` automatically.

reana/cli.py
```diff
@@ -627,6 +627,20 @@ def is_component_runnable_example(component):
     return False
 
 
+def does_component_need_db(component):
```
tiborsimko, line 630: Can you add a tiny test case for this new function, similarly to what we have for others?

reana/cli.py
mvidalgarcia: > docker: Error response from daemon: Invalid container name (_reana-server), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.

mvidalgarcia: `reana-job-controller__` can be used though.

tiborsimko: ... or just start the container name with "postgres" which would also better indicate the nature of the running container... giving `postgres__reana-job-controller`.

BTW we use `postgres:9.6.2` with pinned version when installing REANA, so let's use the same pinned version for the tests too.

PR #360, opened 2020-08-04
global: split cli.py into different modules

diegodelemos:
* `cli.py` was getting very big and unmanageable, around 2500 LOC,
  since we expect the tool to grow quickly we are splitting it
  into several modules.


tiborsimko:
Looking at `rg '\bdef ' reana/*.py`, here are some musings:

- `kind_load_docker_image()` to its own kind command section?
- ditto for `kubectl_delete_pod()` to its own kubectl command section?
- `reana/ci.py` into `reana/run.py` to have the analogously named run command section?
- otherwise python, docker, git, cluster, client sections look good
- utils contains a lot of stuff, could be split to each sections, for example:
   - `is_component_python_package()` could be either living in python command file, or in a newly created components command file.  It is better to use `reana/components.py` instead of too generically named `reana/utils.py`.  
   - ditto for `get_current_commit()` or `get_all_branches()` that should rather live in git command file, or in git utils... 
   - ... and there are more of this kind in utils.py. Just looking to make the separation stricter.

tiborsimko:
BTW, since Git cannot track authorship properly while moving so many lines around, please use `--author "REANA team <team@reana.io>"` for this refactoring, as we did for Black.

tiborsimko:
Tested a few scenarios, everything looked OK.

One more comment: in order to retain an option to grow the `reana` repo into doing more stuff than it is doing now, shouldn't we put all the files in a reana-dev motivated subdirectory? E.g. instead of current: 

```console
$ git ls-files reana
reana/__init__.py
reana/cli.py
reana/client.py
reana/cluster.py
reana/config.py
reana/docker.py
reana/git.py
reana/kind.py
reana/kubectl.py
reana/python.py
reana/run.py
reana/utils.py
reana/version.py
```

we could use something like:

```console
reana/cli/cluster.py
reana/cli/...
```

similarly as we do for `reana-client`?

Or perhaps even into `reana-dev` motivated directory such as:

```console
reana/reana_dev/cluster.py
```

to have a very explicit `reana_dev` attribution?

diegodelemos:
I agree. I am amending it to `reana/cli/*.py ` as every REANA repository that has a CLI has either a directory `cli/` or a `cli.py`, so I believe it is more consistent. Moreover, if for some reason we end up having several CLI's we can do `reana/cli/reana-dev/*.py`, `reana/cli/reana-admin/*.py` etc..

tiborsimko:
Note the trouble when installing in production mode:

```console
$ reana-dev --help
  File "/home/simko/.virtualenvs/reana/bin/reana-dev", line 5, in <module>
    from reana.cli.cli import reana_dev
ModuleNotFoundError: No module named 'reana.cli'
```

> Moreover, if for some reason we end up having several CLI's we can do reana/cli/reana-dev/*.py, reana/cli/reana-admin/*.py etc..

Hypothetically, the other way would be better:

```
   reana/reana_dev/cli/*
   reana/reana_client/cli/*
   reana/reana_admin/ui/*
   reana/reana_admin/config/*
   reana/reana_admin/cli/*
```

diegodelemos:
> Note the trouble when installing in production mode:
> ...

and

> Hypothetically, the other way would be better:
> ...

Addressed [here](https://github.com/reanahub/reana/compare/7e8e85d05df5fec71926c8f1edf83f133edcbf08..d825f125deda7a5d012d0719cba7f81c594f1b02).

PR #364, opened 2020-08-05
cli: add git-merge command

### LINKED ISSUE #362: cli: git-merge
Introduce `git-merge` command that would complement `git-checkout` command. 

Its purpose is to merge a set of PRs to `master` and possibly to push

Example workflow: 

The integrator checks a set of PR:

```console
$ reana-dev git-checkout --fetch \
                         -b r-w-controller 401 \
                         -b r-j-controller 302 \
```

When everything is working, the integrator would like to merge some of them to master:

```console
$ reana-dev git-merge -b r-w-controller 401 -b r-j-controller 302 [--push]
```

This command should do, for each given repo:

- git fetch upstream in given repositories to check whether local `pr-401` branch is identical to `upstream/pr/401` to prevent errors;
- if we are up to date, do `git merge --ff-only upstream/pr/401` to `master`
- then delete locally checked out branch `pr-401`
- optionally, when `--push` argument is used, the new `master` branch should be pushed to `upstream` and `origin`;
- note that we already have `git-push` command which could be also amended for that effect (to add an option where to push, `origin` or `upstream`)

For example, I have been using the following alias:

```
merge-push-delete-pr = !sh -c 'git fetch upstream && git checkout master && git merge --ff-only upstream/master && git merge --ff-only upstream/pr/$1 && git branch -d pr-$1 && git push upstream master && git push origin master' -
```

The command would do something similar, but for a set of REANA components associated to a PR group.

Note: we could use simple git commands for now and investigate moving to `gh` command later (starting from `git-fork` and friends).

mvidalgarcia:
closes #362

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -448,7 +448,7 @@ def git_branch(component):  # noqa: D301
 
 
 @click.option(
-    "--branch", "-b", nargs=2, multiple=True, help="Which PR? [number component]"
+    "--branch", "-b", nargs=2, multiple=True, help="Which PR? [component PR#]"
 )
 @click.option("--fetch", is_flag=True, default=False)
 @git_commands.command(name="git-checkout")
@@ -465,7 +465,7 @@ def git_checkout(branch, fetch):  # noqa: D301
                    72`` will create a local branch called ``pr-72`` in the
                    reana-workflow-controller source code directory.
     :param fetch: Should we fetch latest upstream first? [default=False]
-    :type component: str
+    :type branch: list
     :type fetch: bool
     """
     for cpr in branch:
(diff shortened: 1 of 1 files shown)


reana/reana_dev/git.py
```diff
@@ -482,6 +482,52 @@ def git_checkout(branch, fetch):  # noqa: D301
...
+    :type branch: list
+    :type push: bool
+    """
+    for cpr in branch:
+        component, pull_request = cpr
+        component = select_components([component,])[0]
+        if component in REPO_LIST_ALL:
+            for cmd in [
+                "git fetch upstream",
+                "git diff pr-{0}..upstream/pr/{0}".format(pull_request),
```
diegodelemos, line 512: Just a potentially important comment: This won't fail if there is a difference right? Shouldn't it? It would mean we are merging new changes that weren't tested before.

tiborsimko: We should use `--exit-code` option to make sure we won't continue if there are differences. 

(integrator would then update local `pr-N` branch, restest, and continue)

reana/reana_dev/git.py
```diff
@@ -482,6 +482,52 @@ def git_checkout(branch, fetch):  # noqa: D301
...
+                "git checkout master",
+                "git merge --ff-only upstream/master",
+                "git merge --ff-only upstream/pr/{0}".format(pull_request),
+                "git branch -D pr-{0}".format(pull_request),
```
tiborsimko, line 516: Is `-d` working here? Would be better.

(If `pr-n` is equal to `upstream/pr/n`, which is ensured above, then `git branch -d` should be working... Using `-D` is dangerous if some remote updates or local updates get in.)

reana/reana_dev/git.py
diegodelemos: Perhaps we should return exactly that as an advice message on how to act upon the error?

mvidalgarcia:
Updated!

PR #371, opened 2020-08-06
cli: add git-upgrade-shared-modules command

### LINKED ISSUE #365: cli: git-upgrade-common-modules
Introduce `git-upgrade-common-modules` command that would look at which common modules are used in a component and would upgrade the versions.

Example workflow:

The integrator checks a set of PR:

```
$ reana-dev git-checkout --fetch \
						 -b r-w-controller 155 \
						 -b r-j-controller 283 \
						 -b r-commons 401 \
						 -b r-db 302
```

After r-commons and r-db are merged and released as X.YaN, we need to bump their minimal versions in r-w-controller and r-j-controller to that version.

A new command could do it:

```
$ reana-dev git-upgrade-common-modules \
						 -c r-w-controller
						 -c r-j-controller
```

For each component such as r-w-controller, the command would do:

- check which shared modules are used (e.g. r-db) and for each shared module do the following:
- check whether the latest commit in r-db directory is a release one with existing git tag; if not warn and abort
- check whether the current branch in r-w-controller is a master branch, if it is, warn and abort
- if we have a feature branch checked out, bump `setup.py` r-db version, bump `requirements.txt` r-db version, and amend last commit
- show git diff, ask for confirmation (at least until the feature is fully developed and trusted), and push to feature pr branch

The integrator would then perform run-ci and merge the rest via another new `git-merge` command that will be ticketised separately.

P.S. The behaviour could be influenced by options such as:

- `--use-latest-known-tag` which would skip the first test (whether r-db has release commit the last)
- `--create-new-version-bump-commit` which would not amend, but create new commit, in the second test

P.S. See also the old pre-refactoring command that was bumping shared module version everywhere; this one does similar job, but is more nuanced.

P.S. Note for the integrator: if r-commons is bumped, `-c r-client` should be probably added. Something not to forget about. So the default would be `-c CLUSTER -c CLIENT` perhaps if no `-c` is passed.

mvidalgarcia:
closes #365

##### TODO
- [x] Update docstrings
- [x] Find a better/different name for _old_ `git-upgrade-shared-modules` (the one that creates PRs)
- [x] Verify that both work as expected

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -23,9 +23,11 @@ from reana.config import (
 )
 from reana.reana_dev.utils import (
     display_message,
+    fetch_latest_pypi_version,
     get_srcdir,
     run_command,
     select_components,
+    update_module_in_cluster_components,
 )
 
 
@@ -724,7 +726,7 @@ def git_push(component):  # noqa: D301
             run_command(cmd, component)
 
 
-@git_commands.command(name="git-shared-modules-upgrade")
+@git_commands.command(name="git-upgrade-all-shared-modules")
 @click.option(
     "--commit-and-publish",
     is_flag=True,
@@ -734,33 +736,11 @@ def git_push(component):  # noqa: D301
     ' "installation: shared packages version bump" message will'
     " be created for all affected CLUSTER components.",
 )
-def git_shared_modules_upgrade(commit_and_publish):
-    """Upgrade all cluster components to latest REANA-Commons version."""
+def git_upgrade_all_shared_modules(commit_and_publish):
+    """Upgrade all cluster components to latest REANA-Commons/DB version.
 
-    def _fetch_latest_pypi_version(package):
-        """Fetch latest released version of a package."""
-        import requests
-
-        pypi_rc_info = requests.get(
-            "https://pypi.python.org/pypi/{}/json".format(package)
-        )
-        return sorted(pypi_rc_info.json()["releases"].keys())[-1]
-
-    def _update_module_in_cluster_components(module, new_version):
-        """Update the specified module version in all affected components."""
-        components_to_update = {
-            "reana-commons": COMPONENTS_USING_SHARED_MODULE_COMMONS,
-            "reana-db": COMPONENTS_USING_SHARED_MODULE_DB,
-        }
-        for component in components_to_update[module]:
-            update_setup_py_dep_cmd = (
-                'LN=`cat setup.py | grep -n -e "{module}.*>=" | cut -f1 -d: `'
-                '&& sed -i.bk "`echo $LN`s/>=.*,</>={new_version},</" '
-                "setup.py && rm setup.py.bk".format(
-                    module=module, new_version=new_version
-                )
-            )
-            run_command(update_setup_py_dep_cmd, component)
+    Optionally create commits and PRs in all components bumping version.
+    """
 
     def _commit_and_publish_version_bumps(components):
         """Create commit and version bump PRs for all specified components."""
@@ -791,15 +771,8 @@ def git_shared_modules_upgrade(commit_and_publish):
     ).union(set(COMPONENTS_USING_SHARED_MODULE_COMMONS))
 
     for module in REPO_LIST_SHARED:
-        last_version = _fetch_latest_pypi_version(module)
-        _update_module_in_cluster_components(module, last_version)
-        click.secho(
-            "✅ {module} updated to: {last_version}".format(
-                module=module, last_version=last_version
-            ),
-            bold=True,
-            fg="green",
-        )
+        last_version = fetch_latest_pypi_version(module)
+        update_module_in_cluster_components(module, last_version)
 
     if commit_and_publish:
         click.secho("Creating version bump commits and pull requests...")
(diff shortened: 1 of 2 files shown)


reana/reana_dev/utils.py
```diff
@@ -335,3 +337,93 @@ def get_prefixed_component_name(component):
...
+        tag = run_command("git tag --points-at", module, return_output=True)
+        if not tag:
+            click.secho(
+                f"[ERROR] Last commit of {module} does not point to a tag."
+                " Use `--use-latest-known-tag` if you want to proceed using the lastest"
+                " known tag.",
+                err=True,
+                fg="red",
+            ),
+            sys.exit(1)
```
diegodelemos, line 401: Could you extract this logic to a function? For example `is_last_commit_tagged(component)` and return a boolean, so we customise the error message in each case? We will need this in many other places.

reana/reana_dev/utils.py
```diff
@@ -335,3 +337,93 @@ def get_prefixed_component_name(component):
...
+
+    def _check_is_feature_branch(component):
+        branch = get_current_branch(get_srcdir(component))
+        if branch == "master":
+            click.secho(
+                f"[ERROR] Component {component} current branch is master.",
+                err=True,
+                fg="red",
+            ),
+            sys.exit(1)
```
diegodelemos, line 367: This is something we will need in other places too.

reana/reana_dev/utils.py
```diff
@@ -335,3 +337,93 @@ def get_prefixed_component_name(component):
...
+
+    def _replace_pypi_version(
```
diegodelemos, line 369: As you mentioned in the chat, maybe we can come up with something generic for this find and replace in different files? There is a [different implementation in `git-create-release-commit`](https://github.com/reanahub/reana/pull/372/files#diff-5ac538e6b7a8256fc232e7322372c8caR213) for a slightly different use case. Either we do it now, or we merge as is and refactor _a posteriori_, what do you think?

mvidalgarcia: I'll give it a try now and I'll come back to you.

mvidalgarcia: You can check it now. I've tried to be as explicit as possible.

reana/reana_dev/utils.py
```diff
@@ -335,3 +337,123 @@ def get_prefixed_component_name(component):
...
+    tag = run_command("git tag --points-at", package, return_output=True)
+    return bool(tag)
+
+
+def is_feature_branch(component):
+    """Check whether component current branch is different from master."""
+    return get_current_branch(get_srcdir(component)) != "master"
+
+
+def replace_version_string(
```
diegodelemos, line 361: Thanks, looks great! 😊 This covers all the use cases in the create release commit. Just two observations:
- We could rename the function since this replaces strings in component files, no matter what they are
- What about renaming `module_grep` to something along the lines of: `line_selector_expression`/`line_regex`/`line_pattern` (not very gifted today with the naming but you know what I mean)

mvidalgarcia: - Mmh what about `replace_string`, `replace_regex`, `sed_file`...?
- Naming on these tricky functions is always hard, maybe `line_selector_regex`?

mvidalgarcia: Updated, feel free to change it if needed.

reana/reana_dev/git.py
```diff
@@ -767,4 +740,92 @@ def _commit_and_publish_version_bumps(components):
...
+                f"git add setup.py requirements.txt && {commit_cmd}", component,
+            )
+
+    def _push_to_origin(components):
+        for component in components:
+            branch = run_command(
+                "git branch --show-current", component, return_output=True
+            )
+            run_command(
+                f"git push --force origin {branch}", component,
```
mvidalgarcia, line 810: I'm not fully sure about this part. If one is checking out a feature branch of a colleague, will this upgrade their PR?

tiborsimko: Nope, if you are checking only, you are in "latest" or "debug" mode, so you wouldn't need to upgrade minimal versions. The command to upgrade common modules is only used by the integrator, where it is OK to update the PR, because it saves time since the integrator won't need to ask people "please bump r-commons in the other PR in this set".  So I consider it as a "service provided by integrators so that developers don't have to do unnecessary bumping".

tiborsimko:
> Find a better/different name for old git-upgrade-shared-modules (the one that creates PRs)

I would simply remove the old function, because the new functions can take over; a workflow like:

```console
$ reana-dev git-create-release-commit -c CLUSTER -c CLIENT
$ reana-dev git-upgrade-shared-modules -c CLUSTER -c CLIENT
$ reana-dev git-create-pr -c CLUSTER -c CLIENT
```

If some functionality is missing, e.g. force creation of a release commit in order to pass from `0.7.0aN` to `0.7.0`, then we can always add this as new options to the commands, say `--force-new-tag 0.7.0` and such.

reana/reana_dev/utils.py
```diff
@@ -335,3 +337,123 @@ def get_prefixed_component_name(component):
...
+            sys.exit(1)
+
+        replace_version_string(
+            file_="setup.py",
+            find=">=.*,<",
+            replace=f">={new_version},<",
+            line_selector_regex=f"{module}.*>=",
+            component=component,
+        )
+        replace_version_string(
```
tiborsimko, line 444: I get troubles with this branch:

```console
$ reana-dev git-upgrade-shared-modules -c r-w-controller
...
  File "/home/simko/.virtualenvs/reana/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/simko/.virtualenvs/reana/lib/python3.8/site-packages/reana/reana_dev/git.py", line 817, in git_upgrade_shared_modules
    update_module_in_cluster_components(
  File "/home/simko/.virtualenvs/reana/lib/python3.8/site-packages/reana/reana_dev/utils.py", line 437, in update_module_in_cluster_components
    replace_version_string(
NameError: name 'replace_version_string' is not defined
```

reana/reana_dev/git.py
```diff
@@ -724,95 +726,92 @@ def git_push(component):  # noqa: D301
...
-        )
-        click.secho(
-            "PR list 👉  https://github.com/search?q="
-            "org%3Areanahub+is%3Apr+is%3Aopen&"
-            "unscoped_q=is%3Apr+is%3Aopen",
-            fg="green",
-        )
+    _create_commit_or_amend(components)
+    ctx.invoke(git_diff, component=component)
+    if click.confirm("Do you want to push your feature branches?"):
```
tiborsimko, line 813: Instead of having an interactive question, what about adding a new `--push` CLI option, similarly to the `git-merge` command, and only push when `--push` option is passed?  (By default, it would not be.) Advantage: consistent behaviour of several `git-...` commands improves UX.

reana/reana_dev/utils.py
```diff
@@ -335,3 +337,123 @@ def get_prefixed_component_name(component):
...
+
+        replace_string(
+            file_="setup.py",
+            find=">=.*,<",
+            replace=f">={new_version},<",
+            line_selector_regex=f"{module}.*>=",
+            component=component,
+        )
+        replace_string(
+            file_="requirements.txt",
```
tiborsimko, line 445: Updating `requirements.txt` should be optional, because e.g. `r-client` does not contain it.

Currently it gives an error:

```console
$ reana-dev git-upgrade-shared-modules -c r-client --create-new-version-bump-commit
...
[2020-08-11T11:10:33] : cat requirements.txt | grep -n -e "reana-commons.*==" | cut -f1 -d:
cat: requirements.txt: No such file or directory
```

Expected behaviour is: if requirements.txt file exists, update it; otherwise continue.

reana/reana_dev/git.py
```diff
@@ -724,95 +726,92 @@ def git_push(component):  # noqa: D301
...
+    "--use-latest-known-tag",
+    is_flag=True,
+    default=False,
+    help="Should the upgrade use the latest known tag of the shared modules?"
+    " If so, the latest known tag of the shared modules will be used to upgrade the"
+    " provided components. Else, the upgrade will only occur if the latest commit of the"
+    " shared modules points at a tag, in other case, the command will be aborted.",
+)
+@click.option(
+    "--create-new-version-bump-commit",
```
tiborsimko, line 747: Upon second thought, what about inverting the logic and creating new version bump commit by default unless `--amend` is passed? Advantage: it would behave just like `git`, so less surprises for people.

If we like it, then:

- let's rename `--create-new-version-bump-commit` to `--amend`
- let's make it false by default
- let's invert the implementation to always make release commits unless `--amend` is used.

WDYT?

reana/reana_dev/git.py
```diff
@@ -724,95 +726,98 @@ def git_push(component):  # noqa: D301
...
+)
+@click.option(
+    "--amend",
     is_flag=True,
     default=False,
-    help="Should the changes be committed and pull requests created?"
-    " If so, a commit and a PR with"
-    ' "installation: shared packages version bump" message will'
-    " be created for all affected CLUSTER components.",
+    help="Should the changes be integrated as part of the latest commit of each component?.",
```
tiborsimko, line 750: End phrase with `?` instead of `?.`

reana/reana_dev/git.py
```diff
@@ -724,95 +726,98 @@ def git_push(component):  # noqa: D301
...
 )
-def git_shared_modules_upgrade(commit_and_publish):
-    """Upgrade all cluster components to latest REANA-Commons version."""
+@click.option(
+    "--push",
+    is_flag=True,
+    default=False,
+    help="Should the feature branches with the upgrades be pushed?.",
```
tiborsimko, line 756: End phrase with `?` instead of `?.`

tiborsimko: ... also, "be pushed to origin?" because elsewhere in other r-dev commands we are pushing to both origin/upstream... so it may help to be more explicit here.

PR #372, opened 2020-08-06
cli: add git-create-release-commit

### LINKED ISSUE #367: cli: git-create-release-commit
Introduce `git-create-release-commit` command that for each component would see the list of past existing tags and that would create a new release commit with incremented version number.

Example: We would like to release new versions of r-commons, r-client, r-w-controller, r-j-controller. We would use the following command:

```console
$ reana-dev git-create-release-commit \
	-c r-w-controller \
	-c r-commons \
	-c r-client \
	-c r-j-controller
```

The command would do:

- check branch name which is locally checked out -> local-branch
  - if local-branch is master, create new branch release-vX.Y.ZaN where X.Y.ZaN will be determined as stated below
  - if local-branch is some feature-branch, use that
- amend `component/version.py` to bump version number to `X.Y.ZaN`
- amend `docs/openapi.json` if applicable
- amend all the other things as necessary
- create new release commit of the usual format `release: vX.Y.ZaN`

How to determine version number:

- read old tag from `component/version.py`
- do `git tag -l` in the repository to make double sure it is the right one
- increment it as follows:
  - if latest tag is `X.Y.Z`, create `X.Y.(Z+1)`
  - if latest tag is `X.Y.Z(a|b|rc)N`, create `X.Y.Z(a|b|rc)(N+1)`
  - if latest tag is `X.Y.Z.devYYYYMMDD`, create `X.Y.Z.dev(YYYMMDD-of-today)` (eventually also `hhmmss` if both are identical)

P.S. For safety, before doing any work, the command should check whether the latest commit is already of the form `release: vXXX` and if it is, and if the tag exists, then abort, because there is apparently nothing to be released.

diegodelemos:
* Closes #367.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -10,8 +10,8 @@
 
 import datetime
 import os
-import sys
 import subprocess
+import sys
 
 import click
 
@@ -21,9 +21,12 @@ from reana.config import (
     REPO_LIST_ALL,
     REPO_LIST_SHARED,
 )
+
 from reana.reana_dev.utils import (
+    bump_component_version,
     display_message,
     fetch_latest_pypi_version,
+    get_current_component_version_from_source_files,
     get_srcdir,
     run_command,
     select_components,
@@ -82,6 +85,57 @@ def get_current_commit(srcdir):
         .decode()
         .rstrip("\r\n")
     )
+
+
+def git_is_current_version_tagged(component):
+    """Determine whether the current version in source code is present as a git tag."""
+    current_version = get_current_component_version_from_source_files(component)
+    is_version_tagged = int(
+        run_command(
+            f"git tag --list {current_version} | wc -l",
+            component,
+            display=False,
+            return_output=True,
+        )
+    )
+    return bool(is_version_tagged)
+
+
+def git_create_release_commit(component, next_version=None):
+    """Create a release commit for the given component."""
+    if "release:" in get_current_commit(get_srcdir(component)):
+        display_message("Nothing to do, last commit is a release commit.", component)
+        return False
+
+    current_version = get_current_component_version_from_source_files(component)
+    if not current_version:
+        display_message(
+            f"Version cannot be autodiscovered from source files.", component
+        )
+        sys.exit(1)
+    elif not git_is_current_version_tagged(component) and not next_version:
+        display_message(
+            f"Current version ({current_version}) "
+            "not present as a git tag, please release it and add a tag.",
+            component,
+        )
+        sys.exit(1)
+
+    next_version, modified_files = bump_component_version(
+        component, current_version, next_version=next_version
+    )
+
+    if (
+        run_command(
+            "git branch --show-current", component, display=False, return_output=True,
+        )
+        == "master"
+    ):
+        run_command(f"git checkout -b release-{next_version}", component)
+
+    run_command(f"git add {' '.join(modified_files)}", component)
+    run_command(f"git commit -m 'release: {next_version}'", component)
+    return True
 
 
 @click.group()
@@ -818,6 +872,47 @@ def git_upgrade_shared_modules(
     ctx.invoke(git_diff, component=component)
     if push:
         _push_to_origin(components)
+
+
+@click.option(
+    "--component",
+    "-c",
+    required=True,
+    multiple=True,
+    help="Which components? [shortname|name|.|CLUSTER|ALL]",
+)
+@click.option(
+    "--version", "-v", help="Shall we manually specify component's next version?",
+)
+@git_commands.command(name="git-create-release-commit")
+def git_create_release_commit_command(component, version):  # noqa: D301
+    """Create a release commit for the specified components.
+
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
+                         * (5) special value 'CLIENT' that will expand to
+                               cover all REANA client components;
+                         * (6) special value 'DEMO' that will expand
+                               to include several runable REANA demo examples;
+                         * (7) special value 'ALL' that will expand to include
+                               all REANA repositories.
+    :param version: Manually specifies the version for the component. If not provided,
+        the last version will be auto-incremented..
+    :type component: str
+    :type version: str
+    """
+    components = select_components(component)
+    for component in components:
+        if git_create_release_commit(component, next_version=version):
+            display_message("Release commit created.", component)
 
 
 git_commands_list = list(git_commands.commands.values())
--- a/reana/reana_dev/utils.py
+++ b/reana/reana_dev/utils.py
@@ -9,15 +9,24 @@
 """`reana-dev` related utils."""
 
 import datetime
+import importlib.util
+import json
 import os
 import subprocess
 import sys
 
 import click
+import semver
+import yaml
+from packaging.version import InvalidVersion, Version
 
 from reana.config import (
     COMPONENTS_USING_SHARED_MODULE_COMMONS,
     COMPONENTS_USING_SHARED_MODULE_DB,
+    HELM_VERSION_FILE,
+    JAVASCRIPT_VERSION_FILE,
+    OPENAPI_VERSION_FILE,
+    PYTHON_VERSION_FILE,
     REPO_LIST_ALL,
     REPO_LIST_CLIENT,
     REPO_LIST_CLUSTER,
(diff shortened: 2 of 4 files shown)


reana/reana_dev/git.py
```diff
@@ -82,6 +91,195 @@ def get_current_commit(srcdir):
...
+
+    version_files = get_component_version_files(component)
+    next_version = ""
+    if version_files.get(HELM_VERSION_FILE):
+        next_version = bump_semver2_version(current_version)
+        _replace_version_string(
+            file_=version_files.get(HELM_VERSION_FILE),
+            current_version=current_version,
+            next_version=next_version,
+        )
```
mvidalgarcia, line 228: What do you think about saving the args of this function in vars and call it only once at the end?

reana/reana_dev/utils.py
```diff
@@ -335,3 +344,147 @@ def get_prefixed_component_name(component):
...
+    except InvalidVersion as e:
+        click.echo(
+            f"Current {current_version} is not a valid PEP440 version. Please amend it"
+        )
+
+
+def bump_component_version(component, current_version):
+    """Bump to next component version."""
+
+    def _replace_version_string(file_=None, current_version=None, next_version=None):
```
diegodelemos, line 462: Remove and use the implementation from https://github.com/reanahub/reana/pull/371/files#diff-f1d005e86d9c3c61bd4155479765040aR361

reana/reana_dev/git.py
```diff
@@ -82,6 +85,45 @@ def get_current_commit(srcdir):
...
+        display_message("Nothing to do, last commit is a release commit.", component)
+        return False
+
+    if not git_is_current_version_tagged(component):
+        display_message(
+            f"Current version ({get_component_current_version_from_file(component)}) "
+            "not present as a git tag, plese release it and add a tag.",
+            component,
+        )
+        return False
```
tiborsimko, line 109: Expected to see this error message for a repo that does not contain any tags and the latest commit is not a release commit, but got instead a new `release-` branch created.

Example:

```console
$ reana-dev git-create-release-commit -c r-a-vomsproxy
[2020-08-07T16:36:21] reana-auth-vomsproxy: git checkout -b release-
Switched to a new branch 'release-'
[2020-08-07T16:36:21] reana-auth-vomsproxy: git commit -m 'release: '
On branch release-
nothing to commit, working tree clean
[2020-08-07T16:36:21] reana-auth-vomsproxy: Command 'git commit -m 'release: '' returned non-zero exit status 1.
```

reana/reana_dev/utils.py
```diff
@@ -335,3 +344,147 @@ def get_prefixed_component_name(component):
...
+        )
+        if file_relative_path:
+            version_files[file_] = os.path.join(
+                get_srcdir(component=component), file_relative_path
+            )
+
+    return version_files
+
+
+def get_component_current_version_from_file(component):
```
tiborsimko, line 372: Naming comment: `..._from_file()` evokes a feeling that it is getting version from one file, so I was expecting to see it as argument.  Perhaps rename to `get_current_component_version_from_source_files()` or somesuch.

reana/reana_dev/git.py
```diff
@@ -82,6 +85,45 @@ def get_current_commit(srcdir):
...
+
+    if (
+        run_command(
+            "git branch --show-current", component, display=False, return_output=True,
+        )
+        == "master"
+    ):
+        run_command(f"git checkout -b release-{next_version}", component)
+
+    run_command(f"git commit -m 'release: {next_version}'", component)
```
tiborsimko, line 123: You need to stage modified files first? Two options:

- use wild staging: `git commit -a -m ...` could do the job, but it might be dangerous
- use targeted staging: `git add <version_file>` to add only the version file you incremented (python or npm)

reana/reana_dev/git.py
```diff
@@ -82,6 +85,45 @@ def get_current_commit(srcdir):
...
+    tags_history = run_command(
+        "git tag --list", component, display=False, return_output=True
+    )
+    current_version = get_component_current_version_from_file(component)
+    return current_version in tags_history
+
+
+def git_create_release_commit(component):
+    """Create a release commit for the given component."""
+    if "release: v" in get_current_commit(get_srcdir(component)):
```
tiborsimko, line 99: It is good to check for `release: vXXX`, but when you run the `git-create-release-commit` function, the release branch being created (elsewhere in the code in line 121, not in this place) is called `release-0.7.0a4` without leading `v`. Shall we use leading `v` in branch names for consistency?

reana/reana_dev/utils.py
```diff
@@ -335,3 +344,147 @@ def get_prefixed_component_name(component):
...
+def bump_pep440_version(current_version, part=None):
+    """Bump a PEP440 version string.
+
+    :param current_version: current version to be bumped
+    :param part: part of the PEP440 version to bump (one of: [major, minor, micro, pre]).
+    :type current_version: str
+    :type part: str
+
+    :return: String representation of the next version
+    :rtype: string
```
tiborsimko, line 436: Note: the function does not bump `dev.YYYYMMDD` into `dev.<TODAY's YYYMMDD>`, it only increments the value by one.  Perhaps make it a special case to be incremented otherwise here, as described in the issue? (Or we can forget it, since we plan to switch to a/b/rc everywhere...)

reana/reana_dev/git.py
tiborsimko: This works now, but the exit code is not set:

```console
~r/reana pr-372 reana $ reana-dev git-create-release-commit -c r-a-vomsproxy
[2020-08-11T09:52:32] reana-auth-vomsproxy: Version cannot be autodiscovered.
[2020-08-11T09:52:32] reana-auth-vomsproxy: Current version () not present as a git tag, plese release it and add a tag.

~r/reana pr-372 reana $ echo $?
0
```

Please add `sys.exit(1)` so that we fail properly if this command is plugged into a longer script.

reana/reana_dev/git.py
tiborsimko: I still get `release-0.7.0a4` branch names, so this does not seem to be resolved. Shall we use or remove leading `v` everywhere except git tags? Let's sync IRL.

reana/reana_dev/git.py
```diff
@@ -84,6 +87,50 @@ def get_current_commit(srcdir):
...
+    current_version = get_current_component_version_from_source_files(component)
+    if not current_version:
+        display_message(
+            f"Version cannot be autodiscovered from source files.", component
+        )
+        sys.exit(1)
+    elif not git_is_current_version_tagged(component):
+        display_message(
+            f"Current version ({current_version}) "
+            "not present as a git tag, plese release it and add a tag.",
```
tiborsimko, line 114: Hmm. I think this check would prevent the command from running on many of our repositories, because we don't have git tags for every YYYYMMDD version in the past, for example:

```console
$ reana-dev git-create-release-commit -c r-client
[2020-08-11T19:31:40] reana-client: Current version (0.7.0.dev20200706) not present as a git tag, plese release it and add a tag.
```

Moreover, for new components, there will be no known git tags, so we should perhaps introduce a CLI option to make the check optional? E.g. `--no-tag-check`? Or even better `--tag X.Y.Z` which, when the option is passed, it would not auto-discover anything but used the passed value? Would be used for "first tags" releases...

P.S. Note the "plese" typo.

reana/reana_dev/git.py
```diff
@@ -84,6 +87,50 @@ def get_current_commit(srcdir):
     )
 
 
+def git_is_current_version_tagged(component):
+    """Determine whether the current version in source code is present as a git tag."""
+    tags_history = run_command(
+        "git tag --list", component, display=False, return_output=True
+    )
+    current_version = get_current_component_version_from_source_files(component)
+    return current_version in tags_history if tags_history else False
```
tiborsimko, line 96: This check may be dangerous, before the command returns one big string:

```
In [5]: run_command("git tag --list", "reana-commons", display=False, return_output=True)
Out[5]: 'v0.3.0\nv0.3.1\nv0.4.0\nv0.5.0\nv0.6.0\nv0.7.0a1\nv0.7.0a2\nv0.7.0a3'
```

so in theory one could get mismatches via matching substrings.

Perhaps change the implementation to check for a tag existence via `git tag -l <value>` technique.  E.g. compare:

```console
$ git tag -l 0.3.0 | wc -l
0
$ git tag -l v0.3.0 | wc -l
1
```

reana/reana_dev/git.py
```diff
@@ -820,4 +874,45 @@ def _push_to_origin(components):
...
 
+@click.option(
+    "--component",
+    "-c",
+    required=True,
+    multiple=True,
+    help="Which components? [shortname|name|.|CLUSTER|ALL]",
+)
+@click.option(
+    "--next-version", "-v", help="Shall we manually specify component's next version?",
```
tiborsimko, line 885: All works well :+1: One last naming comment: `-v` and `--next-version` do not play well together, it is usually good to keep the first letter identical.  Two options:

- keep `--next-version` name and remove short `-v` variant; we can be simply using long name

- change to `--version` name and keep short `-v`; it makes sense since you are basically saying "git create new release commit with version X.Y.Z"

reana/reana_dev/git.py
```diff
@@ -820,4 +874,45 @@ def _push_to_origin(components):
...
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
+                         * (5) special value 'CLIENT' that will expand to
+                               cover all REANA client components;
+                         * (6) special value 'DEMO' that will expand
+                               to include several runable REANA demo examples;
+                         * (7) special value 'ALL' that will expand to include
+                               all REANA repositories.
+    :param next_version: Manually specifies the version for the component. If not
+        provided the next version is autodiscovered.
```
tiborsimko, line 908: Instead of "autodiscovered", it would be more precise to say "If not provided, the last version will be auto-incremented."

PR #374, opened 2020-08-11
cli: git-create-pr

### LINKED ISSUE #366: cli: git-create-pr
Introduce `git-create-pr` command that would accept a set of components and issue PRs for them.

Example workflow:

The developer works on a feature that requires a set of branches to various components, for example:

- r-w-controller, branch fix-pod-timeout
- r-commons, branch fix-k8s-config
- r-j-controller, branch fix-something-else

i.e. the branches do not necessarily have to be named the same.

The developer finishes the implementation and runs the usual local test to make sure everything is OK:

```console
$ reana-dev run-ci -m /var/reana:/var/reana --exclude-components=r-ui,r-a-vomsproxy,r-a-krb5 -c r-d-helloworld
```

Once this is successful, the developer would like to issue PRs for the above branches easily using one command.

First, the developer checks out the branch status:

```console
$ reana-dev git-fetch -c CLUSTER
$ reana-dev git-status -c CLUSTER | grep -v '@ master'
```

Second, if everything is ready, the developer would use the new command:

```
$ reana-dev git-create-pr \
	-c r-w-controller \
	-c r-commons \
	-c r-j-controller
```

The command would do:

- check branch name which is locally checked out; abort if it is `master`
- do git push origin branch-name
- issue PR against upstream/component

Note: The command could actually check first whether the local branch is up-to-date against upstream master (which is being done in `git-status` currently) to ease the life of the developer.

P.S. The command could make use of the new `gh` or the old `hub` GitHub CLI client.

diegodelemos:
* Closes #366.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -28,6 +28,7 @@ from reana.reana_dev.utils import (
     fetch_latest_pypi_version,
     get_current_component_version_from_source_files,
     get_srcdir,
+    is_feature_branch,
     run_command,
     select_components,
     update_module_in_cluster_components,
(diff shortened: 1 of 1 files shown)


reana/reana_dev/git.py
```diff
@@ -815,4 +828,58 @@ def _commit_and_publish_version_bumps(components):
...
+            get_current_branch(get_srcdir(component)) == "master"
+        ):  # replace with is_feature_branch from #371
+            display_message(
+                "You are trying to create PR but the current branch is master, please amend.",
+                component,
+            )
+        else:
+            report_component_branch_difference(
+                component, "upstream/master", exit_code=True
+            )
```
tiborsimko, line 881: Please add some informative message to people, currently the would see:

```console
$ reana-dev git-create-pr -c r-w-controller
reana-workflow-controller @ test2 (1 AHEAD 5 BEHIND upstream/master) @ 4faf564 test
```

It may be good to say something like:

```
Error: please rebase your feature branch against latest master.
```

(Although beware that there will be `maint-0.6` and `maint-0.7`, but we shall upgrade the command later to account for this possibility (=to check against a different branch than master). For now, the simple error message is enough.)

reana/reana_dev/git.py
```diff
@@ -815,4 +828,58 @@ def _commit_and_publish_version_bumps(components):
...
+            "hub pr list -L 1",
+        ]:
+            run_command(cmd, component)
+
+    for component in select_components(component):
+        if (
+            get_current_branch(get_srcdir(component)) == "master"
+        ):  # replace with is_feature_branch from #371
+            display_message(
+                "You are trying to create PR but the current branch is master, please amend.",
```
tiborsimko, line 875: ... please switch to the wanted feature branch.

Also, please add `sys.exit(1)` so that this command would indicate failure.

reana/reana_dev/git.py
```diff
@@ -820,4 +884,90 @@ def _push_to_origin(components):
...
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
```
tiborsimko, line 934: Actually `default` is not `CLUSTER`, if you don't pass any `-c`, you get a traceback:

```console
$ reana-dev git-create-pr
...
TypeError: git_create_pr_command() missing 1 required positional argument: 'component'
```

Perhaps it is good not to introduce any default, but require users to always provide `-c` option, as this could be potentially disruptive command creating a plethora of PRs by mishap.

Ditto for other potentially disruptive commands such as `git-create-release-commit`, let's require `-c` for all those?

PR #376, opened 2020-08-12
cli: release-pypi

diegodelemos:
* Closes #368.

PR #377, opened 2020-08-12
cli: git-tag

### LINKED ISSUE #369: cli: git-tag
Introduce `git-tag` command that for each component would create git tag corresponding to the release.

Example: We would like to release r-commons and r-client, so we prepare release commit, create tag, and upload to pypi:

```console
$ reana-dev git-create-release-commit -c r-commons -c r-client
$ reana-dev git-tag -c r-commons -c r-client
$ reana-dev release-pypi -c r-commons -c r-client
```

The new command `git-tag` would do:

- check whether the latest commit is `release: vX.Y.Z` one; abort if not (and advice to use `r-dev git-create-release-commit` command)
- detect version number of the component by reading `compontent/version.py` -> X.Y.ZaN
- do `git tag vX.Y.ZaN`

The final status is that the component should have the last commit the usual `release: vX.Y.Z` commit and the git tag corresponding to the release commit version.

P.S. We do this separately, and on locally checked out branches by the integrator, because we shall be releasing packages on PyPI, but if something fails during this process (e.g. Python trove classifiers for PyPI are wrong), we would still be able to amend it before pushing the release and the tag to `upstream` on GitHub.

diegodelemos:
* Closes #369.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -997,4 +997,46 @@ def git_create_pr_command(component):  # noqa: D301
             _git_create_pr(component)
 
 
+@click.option(
+    "--component",
+    "-c",
+    required=True,
+    multiple=True,
+    help="Which components? [name|CLUSTER]",
+)
+@git_commands.command(name="git-tag")
+def git_tag(component):  # noqa: D301
+    """Create the corresponding git tag for components with release commits.
+
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components;
+                         * (5) special value 'CLIENT' that will expand to
+                               cover all REANA client components;
+                         * (6) special value 'DEMO' that will expand
+                               to include several runable REANA demo examples;
+                         * (7) special value 'ALL' that will expand to include
+                               all REANA repositories.
+    :type component: str
+    """
+    components = select_components(component)
+    for component in components:
+        if not is_last_commit_release_commit(component):
+            click.secho(
+                "The last commit is not a release commit. Please use `reana-dev git-create-release-commit`.",
+                fg="red",
+            )
+            sys.exit(1)
+
+        current_version = get_current_component_version_from_source_files(component)
+        run_command(f"git tag {current_version}")
+
+
 git_commands_list = list(git_commands.commands.values())


reana/reana_dev/git.py
```diff
@@ -991,4 +992,46 @@ def _git_create_pr(comp):
...
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
```
tiborsimko, line 1015: `[default]` should not be here, we should require a component to be passed, as in the other PR.

reana/reana_dev/git.py
```diff
@@ -29,6 +29,7 @@
     get_current_component_version_from_source_files,
     get_srcdir,
     is_feature_branch,
+    is_last_commit_release_commit,
```
tiborsimko, line 32: This one is defined I believe in #376, so this PR can wait until that one is merged.

reana/reana_dev/git.py
```diff
@@ -991,4 +992,46 @@ def _git_create_pr(comp):
...
+    for component in components:
+        if not is_last_commit_release_commit(component):
+            click.secho(
+                "The last commit is not a release commit. Please use `reana-dev git-create-release-commit`.",
+                fg="red",
+            )
+            sys.exit(1)
+
+        current_version = get_current_component_version_from_source_files(component)
+        run_command(f"git tag v{current_version}")
```
tiborsimko, line 1034: Let's remove leading `v` from tags, as discussed in the other PR?

reana/reana_dev/git.py
tiborsimko: (... or move the function here if we are merging this first?!)

reana/reana_dev/git.py
```diff
@@ -997,4 +997,46 @@ def _git_create_pr(comp):
...
+                               to include several runable REANA demo examples;
+                         * (7) special value 'ALL' that will expand to include
+                               all REANA repositories.
+    :type component: str
+    """
+    components = select_components(component)
+    for component in components:
+        if not is_last_commit_release_commit(component):
+            click.secho(
+                "The last commit is not a release commit. Please use `reana-dev git-create-release-commit`.",
```
tiborsimko, line 1033: Works well :+1: 

Cosmetic comment: in the output message

```
$ reana-dev git-tag -c r-commons
The last commit is not a release commit. Please use `reana-dev git-create-release-commit`.
```

we could print out also argument, so that users can copy/paste if they wish:

```
The last commit is not a release commit. Please use `reana-dev git-create-release-commit -c r-commons` first.
```

PR #380, opened 2020-08-14
cli: release-docker

### LINKED ISSUE #370: cli: release-docker
Introduce `release-docker` command that for each component would push local image to Docker Hub.

Example: We are ready to release a new version of REANA. The release manager tested everything locally using:

```console
$ reana-dev run-ci -m /var/reana:/var/reana --exclude-components=r-a-vomsproxy,r-a-krb5 -c r-d-helloworld --no-cache
```

(Beware: r-a-vomsproxy and r-a-krb5 are hard-coded to 1.0.0 in r-j-controller, so currently there is no need to build and test them locally.)

If everything is OK, we are ready to tag and push new docker images using:

```console
$ reana-dev release-docker -c CLUSTER
```

The new command would do:

- check whether it is a dockerised component; skip if not (e.g. for r-commons)
- check whether the latest commit is `release: vX.Y.Z` one; abort if not (and advice to use `r-dev git-create-release-commit` coming in the chain before this command)
- check whether the latest commit is also git-tagged; abort if not (and advice to use `r-dev git-tag` coming in the release chain before this command)

Only when all these tests are OK we are ready to push images.  This is because we don't want to push some images and then later see that for some component there are problems. So the second phase would do:

- detect git tag vX.Y.ZaN assigned to the release commit
- create docker image version number corresponding to semantic versioning, such as:
  - if git tag is `vX.Y.Z`, the docker tag will be `X.Y.Z`
  - if git tag is `vX.Y.ZaN`, the docker tag will be `X.Y.Z-alpha.N`
  - if git tag is `vX.Y.ZbN`, the docker tag will be `X.Y.Z-beta.N`
  - if git tag is `vX.Y.ZrcN`, the docker tag will be `X.Y.Z-rc.N`
  - if git tag is `vX.Y.ZdevYYYMMDD`, the docker tag will be `X.Y.Z-dev.YYYYMMDD` (but we don't really want to use these anymore, so we could also warn and abort...)
  - if git tag does not match any of the above, abort with an error
 - tag images locally e.g. `docker tag reanahub/reana-auth-vomsproxy:latest reanahub/reana-auth-vomsproxy:1.0.0a7`
 - push images to docker hub e.g. `docker push reanahub/reana-auth-vomsproxy:1.0.0a7`

Note that we already have command `reana-dev docker-push` which can be used and which allows to push to user repositories for testing, e.g. using `johndoe` instead of `reanahub` by means of the `-u` option. We should preserve this option here too.

Note: this command is a step in the release chain that will be used by the `release-helm` command described in #363 that will do the rest of the release procedures. It could exist separately or could be part of the `release-helm` command, since we rarely want to push docker images without pushing Helm charts.

(Well, we do for `reana-auth-krb5` or for `reana-env-jupyter`, so if the above procedure works for these components, then so much the better! But we can still do these "by hand", we don't necessarily have to automatise these already... Plus the git tags would not probably be PEP 440 compatible (`vX.Y.ZaN`) but rather semver compatible already (`vX.Y.Z-alpha.N`) if the given component is not written in Python.)

diegodelemos:
* Closes #370.

### DIFF
--- a/reana/reana_dev/release.py
+++ b/reana/reana_dev/release.py
@@ -12,23 +12,135 @@ import sys
 from time import sleep
 
 import click
+import semver
 
+from reana.reana_dev.docker import docker_push
 from reana.reana_dev.git import (
     git_clean,
     git_is_current_version_tagged,
     is_last_commit_release_commit,
 )
 from reana.reana_dev.utils import (
+    display_message,
     fetch_latest_pypi_version,
     get_current_component_version_from_source_files,
+    get_current_tag,
+    is_component_dockerised,
+    parse_pep440_version,
     run_command,
     select_components,
+    translate_pep440_to_semver2,
 )
+
+
+def is_component_releasable(component, exit_code=False, display=False):
+    """Determine whether a component is releasable.
+
+    Last commit should be a release commit and the new version should be git tagged.
+
+    :param component: Component to determine whether if it is releasable or not.
+    :param exit_code: Whether the program should exit with error exit code if
+        the condition is not met.
+    :param display: Whether error messages providing instructions on how to fix
+        the problem should be displayed to stdout.
+
+    :type component: str
+    :type exit_code: bool
+    :type display: bool
+    :rtype: bool
+    """
+    is_releasable = True
+    error_message = ""
+    if not is_last_commit_release_commit(component):
+        error_message = "The last commit is not a release commit. Please use `reana-dev git-create-release-commit`."
+        is_releasable = False
+    if not git_is_current_version_tagged(component):
+        error_message = (
+            "The current version is not tagged. Please use `reana-dev git-tag`."
+        )
+        is_releasable = False
+
+    if error_message and display:
+        display_message(error_message, component)
+    if not is_releasable and exit_code:
+        sys.exit(1)
+
+    return is_releasable
 
 
 @click.group()
 def release_commands():
     """Release commands group."""
+
+
+@click.option(
+    "--component",
+    "-c",
+    required=True,
+    multiple=True,
+    help="Which components? [name|CLUSTER]",
+)
+@click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option("--image-name", help="Should the component have a custom image name?")
+@release_commands.command(name="release-docker")
+@click.pass_context
+def release_docker(ctx, component, user, image_name):  # noqa: D301
+    """Release a component on Docker Hub.
+
+    \b
+    :param components: The option ``component`` can be repeated. The value may
+                       consist of:
+                         * (1) standard component name such as
+                               'reana-workflow-controller';
+                         * (2) short component name such as 'r-w-controller';
+                         * (3) special value '.' indicating component of the
+                               current working directory;
+                         * (4) special value 'CLUSTER' that will expand to
+                               cover all REANA cluster components [default];
+                         * (5) special value 'CLIENT' that will expand to
+                               cover all REANA client components;
+                         * (6) special value 'DEMO' that will expand
+                               to include several runable REANA demo examples;
+                         * (7) special value 'ALL' that will expand to include
+                               all REANA repositories.
+    :type component: str
+    """
+    components = select_components(component)
+    cannot_release_on_dockerhub = []
+    for component_ in components:
+        if not is_component_dockerised(component_):
+            cannot_release_on_dockerhub.append(component_)
+        is_component_releasable(component_, exit_code=True, display=True)
+
+        current_tag = get_current_tag(component_)
+        full_image_name = f"{user}/{image_name or component_}"
+        docker_tag = ""
+
+        if parse_pep440_version(current_tag):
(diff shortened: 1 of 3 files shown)


reana/reana_dev/release.py
```diff
@@ -12,25 +12,127 @@
...
+    "-c",
+    required=True,
+    multiple=True,
+    help="Which components? [name|CLUSTER]",
+)
+@click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option("--image-name", help="Should the component have a custom image name?")
+@release_commands.command(name="release-docker")
+@click.pass_context
+def release_docker(ctx, component, user, image_name):  # noqa: D301
```
tiborsimko, line 87: Works well :+1: I have three observations, we can decide whether to address some of them now or later.

tiborsimko: Observation no.1: special components:

```console
$ reana-dev release-docker -c r-a-krb5
[2020-08-14T10:08:57] reana-auth-krb5: git describe --tags
Something went wrong while translating 1.0.1 to semver2.
```

Expected to see 1.0.1 reused, however this component is a bit special for the release procedures, so we could address it later.

tiborsimko: Observation no.2: not eligible components:

```console
$ reana-dev release-docker -c r-client
$ echo $?
0
```

Expected to see exit code 1 and a message like "Sorry, reana-client is not releasable on Docker Hub."

tiborsimko: Observation no.3: pushing eligible components to custom Docker Hub organisation:

```console
$ reana-dev release-docker -c . -u tiborsimko
[2020-08-14T10:18:21] reana-server: git describe --tags
[2020-08-14T10:18:21] reana-server: docker tag reanahub/reana-server:latest reanahub/reana-server:0.7.0-alpha.1
[2020-08-14T10:18:21] reana-server: docker push tiborsimko/reana-server:0.7.0-alpha.1
The push refers to repository [docker.io/tiborsimko/reana-server]
An image does not exist locally with the tag: tiborsimko/reana-server
[2020-08-14T10:18:21] reana-server: Command 'docker push tiborsimko/reana-server:0.7.0-alpha.1' returned non-zero exit status 1.
```

Expected to see local image named `reanahub/*` be pushed to `tiborsimko/*`.

(Also, we could show the output of `git describe --tags`; we currently only show the command, not its output, that people might expect to see, as when seeing `diff` statements during other `reana-dev` commands such as `reana-dev git-upgrade-shared-modules`.)

reana/reana_dev/release.py
```diff
@@ -12,25 +12,127 @@
...
+    """Release a component on dockerhub.com.
```
tiborsimko, line 88: Cosmetics: "on Docker Hub", or "on hub.docker.com" (which is the canonically visible URL).

reana/reana_dev/release.py
tiborsimko: Observation no.4: pushing multiple eligible components:

```
$ reana-dev release-docker -c reana-workflow-controller -c reana-workflow-engine-serial -c reana-message-broker -c reana-ui -c reana-workflow-engine-cwl -c reana-job-controller -c  reana-workflow-engine-yadage -c reana-server
...
[2020-08-14T11:33:46] reana-workflow-engine-cwl: docker push reanahub/reana-workflow-engine-cwl:0.7.0-alpha.1
The push refers to repository [docker.io/reanahub/reana-workflow-engine-cwl]
619a849c3d21: Layer already exists
9e29d7f32ea1: Layer already exists
da59ca332f49: Layer already exists
c85f58793455: Layer already exists
fb04193e9129: Layer already exists
87e62c6cdb14: Layer already exists
a7605bc04ab2: Layer already exists
0571d07c98ad: Layer already exists
f0287054e4ef: Layer already exists
b6622a14e56a: Layer already exists
95ef25a32043: Layer already exists
0.7.0-alpha.1: digest: sha256:a77abf8c0ce46cdb6aad8f373df2727250223eb75f3dcb1fd96765ffb0b7105a size: 2633
[2020-08-14T11:33:50] reana-job-controller: git describe --tags
[2020-08-14T11:33:50] reana-job-controller: docker tag reanahub/reana-workflow-engine-serial:latest reanahub/reana-workflow-engine-serial:0.7.0-alpha.1
[2020-08-14T11:33:50] reana-job-controller: docker push reanahub/reana-job-controller:0.7.0-alpha.1
The push refers to repository [docker.io/reanahub/reana-job-controller]
tag does not exist: reanahub/reana-job-controller:0.7.0-alpha.1
[2020-08-14T11:33:50] reana-job-controller: Command 'docker push reanahub/reana-job-controller:0.7.0-alpha.1' returned non-zero exit status 1.
```

Note that when `r-w-e-serial` was being tagged, `r-j-controller` was attempted to be pushed.

This is probably caused be reusing the `component` function argument internally in the loop:

```
+    components = select_components(component)
+    for component in components:
```

Please change to `for acomponent in components` (or `component_` or something else than "component") for safety here and elsewhere

PR #391, opened 2020-09-03
cli: allow passing email & pwd to cluster-deploy

mvidalgarcia:
closes reanahub/reana-server#270

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,7 +200,7 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 			sleep ${TIMECHECK}; \
 		fi;\
 	done && \
-	source ${PWD}/scripts/create-admin-user.sh ${TRUNC_INSTANCE_NAME} && \
+	source ${PWD}/scripts/create-admin-user.sh ${TRUNC_INSTANCE_NAME} admin@reana.io 123456 && \
 	eval $$(reana-dev client-setup-environment $(addprefix --server-hostname , ${SERVER_URL}))
 
 example: # Run one or several demo examples.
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -200,7 +200,7 @@ deploy: # Deploy/redeploy previously built REANA cluster.
 			sleep ${TIMECHECK}; \
 		fi;\
 	done && \
-	source ${PWD}/scripts/create-admin-user.sh ${TRUNC_INSTANCE_NAME} && \
+	source ${PWD}/scripts/create-admin-user.sh ${TRUNC_INSTANCE_NAME} admin@reana.io mysecretpassword && \
```
mvidalgarcia, line 203: I added these values here for compatibility but I guess Makefile will become obsolete soon.

diegodelemos: I would change `mysecretpassworde` for `123456` as it is the one we classicly use in these cases.

diegodelemos:
When running `run-ci` I get:
```console
$ reana-dev run-ci -m /var/reana:/var/reana -c r-d-helloworld --exclude-components=r-a-vomsproxy
...
[2020-09-03T13:35:37] reana: reana-dev cluster-deploy --mode latest
Usage: reana-dev cluster-deploy [OPTIONS]
Try 'reana-dev cluster-deploy --help' for help.

Error: Missing option '--admin-email'.
[2020-09-03T13:35:37] reana: Command 'reana-dev cluster-deploy --mode latest' returned non-zero exit status 2.
```

I would expect it to run, so maybe run-ci can pass some defaults?

diegodelemos:
I am not able to login with:
```console
$ reana-dev cluster-deploy --mode latest --admin-email admin@localhost.local --admin-password 123456
```

I've checked and there is no user in the DB.

mvidalgarcia:
> **Edited**:
> 
> This happens because of invalid email format, however, the execution doesn't exit with an exit code and the error message is buried.
>

This should be fixed by now.

diegodelemos (approved):

Issue #387, opened 2020-09-01
cli: cluster-delete -m /var/reana:/var/reana


tiborsimko:
The cluster creation command (`cluster-create`) or the run CI command (`run-ci`) have an option to share a cluster directory with the host system. The typical usage is `-m /var/reana:/var/reana` so that `/var/reana` would be persistent and easily accessible from host. 

The cluster deletion command (`cluster-delete`) does not have this option, which sometimes lead to a situation where `/var/reana/...` remains populated with old instance content, which may not be always wanted.

Currently we do things like:

```console
$ reana-dev run-ci -m /var/reana:/var/reana --exclude-components=r-a-vomsproxy,r-a-krb5 -c r-d-helloworld
...
$ reana-dev cluster-delete
$ sudo rm -rf /var/reana/*
```

The aim of this issue is to introduce exactly the same `-m` option for consistency, so that one could do:

```console
$ reana-dev cluster-delete -m /var/reana:/var/reana
```

Note that the parameter is of the format `-m local_path:cluster_nodepath`, so one has to delete only `local_path/*`.

Note that this may require to use `sudo`.

PR #397, opened 2020-09-07
cli: cluster delete mount option 

audrium:
Closes: https://github.com/reanahub/reana/issues/387

### DIFF
--- a/reana/reana_dev/cluster.py
+++ b/reana/reana_dev/cluster.py
@@ -376,11 +376,37 @@ def cluster_unpause():
     run_command(cmd, "reana")
 
 
+@click.option(
+    "-m",
+    "--mount",
+    "mounts",
+    multiple=True,
+    help="Which local path directories are to be deleted? [local_path:cluster_node_path]",
+)
 @cluster_commands.command(name="cluster-delete")
-def cluster_delete():
-    """Delete REANA cluster."""
-    cmd = "kind delete cluster"
-    run_command(cmd, "reana")
+def cluster_delete(mounts):  # noqa: D301
+    """Delete REANA cluster.
+
+    \b
+    Example:
+       $ reana-dev cluster-delete -m /var/reana:/var/reana
+    """
+    cmds = []
+    # delete cluster
+    cmds.append("kind delete cluster")
+    # remove only local paths where cluster path starts with /var/reana for safety
+    for mount in mounts:
+        local_path, cluster_node_path = mount.split(":")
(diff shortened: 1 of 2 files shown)


reana/reana_dev/cluster.py
```diff
@@ -376,10 +376,26 @@ def cluster_unpause():
     run_command(cmd, "reana")
 
 
+@click.option(
+    "-m",
+    "--mount",
+    "mounts",
+    multiple=True,
+    help="Which directories are used to mount the cluster nodes? [local_path:cluster_node_path]",
```
tiborsimko, line 384: Please change the help message to `Which local path directories are to be deleted? [local_path:cluster_node_path]"` in order to better indicate that these will get deleted. (danger!)

reana/reana_dev/cluster.py
```diff
@@ -376,11 +376,25 @@ def cluster_unpause():
...
-    cmd = "kind delete cluster"
-    run_command(cmd, "reana")
+def cluster_delete(mounts):  # noqa: D301
+    """Delete REANA cluster.
+
+    \b
+    Example:
+       $ reana-dev cluster-delete -m /var/reana:/var/reana
+    """
+    cmds = ["rm -rf '{}'".format(mount.split(":")[0]) for mount in mounts]
```
tiborsimko, line 394: (1) In some cases the `/var/reana` is owned by another person, so it may be good to do:

```
rm -rf '{}/*'
```

in the above command.

(2) Also, the command may be dangerous if somebody uses `-m ~/mydata:/path` -- let's beware! I'd say let's add a check and if the value of `mount` is not `/var/reana`, let's refuse to do it... just to be safe...

reana/reana_dev/cluster.py
```diff
@@ -376,11 +376,35 @@ def cluster_unpause():
...
+    \b
+    Example:
+       $ reana-dev cluster-delete -m /var/reana:/var/reana
+    """
+    cmds = []
+    # delete cluster
+    cmds.append("kind delete cluster")
+    # remove only local paths which starts with /var/reana for safety
+    for mount in mounts:
+        local_path = mount.split(":")[0]
```
tiborsimko, line 399: Actually I realised we should test `cluster_node_path` for starting with '/var/reana', because on some systems, users may need to do `cluster-create -m /opt/mybigdisk/reana:/var/reana` in order to mount a big local disk from a non-standard location into cluster's `/var/reana`.  I'll amend this while merging...

tiborsimko (approved):

PR #399, opened 2020-09-07
cli: fix release-pypi

diegodelemos:
* The subprocess call to `twine` does not ask for user and password
  which makes the command fail. Using TWINE config environment
  variables the execution can be done totally unattended.

### DIFF
--- a/reana/reana_dev/release.py
+++ b/reana/reana_dev/release.py
@@ -8,6 +8,7 @@
 
 """`reana-dev`'s release commands."""
 
+import os
 import sys
 from time import sleep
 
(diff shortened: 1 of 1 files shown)


tiborsimko:
> The subprocess call to twine does not ask for user and password
which makes the command fail. 

I was using `~/.pypirc`, so the `twine` command was working fine unattended.  Perhaps switch to using that as well?

Otherwise, if you want to add support for `TWINE_*` environment variables, please make it optional so that the above technique would still work.

diegodelemos:
I opted to not consider `~/.pypirc` as it contains the password in plain text (and the documentation page shows a [warning about it](https://packaging.python.org/specifications/pypirc/)). I think keeping support for `TWINE_*` env variables makes sense as I don't feel comfortable with having that plain text password for the REANA organisation on PyPI on my laptop. Said that, I will make the environment variables optional.

tiborsimko:
> I opted to not consider ~/.pypirc as it contains the password in plain text

(1) It does not need to contain password if you use [keyring](https://twine.readthedocs.io/en/latest/#keyring-support). Here is my setup:

```console
$ cat ~/.pypirc
[distutils]
index-servers=
    pypi
    test

[pypi]
username = johndoe
```

and:

```console
$ keyring set https://upload.pypi.org/legacy/ johndoe
```

In this way, there is no plain text password.

(2) Your laptop hard disk is probably having full-encryption in addition as well, isn't it?

diegodelemos:
(1) I didn't follow the keyring path because I thought it would be a macos only solution, nice to know, thanks for sharing.

(2) Of course :)

Issue #396, opened 2020-09-07
cli: introduce sanity check for `--mode`


tiborsimko:
Several `reana-dev` CLI commands use `--mode` argument:

```console
   --mode TEXT             In which mode to run REANA cluster?
                          (releasehelm,releasepypi,latest,debug)
                          [default=latest]

```

However, these are not checked, so somebody can try a non-existing mode:

```console
$ reana-dev cluster-create --mode=xyzzy
```

or even make a typo:

```console
$ reana-dev cluster-create --mode=devel
```

(meaning `debug`).

Current behaviour: the value of `--mode` option is not checked. This can be misleading.

What would be good is to check the passed value, and if it is not one of the known ones, stop and warn the user.

```console
$ reana-dev cluster-create --mode=devel
Error: `devel` is not one of supported modes. Please use `--help` to see the list of available modes. Exiting.
```

PR #400, opened 2020-09-08
cli: introduce sanity check for `--mode` #396

audrium:
closes https://github.com/reanahub/reana/issues/396

### DIFF
--- a/reana/config.py
+++ b/reana/config.py
@@ -89,6 +89,9 @@ REPO_LIST_SHARED = [
 WORKFLOW_ENGINE_LIST_ALL = ["cwl", "serial", "yadage"]
 """List of supported workflow engines."""
 
+CLUSTER_DEPLOYMENT_MODES = ["releasehelm", "releasepypi", "latest", "debug"]
+"""List of supported modes to run a REANA cluster."""
+
 COMPONENT_PODS = {
     "reana-db": "reana-db",
     "reana-message-broker": "reana-message-broker",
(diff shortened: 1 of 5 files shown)


reana/reana_dev/utils.py
```diff
@@ -714,3 +715,12 @@ def get_current_tag(component):
     """
     cmd = "git describe --tags"
     return run_command(cmd, component, return_output=True, display=True)
+
+
+def validate_mode_option(ctx, param, value):
+    """Validate mode option value."""
+    if value not in CLUSTER_MODES:
+        raise click.BadParameter(
+            "Supported values - {}".format(", ".join(CLUSTER_MODES))
```
tiborsimko, line 724: Current output:

```
Supported values - releasehelm, releasepypi, latest, debug
```

We could format it better as follows:

```
Supported values are 'releasehelm', 'releasepypi', 'latest', 'debug'.
```

reana/config.py
```diff
@@ -89,6 +89,9 @@
 WORKFLOW_ENGINE_LIST_ALL = ["cwl", "serial", "yadage"]
 """List of supported workflow engines."""
 
+CLUSTER_MODES = ["releasehelm", "releasepypi", "latest", "debug"]
```
tiborsimko, line 92: Perhaps call it more descriptively as `CLUSTER_DEPLOYMENT_MODES`?

reana/config.py
```diff
@@ -89,6 +89,9 @@
...
+"""List of supported modes to run a reana cluster."""
```
tiborsimko, line 93: `s/reana/REANA/`

tiborsimko (approved):

PR #404, opened 2020-09-15
scripts: simpler post-installation instructions

tiborsimko:

### DIFF
--- a/helm/reana/templates/NOTES.txt
+++ b/helm/reana/templates/NOTES.txt
@@ -21,14 +21,14 @@ finalise its configuration.
 
 4. Try to run your first REANA example:
 
-    $ firefox https://{{ default "localhost:30443" .Values.reana_url }}
+    $ firefox https://{{ default "localhost:30443" .Values.reana_hostname }}
 
    Or, using command line:
 
     $ # install REANA client
     $ pip install --user reana-client
     $ # set environment variables for REANA client
-    $ export REANA_SERVER_URL=https://{{ default "localhost:30443" .Values.reana_url }}
+    $ export REANA_SERVER_URL=https://{{ default "localhost:30443" .Values.reana_hostname }}
     $ export REANA_ACCESS_TOKEN="$mytoken"
     $ # test connection to the REANA cluster
     $ reana-client ping
--- a/helm/reana/templates/NOTES.txt
+++ b/helm/reana/templates/NOTES.txt
@@ -1,22 +1,40 @@
+The REANA system has been installed!
 
-The REANA system has been installed:
+If you are deploying REANA for the first time, there are a few steps left to
+finalise its configuration.
 
-If you are installing REANA for the first time, there are a few steps left to finalise its configuration.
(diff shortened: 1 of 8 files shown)


helm/reana/templates/NOTES.txt
```diff
@@ -1,22 +1,20 @@
...
 
-1. Get the REANA-Server pod name:
+    $ kubectl exec deployment/reana-server -- ./scripts/create-database.sh
 
-    $ REANA_SERVER=$(kubectl get pod -l "app={{ .Release.Name }}-server" -o name -o jsonpath='{.items[0].metadata.name}')
+2. Create administrator user and corresponding access token:
 
-2. Initialise the database:
+    $ mytoken=$(kubectl exec deployment/reana-server -- flask reana-admin \
+        create-admin-user --email john.doe@example.org --password mypass1234)
```
tiborsimko, line 13: BTW this command outputs also the following warning:

```
/usr/local/lib/python3.8/site-packages/flask_sqlalchemy/__init__.py:833: FSADeprecationWarning: 
SQLALCHEMY_TRACK_MODIFICATIONS adds significant overhead and will be disabled by default 
in the future.  Set it to True or False to suppress this warning.
 warnings.warn(FSADeprecationWarning(
```

We may want to turn it off?!

helm/reana/templates/NOTES.txt
```diff
@@ -1,22 +1,20 @@
...
+3. Store administrator access token as Kubernetes secret:
 
-3. Create your administrator user and store the token:
-
-    $ kubectl exec $REANA_SERVER -- flask reana-admin create-admin-user --email john.doe@example.org --password mysecretpassword
-    <reana-admin-access-token-value>
-    $ read -s REANA_ADMIN_ACCESS_TOKEN  # paste the secret here
     $ kubectl create secret generic {{ .Release.Name }}-admin-access-token \
-      --from-literal=ADMIN_ACCESS_TOKEN="$REANA_ADMIN_ACCESS_TOKEN"
+        --from-literal=ADMIN_ACCESS_TOKEN="$mytoken"
```
tiborsimko, line 19: Hmm, perhaps we should also add step 4 saying how to set tokens and run first example?

diegodelemos (approved):

helm/reana/templates/NOTES.txt
tiborsimko: Added to https://github.com/reanahub/reana-server/pull/278

helm/reana/templates/NOTES.txt
tiborsimko: Added.

tiborsimko: BTW while doing so I noticed that we are using `reana_url` for something that is not an "URL" strictly speaking; it is rather a "hostname", because the protocol part is left out... so it would be better to call it `reana_hostname` rather than `reana_url`. Let's discuss IRL whether we change this for clarity before releasing?

PR #407, opened 2020-09-18
cli: fix run-ci

diegodelemos:


reana/reana_dev/cluster.py
```diff
@@ -319,15 +317,21 @@ def job_mounts_to_config(job_mounts):
...
-        helm_install,
-        "kubectl config set-context --current --namespace={}".format(namespace),
-        os.path.join(
-            get_srcdir("reana"),
-            f"scripts/create-admin-user.sh {instance_name} {admin_email} {admin_password}",
-        ),
-    ]
+    cmds = []
+    if mode in ("debug"):
+        cmds.append("reana-dev git-submodule --update")
```
diegodelemos, line 322: In theory, we shouldn't need this second `reana-dev git-submodule --update`, however, it is the only way I could make this work.

On the one hand, it is not logical, I've done some research and I see no clear explanation of why it happens (more info [here](https://setuptools.readthedocs.io/en/latest/setuptools.html#development-mode)). On the other hand, in our use case it makes perfect sense because it covers the case of developing submodules e.g:
```
$ reana-dev run-ci --mode debug
$ # developer creates new feature with changes in R-Commons and R-Server
$ reana-dev cluster-undeploy
$ reana-dev cluster-deploy --mode debug  # thanks to the debug flag, the new R-Commons will be fetched and updated
```

scripts/create-admin-user.sh
```diff
@@ -27,7 +27,7 @@ fi
 while [ "0" -ne "$(kubectl exec deployment/reana-db -- pg_isready -U reana -h 127.0.0.1 -p 5432 &> /dev/null && echo $? || echo 1)" ]
 do
     echo "Waiting for deployment/reana-db to be ready..."
-    sleep 5
+    sleep 10
```
tiborsimko, line 30: BTW sleep 10 should not be helping much here, since this check is done within a big loop waiting for pod to occur, we can just as well be checking every 5 seconds...

diegodelemos: It is not because of this fix, it is because during the many times I've run the command, I got many connection errors because of Posgres not being ready, so 5 seconds for my setup/laptop is not enough, I will create a separate commit.

reana/reana_dev/cluster.py
```diff
@@ -319,15 +317,21 @@ def job_mounts_to_config(job_mounts):
...
-        helm_install,
-        "kubectl config set-context --current --namespace={}".format(namespace),
-        os.path.join(
-            get_srcdir("reana"),
-            f"scripts/create-admin-user.sh {instance_name} {admin_email} {admin_password}",
-        ),
-    ]
+    cmds = []
+    if mode in ("debug"):
+        cmds.append("reana-dev python-install-eggs")
```
tiborsimko, line 322: Inverting the commands, eggs before submodule update, does the trick! :+1: :partying_face: :tada:

tiborsimko (approved):

scripts/create-admin-user.sh
tiborsimko: Hmm, for me 5 secs seems fine, but we may indeed increase a bit if necessary for some hardware.

reana/reana_dev/cluster.py
tiborsimko: > cli: fix cluster-deploy 

You could also call the commit headline "cli: fix cluster-deploy debug mode"

PR #419, opened 2020-10-05
cli: new `--base` option for `git-*` commands

### LINKED ISSUE #413: cli: add base branch name option to git-* commands
When working with various branches such as `maint-0.6`, the `reana-dev git-status` and friends are not working properly, because they show differences always against the `master` as the base branch.

The goal of this issue is to introduce a new option, perhaps `--base`, that will allow to instruct "I-am-working-on-maint-0.6-now". This will be especially needed since we'll be having a mesh of `maint-0.6`, `maint-0.7`, `maint-0.8`, `master` etc soon.

Note that this will affect many commands, virtually all of existing `git-*` ones.  We'd also need a command to quickly switch base, e.g. `reana-dev git-checkout --base maint-0.7 -c CLUSTER`, which would be a new behaviour.

tiborsimko:
Adds new `--base` option to all `git-*` commands to facilitate working
against different git base branch, such as `maint-0.7`.

Adds new command `git-checkout` that allows to switch branches in many
components at one ago.

Renames old `git-checkout` to `git-checkout-pr` accordingly.

Fixes some parallel flake8-discovered warnings.

Closes #413.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -18,12 +18,14 @@ import click
 from reana.config import (
     COMPONENTS_USING_SHARED_MODULE_COMMONS,
     COMPONENTS_USING_SHARED_MODULE_DB,
+    GIT_DEFAULT_BASE_BRANCH,
     REPO_LIST_ALL,
     REPO_LIST_SHARED,
 )
 
 from reana.reana_dev.utils import (
     bump_component_version,
+    click_add_git_base_branch_option,
     display_message,
     fetch_latest_pypi_version,
     get_current_component_version_from_source_files,
@@ -102,7 +104,9 @@ def git_is_current_version_tagged(component):
     return bool(is_version_tagged)
 
 
-def git_create_release_commit(component, next_version=None):
+def git_create_release_commit(
+    component, base=GIT_DEFAULT_BASE_BRANCH, next_version=None
+):
     """Create a release commit for the given component."""
     if "release:" in get_current_commit(get_srcdir(component)):
         display_message("Nothing to do, last commit is a release commit.", component)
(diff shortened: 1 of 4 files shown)


reana/reana_dev/git.py
```diff
@@ -569,12 +638,17 @@ def git_checkout(branch, fetch):  # noqa: D301
 @click.option(
     "--branch", "-b", nargs=2, multiple=True, help="Which PR? [component PR#]"
 )
+@click.option(
```
diegodelemos, line 641: What do you think about centralising this option [like `add_workflow_option` in REANA-Client](https://github.com/reanahub/reana-client/blob/577e5dd52f14a75bed32863ccbd419e6c2ebac0d/reana_client/cli/utils.py#L71)?

Do you think it would make sense to add the default branch value to config? I don't see why we would change it ever.. but at least it would be centralised.

tiborsimko: All updated. I have also improved reporting in case base branch does not exist in origin/upstream. Example: 

```console
$ reana-dev git-status -c r-a-vomsproxy -s --base maint-0.7
ERROR: Branch upstream/maint-0.7 does not exist.
reana-auth-vomsproxy @ master @ a315022 release: v1.0.0
```

reana/reana_dev/utils.py
```diff
@@ -363,9 +365,9 @@ def is_last_commit_tagged(package):
     return bool(tag)
 
 
-def is_feature_branch(component):
-    """Check whether component current branch is different from master."""
-    return get_current_branch(get_srcdir(component)) != "master"
+def is_feature_branch(component, base="master"):
```
diegodelemos, line 368:
```suggestion
def is_feature_branch(component, base=GIT_DEFAULT_BASE_BRANCH):
```

diegodelemos (commented):
LGTM, not tested locally yet.

diegodelemos (approved):

PR #421, opened 2020-10-07
helm: add cronjob to update disk quota nightly

mvidalgarcia:
closes reanahub/reana-workflow-controller#333

### DIFF
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -113,3 +113,7 @@ traefik:
   ssl:
     enabled: true
     generateTLS: true
+
+# Quota
+quota:
+  disk_update: "0 3 * * *" # everyday at 3am
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -54,7 +54,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `notifications.email_config.smtp_server`                 | SMTP email server host                                                               | None                                            |
 | `notifications.email_config.smtp_port`                   | SMTP email server port                                                               | None                                            |
 | `notifications.enabled`                                  | Enable REANA system events notifications                                             | false                                           |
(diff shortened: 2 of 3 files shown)


helm/reana/values.yaml
```diff
@@ -113,3 +113,7 @@ traefik:
   ssl:
     enabled: true
     generateTLS: true
+
+# Quota
+quota:
```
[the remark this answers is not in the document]
diegodelemos: I see that,
1. Calculating the resource usage periodically
2. Calculating the resource usage always and always showing it

are two separate things in terms of configuration. While (1) is optional/configurable (2) should always be there even if the quotas have no limits, as I understood from Tibor explanations. (I am not sure whether we are having a naming/misundertanding issue implying that quotas have to always have limits, which if read by the dictionary it is true, but as we have been talking during the sprint we allow "infinite" limit to be a limit).

Regarding the shape of the config looks good, since it will allow us to grow in the direction of `quota.default_user_limit.cpu`, `quota.default_user_limit.disk` etc...

helm/reana/values.yaml
```diff
@@ -113,3 +113,7 @@ traefik:
...
+  disk_update: "0 3 * * *" # everyday at 3am
```
mvidalgarcia, line 119: TBD. I chose a _random_ night time, perhaps it's better to run it pretty early in the morning so users can have fresh quotas? 6am? 7am?

tiborsimko: The users can be living in all time zones (e.g. West Coast USA), so calculating around 2am-5am is usually the best.  One thing to pay attention to is to "calculate until midnight past day" so that users have some clear sharp time reference.  BTW we could also display this information visibly in the dashboard, for example "Last updated 2020-10-14-23:59:59 GMT" as a grey note near the quota charts somewhere on that page.

diegodelemos: Shall we create an issue to implement this?

diegodelemos (approved):

PR #432, opened 2020-10-26
Merge branch 'maint-0.7'

tiborsimko:


helm/reana/values.yaml
```diff
@@ -52,33 +52,33 @@ components:
     docs_url: http://docs.reana.io
     forum_url: https://forum.reana.io
     imagePullPolicy: IfNotPresent
-    image: reanahub/reana-ui:0.7.0-alpha.2
+    image: reanahub/reana-ui:0.7.0
```
diegodelemos, line 55: We don't do with images as we do with the Python packages? (depending already in `0.8.0aN`)

diegodelemos (approved):

helm/reana/values.yaml
tiborsimko: Since we haven't made a consistent `0.8.0-alpha.*` release on Docker Hub yet, I thought of keeping this equal to `maint-0.7` status for now, just in case somebody would try to install Helm `master` branch perhaps... So this is a bit unlike with Python packages, where `0.8.0a0` serves mostly for us only.  (This is a bit like researchers vs developers audience concern.)

PR #433, opened 2020-10-26
ci: added github actions workflow

audrium:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,145 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Fetch history
+        run: git fetch --prune --unshallow
+
+      - name: Run chart-testing (lint)
+        id: lint
+        uses: helm/chart-testing-action@v1.1.0
```
diegodelemos, line 142: After skimming through this action docs, I see it uses [kind-action](https://github.com/helm/kind-action). We could use it to do nightly builds inside this repo [using cron job style in GH Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events). CC @tiborsimko if this would make sense we could create a ticket.

reana/reana_dev/run.py
```diff
@@ -144,10 +144,14 @@ def run_commands():
     help="Which examples to run? [default=DEMO]",
 )
 @click.option(
-    "--admin-email", required=True, help="Admin user email address",
```
diegodelemos, line 147: Are these black changes being reported locally or only in GitHub actions? If they aren't locally it might have to do with the fact that [we do not use the latest Black version](https://github.com/reanahub/pytest-reana/blob/master/setup.py#L37).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,145 @@
...
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+
+      - name: Check Python code formatting
+        run: |
+          pip install black
```
diegodelemos, line 40: We are getting an unsupported black version for REANA (we pin to less than `20` see [here](https://github.com/reanahub/pytest-reana/blob/3a2f8b49462538ce6424b033548e6853dfa7e60e/setup.py#L37)) (cc @ParthS007 as [R-Client PR](https://github.com/reanahub/reana-client/pull/437/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR32) is also affected):
```
$ pip install black | grep "black-20"
Processing /Users/rodrigdi/Library/Caches/pip/wheels/95/a4/59/10cd5378d52f92cdb45025f040e4686e10ae5217961c25fd66/black-20.8b1-py3-none-any.whl
Successfully installed black-20.8b1
```
We will need to:
- Pin `black`
or
- Install `pytest-reana` instead

I think installing `pytest-reana` is the best option in the long term so the `black` dependency is controlled there only. For `maint-X.Y` branches we would need to change this file to install `pytest-reana<X+1` cc @tiborsimko

ParthS007: I will make the change in `r-client` PR as well 👍

.github/workflows/ci.yml
tiborsimko: It might be good if resources allow... Let's recall it for later.

tiborsimko (approved):

PR #436, opened 2020-11-02
cli: add command to generate build status wiki

### LINKED ISSUE #435: wiki: update CI build links
Update CI build links in https://github.com/reanahub/reana/wiki

Introduce `maint-0.7` and `master` instead of only one `Build` column.

mvidalgarcia:
closes #435

### DIFF
--- a/reana/reana_dev/wiki.py
+++ b/reana/reana_dev/wiki.py
@@ -1,0 +1,141 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of REANA.
+# Copyright (C) 2020 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""`reana-dev`'s wiki commands."""
+
+import click
+
+from reana.config import (
+    CODECOV_REANAHUB_URL,
+    GIT_SUPPORTED_MAINT_BRANCHES,
+    GITHUB_REANAHUB_URL,
+    REANA_LIST_DEMO_ALL,
+)
+
+
+@click.group()
+def wiki_commands():
+    """Wiki commands group."""
+
+
+@wiki_commands.command(name="wiki-create-build-status-page")
+def create_build_status_page():
+    """Generate Markdown for the Build Status wiki page."""
+    sections = {
+        "researchers": {
+            "title": "For researchers",
+            "description": "Find out how you can use REANA to describe, run, preserve and reuse your analyses.",
+            "packages": {
+                "reana-client": {},
+                "docs.reana.io": {"coverage": False, "docs": False},
+                "www.reana.io": {"coverage": False, "docs": False},
+            },
+        },
+        "administrators": {
+            "title": "For administrators",
+            "description": "Install and manage the REANA reusable analysis platform on your own compute cloud.",
+            "packages": {
+                "reana-commons": {},
+                "reana-db": {},
+                "reana-job-controller": {},
+                "reana-message-broker": {},
+                "reana-server": {},
+                "reana-ui": {},
+                "reana-workflow-controller": {},
+                "reana-workflow-engine-cwl": {},
+                "reana-workflow-engine-serial": {},
+                "reana-workflow-engine-yadage": {},
+                "reana-workflow-monitor": {},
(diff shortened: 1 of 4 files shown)


reana/reana_dev/wiki.py
```diff
@@ -0,0 +1,148 @@
...
+                "reana-demo-cdci-integral-data-reduction": {},
+                "reana-demo-cms-h4l": {},
+                "reana-demo-cms-reco": {},
+                "reana-demo-cms-dimuon-mass-spectrum": {},
+                "reana-demo-fcchh-fullsim": {},
+                "reana-demo-helloworld": {},
+                "reana-demo-lhcb-d2pimumu": {},
+                "reana-demo-lhcb-mc-production": {},
+                "reana-demo-root6-roofit": {},
+                "reana-demo-worldpopulation": {},
```
diegodelemos, line 88: It would be great if for all these lists we use the same config that `reana-dev` uses in general, e.g. in this case [`REPO_LIST_DEMO`](https://github.com/reanahub/reana/blob/b962864f6fe23a2864053b7d047aa1ad33e6572a/reana/config.py#L11). CC @tiborsimko is there any reason why `REPO_LIST_DEMO` doesn't contain all demos and part of them are just in `REPO_LIST_ALL`? Maybe with this PR comes the time to harmonise it.

reana/reana_dev/wiki.py
```diff
@@ -0,0 +1,148 @@
...
+        click.echo(header)
+        click.echo(separator)
+
+    def _print_table(components, simple=False):
+        if simple:
+            headers = ["Package", "Build", "Version"]
+        else:
+            headers = [
+                "Package",
+                "`maint-0.7`",
```
diegodelemos, line 113: What do you think about extracting `maint-0.7` to a variable? e.g. `MAINT_BRANCHES = ["maint-0.7"]` so when:
- A new maint branch is created we just add it to the list.
- When we drop support for a given version, we just remove it from the list.

(_affecting code below_)

reana/reana_dev/wiki.py
tiborsimko: `REPO_LIST_DEMO` contains only known "runnable examples", i.e. ones that are fast to run locally on a typical developer box.  They are used for `reana-dev run-ci -c DEMO`.

Hence the other examples, that are "non-runnable" locally, are living outside only as part of `REPO_LIST_ALL`. 

If we want we could amend this and introduce `REPO_LIST_DEMO_RUNNABLE`, `REPO_LIST_DEMO_ALL` for example.

mvidalgarcia: Updated.

reana/reana_dev/wiki.py
```diff
@@ -0,0 +1,140 @@
...
+
+"""`reana-dev`'s wiki commands."""
+
+import click
+
+from reana.config import REANA_LIST_DEMO_ALL
+
+GITHUB_REANAHUB_URL = "https://github.com/reanahub"
+CODECOV_REANAHUB_URL = "https://codecov.io/gh/reanahub"
+MAINT_BRANCHES = ["maint-0.7"]
```
tiborsimko, line 17: Note that we have `GIT_DEFAULT_BASE_BRANCH` in `reana/config.py`.  So this variables could go there and use a similar name, perhaps `GIT_SUPPORTED_MAINT_BRANCHES`?  You could also move the other two there.

reana/reana_dev/wiki.py
```diff
@@ -0,0 +1,140 @@
...
+            "simple": True,
+            "title": "Authentication",
+            "description": "Selected authentication environments.",
+            "packages": {"reana-auth-krb5": {}, "reana-auth-vomsproxy": {},},
+        },
+        "examples": {
+            "simple": True,
+            "title": "Examples",
+            "description": "Selected reusable analysis examples.",
+            "packages": {demo: {} for demo in REANA_LIST_DEMO_ALL},
```
tiborsimko, line 80: Perhaps we need to `.sort()` them so that demo repositories are coming up alphabetically?  (thinking about that now the variable is a join of two lists)

diegodelemos (approved):

PR #443, opened 2020-11-09
cli: `git-*` commands completed with `--exclude-components` option

tiborsimko:
Adds `--exclude-components` option to many `git-*` commands, allowing
to do things like:

$ reana-dev git-status -c CLIENT -c CLUSTER --base maint-0.7 \
    --exclude-components r-a-vomsproxy,r-a-krb5

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -320,8 +320,13 @@ def git_fork(component, browser):  # noqa: D301
     default=["CLUSTER"],
     help="Which components? [shortname|name|.|CLUSTER|ALL]",
 )
+@click.option(
+    "--exclude-components",
+    default="",
+    help="Which components to exclude from command? [c1,c2,c3]",
+)
 @git_commands.command(name="git-clone")
-def git_clone(user, component):  # noqa: D301
+def git_clone(user, component, exclude_components):  # noqa: D301
     """Clone REANA source repositories from GitHub.
 
     If the ``user`` argument is provided, the ``origin`` will be cloned from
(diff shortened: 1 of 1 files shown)


reana/reana_dev/git.py
```diff
@@ -349,12 +355,14 @@ def git_clone(user, component):  # noqa: D301
...
+    :param exclude_components: List of components to exclude from command.
     :type user: str
-
+    :type component: str
+    :type exclude_components: str
     """
-    components = select_components(component)
+    if exclude_components:
+        exclude_components = exclude_components.split(",")
+    components = select_components(component, exclude_components)
```
diegodelemos, line 365: Minor: What do you think about adding a common decorator like, `@click_add_exclude_components_option`, with a callback that performs this logic, [like we do for tokens in R-Client](https://github.com/reanahub/reana-client/blob/2436dde2df70ca835afcc49ad81f1f91ce79e030/reana_client/cli/utils.py#L23-L37)?

We could even have a single option decorator for components `@click_add_components_option` with the possibility of taking a parameter `allow_exclude_components=True` so components will always be the result of select components with or without exclude components filter.

```python
@git_commands.command(name="git-fork")
@click_add_components_option
def git_fork(components, ...):
    ...
@git_commands.command(name="git-status")
@click_add_components_option(allow_exclude_components=True)
def git_status(components, ...): 
    ...
```

diegodelemos (approved):
LGTM, used a couple of commands with `--exclude-components` during REANA `0.7.1` release.

PR #446, opened 2020-11-11
cli: push to upstream on gh-pages update

diegodelemos:


reana/reana_dev/release.py
```diff
@@ -233,7 +233,7 @@ def release_helm(ctx, user):  # noqa: D301
             f"cp -f {get_srcdir(component) + os.sep + index_path}/index.yaml {gh_pages_worktree}/index.yaml && "
             f"git add index.yaml && "
             f"git commit -m 'index.yaml: {version}' && "
-            f"git push origin {github_pages_branch} && "
+            f"git push upstream {github_pages_branch} && "
```
tiborsimko, line 236: Is it safe to push to `upstream` without checking?

FWIW I was using the following statements to push to `upstream` once checked: https://github.com/reanahub/reana/wiki/Releasing#helm-charts  We can adapt accordingly.

P.S. Needs rebase against latest `master`.

diegodelemos: I see what you mean, I agree with the check.

Regarding how safe it is.. this is exactly what the chart releaser GitHub action does... but you never know and since we do it manually we might as well do a check since breaking the `index.yaml` file would effectively break the REANA Helm Chart repository and that would be dangerous.

PR #466, opened 2020-11-25
release: 0.8.0a1

mvidalgarcia:


mvidalgarcia:
Do I need to update [`prefetch-images.sh`](https://github.com/reanahub/reana/blob/master/scripts/prefetch-images.sh) or we only do this for non-alpha versions?

helm/reana/values.yaml
```diff
@@ -52,30 +52,30 @@ components:
     docs_url: http://docs.reana.io
     forum_url: https://forum.reana.io
     imagePullPolicy: IfNotPresent
-    image: reanahub/reana-ui:0.7.0
+    image: reanahub/reana-ui:0.8.0-alpha.1
```
tiborsimko, line 55: Please edit also `scripts/prefetch-images.sh` to proper 0.8.0 tags everywhere so that researcher-oriented installation instructions would work smoothly.

(Currently, the two files `helm/reana/values.yaml` and `scripts/prefetch-images.sh` are living separately but should advertise the same versions... We could enrich `reana-dev` command to update them both.)

tiborsimko: You can also update `reana/version.py` to have `0.8.0a1` there as well

tiborsimko (approved):

PR #470, opened 2020-12-16
config: deprecation of reana-workflow-monitor

tiborsimko:


mvidalgarcia (approved):
Removed also from https://github.com/reanahub/reana/wiki

PR #473, opened 2021-01-08
helm: expose user_confirmation flag to ui config

### LINKED ISSUE #472: auth: email confirmation workflow
Implement email confirmation full workflow for newly registered users.

mvidalgarcia:
closes #472

By default, user confirmation is enabled, to disable it add env var `REANA_USER_CONFIRMATION: false` in `values.yaml`:
```diff
@@ -60,6 +77,7 @@ components:
     image: reanahub/reana-server:0.7.1
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
+      REANA_USER_CONFIRMATION: false
     uwsgi:
       processes: 6
       threads: 4
```

### DIFF
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -60,6 +60,7 @@ components:
     image: reanahub/reana-server:0.7.1
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
+      REANA_USER_CONFIRMATION: true
     uwsgi:
       processes: 6
       threads: 4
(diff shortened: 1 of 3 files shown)


audrium (commented):
Maybe it's worth appending the [`components.reana_server.environment`](https://github.com/reanahub/reana/tree/master/helm/reana) default value with `REANA_USER_CONFIRMATION: true` so that user would see what could be modified for this value

audrium (approved):
LGTM!

helm/reana/values.yaml
```diff
@@ -60,6 +60,7 @@ components:
     image: reanahub/reana-server:0.7.1
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
+      REANA_USER_CONFIRMATION: true
```
diegodelemos, line 63: Minor: we are just missing a line about this new variable in the [config docs](https://github.com/reanahub/reana/tree/master/helm/reana#configuration).

mvidalgarcia: Forgot to add it as a default value, thanks!

PR #475, opened 2021-02-03
release: 0.7.2

tiborsimko:


helm/reana/values.yaml
```diff
@@ -52,12 +52,12 @@ components:
...
+    image: reanahub/reana-ui:0.7.2
   reana_db:
     enabled: true
   reana_server:
     imagePullPolicy: IfNotPresent
-    image: reanahub/reana-server:0.7.1
+    image: reanahub/reana-server:0.7.3
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
       REANA_USER_CONFIRMATION: true
```
tiborsimko, line 63: BTW it would be more precise to name the option `REANA_USER_EMAIL_CONFIRMATION`.  Dunno if it makes sense to change this at this stage... Since this affects only `reana` and `reana-server` repositories, we could consider this perhaps?

diegodelemos: I agree it is clearer, we can do it now since https://github.com/reanahub/reana-server/pull/336 isn't merged yet.

tiborsimko: Renamed everywhere.

PR #484, opened 2021-02-19
config: configurable job clean up on failure

### LINKED ISSUE #461: helm: configurable TTL pod behaviour
**Current behaviour**

REANA-Workflow-Controller uses 60 seconds default to remove finished pods:

```python
TTL_SECONDS_AFTER_FINISHED = 60
"""Threshold in seconds to clean up terminated (either Complete or Failed)
jobs."""
```

This creates troubles when investigating e.g. CERN HTCondor job submission failures, because the user runtime pods are quickly deleted.

**Expected behaviour**

It would be good to do one or more of the following:

1. In case of failures, keep the pod running and do not garbage-collect it after TTL seconds, so that admins (or user!) can inspect the error. (1a) This would also need a way to safely `kubectl delete job xxx` later (without initiating the workflow restart or some such). (1b) Or we could simply introduce two TTL values, 60 seconds default for successfully finished pods and 86400 seconds for failed pods.
2. Expose TTL value (or values) to the cluster admin, for example via Helm deployment options, so that admins can decide upon best values and eventually redeploy the cluster in times of need.
3. Expose TTL value (or values) to users, so that they can override defaults in their `reana.yaml` files as a new operational option. (This resembles the asset expiry expose task https://github.com/reanahub/reana-workflow-controller/issues/145).

diegodelemos:
* Closes #461.

The PR group to solve #461 is composed of:
- New Helm config to govern the feature: https://github.com/reanahub/reana/pull/484
- New central REANA-Commons config variable: https://github.com/reanahub/reana-commons/pull/249
- New central logic to determine whether to cleanup jobs: https://github.com/reanahub/reana-db/pull/121
- Usage of the new cleanup logic in RWC and RJC:
    - https://github.com/reanahub/reana-workflow-controller/pull/362
    - Second commit https://github.com/reanahub/reana-job-controller/pull/297
- Renaming of `succeeded` status inside RJC to `finished` following the [standard we set in REANA-DB](https://github.com/reanahub/reana-db/blob/a3adc6866ee3233e1c817825709979eb5b6165a2/reana_db/models.py#L291)
    - First commit in https://github.com/reanahub/reana-job-controller/pull/297
    - https://github.com/reanahub/reana-workflow-engine-yadage/pull/179
    - https://github.com/reanahub/reana-workflow-engine-serial/pull/133
    - https://github.com/reanahub/reana-workflow-engine-cwl/pull/169

Instructions to test:
- Deploy REANA with the new config variable, e.g.:
```diff
~/reana $ git diff                                                                                                                                                                                     [2:55:02]
diff --git a/helm/reana/values.yaml b/helm/reana/values.yaml
index 8ed0b7e..12a4ee7 100644
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -69,6 +69,7 @@ components:
     image: reanahub/reana-workflow-controller:0.7.1
     environment:
       SHARED_VOLUME_PATH: /var/reana
+      REANA_KUBERNETES_JOBS_KEEP_ON_STATUS_LIST: failed,finished
   reana_workflow_engine_cwl:
     image: reanahub/reana-workflow-engine-cwl:0.7.2
   reana_workflow_engine_yadage:
```
- Run a workflow which will fail in some step, e.g.:
```diff
~/reana-demo-root6-roofit $ git diff
diff --git a/reana.yaml b/reana.yaml
index b1e8a2b..7fbdfac 100644
--- a/reana.yaml
+++ b/reana.yaml
@@ -18,7 +18,7 @@ workflow:
       - name: fitdata
         environment: 'reanahub/reana-env-root6:6.18.04'
         commands:
-        - root -b -q 'code/fitdata.C("${data}","${plot}")'
+        - unknown
 outputs:
   files:
     - results/plot.png
~/reana-demo-root6-roofit $ reana-client run
```
- Observe `kubectl get pods -w`
    - We should see that the workflow and its two jobs (one `failed`, one `finished`) are not cleaned up (or any other configured combination)
    - One can play changing `REANA_KUBERNETES_JOBS_KEEP_ON_STATUS_LIST` with other values for terminated workflows, for example with the current setup we should expect:
        - `REANA_KUBERNETES_JOBS_KEEP_ON_STATUS_LIST: failed`, just `fitdata` should remain
        - `REANA_KUBERNETES_JOBS_KEEP_ON_STATUS_LIST: finished`, just `gendata` should remain
       
       Note: one can identify if the job is gendata or fitdata from the logs: `kubectl logs reana-run-job-xxxx-yyy | grep "fitdata.C"` or `kubectl logs reana-run-job-xxxx-yyy | grep "gendata.C"`


helm/reana/README.md
```diff
@@ -34,6 +34,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `components.reana_workflow_controller.imagePullPolicy`   | REANA-Workflow-Controller image pull policy                                          | IfNotPresent                                    |
 | `components.reana_workflow_controller.environment.REANA_JOB_HOSTPATH_MOUNTS`   | JSON list of optional hostPath mounts, for all user jobs. Each mount object has a key `name` (name of the mount), `hostPath` (path to the directory to be mounted from the Kubernetes nodes) and `mountPath` (path inside the job containers where the `hostPath` will be mounted)            | None                                            |
+| `components.reana_workflow_controller.environment.REANA_KUBERNETES_JOBS_KEEP_ON_STATUS_LIST` | Keep Kubernetes jobs (workflow engines and jobs) with the configured statuses alive e.g. `"finished,failed"`. Possible statuses `reana_db.models.WorfkflowStatus` and `reana_db.models.JobStatus` see in [REANA-DB documentation](https://reana-db.readthedocs.io/en/latest/). | None |
```
diegodelemos, line 37: The link to REANA-DB cannot be to the correct version, shall I just switch to point to a permalink to the file on GitHub? or just mention REANA-DB docs without a link? The best would definitely be to have a link to `0.7` docs.

mvidalgarcia: `https://reana-db.readthedocs.io/en/stable/` apparently points to `0.7.0` but I'm not sure if it's something automatic or configured by us. Shouldn't it point to `0.7.1` instead?

tiborsimko: Forced the rebuilt of `stable`, now it is 0.7.1 indeed.

(That said, we can activate builds on tags, so that they would appear explicitly named as such...)

tiborsimko: ... example: https://reana-db.readthedocs.io/en/0.7.1/

mvidalgarcia (approved):
Works as expected following the given instructions.

helm/reana/README.md
mvidalgarcia: Having docs pointing to tags looks great. OTOH, if we point to a specific version in this file we take the risk of it getting outdated when we release e.g `0.8.0`.

helm/reana/README.md
```diff
@@ -34,6 +34,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `components.reana_workflow_controller.image`             | [REANA-Workflow-Controller image](https://hub.docker.com/r/reanahub/reana-workflow-controller) to use | `reanahub/reana-workflow-controller:<chart-release-version>` |
 | `components.reana_workflow_controller.imagePullPolicy`   | REANA-Workflow-Controller image pull policy                                          | IfNotPresent                                    |
 | `components.reana_workflow_controller.environment.REANA_JOB_HOSTPATH_MOUNTS`   | JSON list of optional hostPath mounts, for all user jobs. Each mount object has a key `name` (name of the mount), `hostPath` (path to the directory to be mounted from the Kubernetes nodes) and `mountPath` (path inside the job containers where the `hostPath` will be mounted)            | None                                            |
+| `components.reana_workflow_controller.environment.REANA_RUNTIME_KUBERNETES_KEEP_ALIVE_JOBS_WITH_STATUSES` | Keep alive Kubernetes user runtime jobs depending on status (`finished` and/or `failed`). | None |
```
tiborsimko, line 37: Shouldn't the part in parentheses rather be  `(example: "finished,failed")`?

PR #487, opened 2021-03-09
cli: enrich run-example to run full test matrix

diegodelemos:
Extends `reana-dev run-example` to be able to specify compute backends, making it easier to test https://github.com/reanahub/reana-job-controller/issues/304 and allow richer CI scnearios:
```
$ reana-dev run-example --compute-backend ALL --timeout 1800
[2021-03-09T15:17:02] : Running the following matrix:
Demos: ['reana-demo-worldpopulation', 'reana-demo-helloworld', 'reana-demo-root6-roofit', 'reana-demo-atlas-recast']
Workflow engines:['serial', 'yadage', 'cwl']
Compute backends: ['kubernetes', 'htcondorcern', 'slurmcern']
Number of workflows: 36
```

Closes https://github.com/reanahub/reana-job-controller/issues/304.

Caveats:
- ~~Slow: Non-parallelisable for the moment~~
- ~~`TIMEOUT` needs to be increased due to HTCondor and Slurm taking longer to execute~~

### DIFF
--- a/reana/reana_dev/run.py
+++ b/reana/reana_dev/run.py
@@ -9,13 +9,16 @@
 """`reana-dev`'s run commands."""
 
 import os
+import subprocess
 import sys
 import time
 
 import click
 
 from reana.config import (
+    COMPUTE_BACKEND_LIST_ALL,
     EXAMPLE_LOG_MESSAGES,
+    EXAMPLE_NON_STANDARD_REANA_YAML_FILENAME,
     EXAMPLE_OUTPUT_FILENAMES,
     TIMECHECK,
     TIMEOUT,
@@ -39,15 +42,19 @@ def is_cluster_created():
     return False
 
 
-def construct_workflow_name(example, workflow_engine):
+def construct_workflow_name(example, workflow_engine, compute_backend):
     """Construct suitable workflow name for given REANA example.
 
     :param example: REANA example (e.g. reana-demo-root6-roofit)
     :param workflow_engine: workflow engine to use (cwl, serial, yadage)
+    :param compute_backend: compute backend to use (kubernetes, htcondorcern, slurmcern)
     :type example: str
     :type workflow_engine: str
+    :type compute_backend: str
     """
-    output = "{0}-{1}".format(example.replace("reana-demo-", ""), workflow_engine)
+    output = "{0}-{1}-{2}".format(
+        example.replace("reana-demo-", ""), workflow_engine, compute_backend
+    )
     return output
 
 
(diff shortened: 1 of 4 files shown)


reana/reana_dev/run.py
```diff
@@ -98,6 +104,55 @@ def select_workflow_engines(workflow_engines):
...
+
+    """
+
+    def _has_cern_secrets():
+        """Check whether the test user has the correct CERN secrets?."""
+        output = subprocess.check_output(["reana-client", "secrets-list"]).decode(
+            "UTF-8"
+        )
+        required_cern_secrets = [
+            ".keytab",
```
tiborsimko, line 124: FWIW I'm not using `.keytab`, it would be nice to make this optional, or even better to check for the existence of a file name specified in `CERN_KEYTAB` variable, based on its value.

reana/reana_dev/run.py
```diff
@@ -281,7 +348,14 @@ def run_ci(
...
 def run_example(
-    component, workflow_engine, file, timecheck, timeout, parameters, options
+    component,
+    workflow_engine,
+    compute_backend,
+    file,
+    timecheck,
+    timeout,
+    parameters,
+    options,
```
tiborsimko, line 358: Please add new options `--submit-only` which will run everything up to starting workflows.

Later we'll add `--check-only` which will verify statuses and files.

In this way the big job matrix can be sort of parallel...

reana/reana_dev/run.py
tiborsimko: Yes, the check can be better removed. In this way the command could serve to test situations such as missing CERN_USER, or missing keytab... So it is actually better if secrets are not checked at all; it would be the wrapper/caller, one layer above, that would take care of this.

tests/test_cli.py
```diff
@@ -170,10 +170,15 @@ def test_construct_workflow_name():
     from reana.reana_dev.run import construct_workflow_name
 
     for (input_value, output_expected) in (
-        (("reana", "cwl"), "reana-cwl"),
-        (("reana-demo-root6-roofit", "yadage"), "root6-roofit-yadage"),
+        (("reana", "cwl", "kubernetes"), "reana-cwl-kubernetes"),
+        (
+            ("reana-demo-root6-roofit", "yadage", "htcondorcern"),
+            "root6-roofit-yadage-htcondorcern",
+        ),
```
tiborsimko, line 177: Note: some examples use different conventions, e.g. RECAST is only available in Yadage, while the matrix assumes that `reana.yaml` always means Serial. We may want to introduce a new config variable to catch these exceptional cases so that we don't generate `atlas-recast-serial-kubernetes` but rather the proper `atlas-recast-yadage-kubernetes` name.

reana/reana_dev/run.py
```diff
@@ -297,6 +374,9 @@ def run_example(
     :param workflow_engine: The option ``workflow_engine`` can be repeated. The
                             value is the workflow engine to use to run the
                             example. [default=cwl,serial,yadage]
+    :param compute_backend: The option ``compute_backend`` can be repeated. The
+                            value is the compute backend to use to run the
+                            example. [default=kubernetes,htcondorcern,slurmcern]
```
tiborsimko, line 379: The help strings could be a bit misleading, since people might be tempted to write `-b htcondorcern,slurmcern` while the command expects multiple `-b` arguments.  Ditto for worklow engines or demos to run.  Example:

```console
$ reana-dev run-example -c r-d-helloworld -c r-d-worldpopulation -c r-d-r-roofit -b ALL -w serial -w yadage --submit-only
```

Perhaps we can amend either the cosmetics, or accept comma-separated lists,  like for `--exclude-components` ?

reana/reana_dev/run.py
```diff
@@ -279,9 +390,24 @@ def run_ci(
     help="Additional operational options for the workflow execution. "
     "E.g. CACHE=off.",
 )
+@click.option(
+    "--submit-only", is_flag=True, help="Do not wait for workflows to finish."
+)
+@click.option(
+    "--check-only", is_flag=True, help="Wait for previously submitted workflows."
```
tiborsimko, line 397: When `--check-only` is used, it would be good not to loop over statuses, as the PR does now:

```console
$ reana-dev run-example -c DEMO -b kubernetes -b htcondorcern -b slurmcern --check-only 
[2021-03-16T19:30:02] : Running the following matrix:
Demos: reana-demo-root6-roofit,reana-demo-atlas-recast,reana-demo-helloworld,reana-demo-worldpopulation
Workflow engines: cwl,serial,yadage
Compute backends: htcondorcern,slurmcern,kubernetes
[2021-03-16T19:30:07] reana-demo-root6-roofit: reana-client status -w root6-roofit-cwl-htcondorcern
NAME                            RUN_NUMBER   CREATED               STARTED               STATUS    PROGRESS
root6-roofit-cwl-htcondorcern   1            2021-03-16T18:27:58   2021-03-16T18:28:00   running   0/2     
[2021-03-16T19:30:13] reana-demo-root6-roofit: reana-client status -w root6-roofit-cwl-htcondorcern
NAME                            RUN_NUMBER   CREATED               STARTED               STATUS    PROGRESS
root6-roofit-cwl-htcondorcern   1            2021-03-16T18:27:58   2021-03-16T18:28:00   running   0/2     
...
```

but rather print the current status, so that the person running the test suite could take a "snapshot" of the current state.

IOW, in the sequential mode, we want to wait until the end of the loop for each job, and move on to the next job; but in the parallel test mode, we are interested in seeing the snapshot at some moment in time of everything that is running, without waiting for anything, and then to retake that same snapshot 15 minutes later, 30 minutes later, etc.

Can you amend not to do status sleep while doing for the `--check-only` view of the results?

tiborsimko: BTW, this also means not to stop after failures:

```console
$ reana-dev run-example -c DEMO -b kubernetes -b htcondorcern -b slurmcern --check-only
...
[2021-03-16T20:00:28] reana-demo-root6-roofit: Command 'reana-client logs -w root6-roofit-cwl-htcondorcern | grep -c 'RooChebychev::bkg'' returned non-zero exit status 1.
```

but rather continue making the "snapshot" of the test cases, and then finish with some mini-report like:

```console
[2021-03-16T20:02:27] reana-dev run-example demo summary:
- 29 demo workflows run
- 27 passed
- 2 failed (foo-cwl-slurmcern, bar-yadage-kubernetes)
```

tiborsimko: P.S. Also, can you please enrich `CHANGES.rst` with a one-liner about this improvement?

diegodelemos: @tiborsimko It should be ready now, the improvement for `--check-only` is documented in https://github.com/reanahub/reana/issues/489, thanks for the detailed description 🙏

tiborsimko (approved):

Issue #500, opened 2021-05-04
mq: make message queue persistent and resilient


diegodelemos:
Current message queue is subject to lose workflow submissions in the event of message broker getting restarted/relocated:

Plan:
- Investigate how to persist the RabbitMQ queue, e.g. https://www.rabbitmq.com/persistence-conf.html
- Test disaster recovery mode, i.e. what happens if we abruptly kill the RabbitMQ instance? e.g. `kubectl delete pod reana-message-broker-xxxxx-yyyy-zzz`

ParthS007:
Reference Links:

- https://www.rabbitmq.com/persistence-conf.html
- https://www.cloudamqp.com/blog/how-to-persist-messages-during-RabbitMQ-broker-restart.html
- https://kousiknath.medium.com/dabbling-around-rabbit-mq-persistence-durability-message-routing-f4efc696098c

PR #508, opened 2021-05-28
Mount volume for persistance storage of rabbitmq data

ParthS007:
closes https://github.com/reanahub/reana/issues/500

Testing Instructions without PR:

1. `max parallel running  workflows helm variable` -> 1
2. Launch n workflows so that some workflows remain in the ready queue.
3. For ex: If n=3, While the 1st is running and 2nd and 3rd are in the queue in `ready state`, kill the queue pod (`message-broker`)
5. The Pod will be restarting again and check `/var/lib/rabbitmq/mnesia`, there will be files with the new name and our previous data is gone.

Testing Instructions with this PR and https://github.com/reanahub/reana-commons/pull/263:

1. `max parallel running  workflows helm variable` -> 1
2. Rebuild the message broker and load the new image.
3. Before killing, using the UI, create a new durable queue and publish a durable message to see if the queue and the message stays after restart.
4. You can also test the without PR case but some message should be in the `ready state`. 
5. Kill the queue pod (`message-broker`)
6. The Pod will be restarting again and check `/var/lib/rabbitmq/mnesia` before and after, there will be files with the same name and our data is persisted.

### DIFF
--- a/helm/reana/templates/reana-message-broker.yaml
+++ b/helm/reana/templates/reana-message-broker.yaml
@@ -41,8 +41,15 @@ spec:
           name: tcp
         - containerPort: 15672
           name: management
+        volumeMounts:
+        - name: data
+          mountPath: /var/lib/rabbitmq/mnesia
       {{- if .Values.node_label_infrastructure }}
       {{- $full_label := split "=" .Values.node_label_infrastructure }}
       nodeSelector:
         {{ $full_label._0 }}: {{ $full_label._1 }}
       {{- end }}
+      volumes:
+      - name: data
+        hostPath:
+          path: /var/reana/rabbitmq/mnesia
--- a/helm/reana/templates/reana-message-broker.yaml
+++ b/helm/reana/templates/reana-message-broker.yaml
@@ -5,15 +5,24 @@ metadata:
   name: {{ include "reana.prefix" . }}-message-broker
   namespace: {{ .Release.Namespace }}
 spec:
+  type: "NodePort"
   ports:
    - port: 5672
      targetPort: 5672
      name: "tcp"
      protocol: TCP
+  {{- if .Values.debug.enabled }}
+   - port: 31672
+     targetPort: 15672
+     nodePort: 31672
+     name: "management"
+     protocol: TCP
+  {{- else }}
    - port: 15672
      targetPort: 15672
      name: "management"
      protocol: TCP
+  {{- end }}
   selector:
     app: {{ include "reana.prefix" . }}-message-broker
 ---
--- a/reana/reana_dev/cluster.py
+++ b/reana/reana_dev/cluster.py
@@ -114,6 +114,11 @@ def cluster_create(mounts, mode, worker_nodes):  # noqa: D301
                     "hostPort": 32580,
                     "protocol": "TCP",
                 },  # maildev
+                {
+                    "containerPort": 31672,
+                    "hostPort": 31672,
+                    "protocol": "TCP",
+                },  # rabbitmq
             ]
         )
 


helm/reana/templates/reana-message-broker.yaml
```diff
@@ -41,8 +41,15 @@ spec:
           name: tcp
         - containerPort: 15672
           name: management
+        volumeMounts:
+        - name: data
+          mountPath: /var/lib/rabbitmq/mnesia
```
tiborsimko, line 46: The mounting works well, but it may be confusing that `/var/reana/rabbitmq/rabbit@localhost` gets mounted into the pod as `/var/reana/rabbitmq/mnesia/rabbit@localhost`?  I'd suggest to keep the two paths exactly the same, i.e. including `mnesia` on both.

helm/reana/templates/reana-message-broker.yaml
```diff
@@ -41,8 +43,15 @@ spec:
...
       {{- if .Values.node_label_infrastructure }}
       {{- $full_label := split "=" .Values.node_label_infrastructure }}
       nodeSelector:
         {{ $full_label._0 }}: {{ $full_label._1 }}
       {{- end }}
+      volumes:
+      - name: data
+        hostPath:
+          path: /var/lib/rabbitmq/mnesia
```
tiborsimko, line 57: The fix is wrong, the host path should be `/var/reana/...`:

```diff
diff --git a/helm/reana/templates/reana-message-broker.yaml b/helm/reana/templates/reana-message-broker.yaml
index f0786ce..7c49f3e 100644
--- a/helm/reana/templates/reana-message-broker.yaml
+++ b/helm/reana/templates/reana-message-broker.yaml
@@ -52,4 +52,4 @@ spec:
       volumes:
       - name: data
         hostPath:
-          path: /var/lib/rabbitmq/mnesia
+          path: /var/reana/rabbitmq/mnesia

```

ParthS007: Fixed, thanks for the catch. 👍

helm/reana/templates/reana-message-broker.yaml
```diff
@@ -5,13 +5,15 @@ metadata:
...
   namespace: {{ .Release.Namespace }}
 spec:
+  type: "NodePort"
   ports:
    - port: 5672
      targetPort: 5672
      name: "tcp"
      protocol: TCP
-   - port: 15672
+   - port: 32585
```
tiborsimko, line 14: Shouldn't this be done conditionally only, depending on `{{ .Values.debug.enabled }}`?  

Also, changing target port from 15672 to 32585 would require to amend also `r-m-broker`'s Dockerfile.  It is best to keep internallly 15672 and amend the exposing port only (and only when debug is set).  

Option 1: let's merge the PR without port and exposure changes and you'd build on top.  Option 2: please keep it but then amend the port scenario back to 15672 and add the conditional?

ParthS007: I would go for option 2, testing it rn 👍

helm/reana/templates/reana-message-broker.yaml
```diff
@@ -5,15 +5,24 @@ metadata:
...
      name: "tcp"
      protocol: TCP
+  {{- if .Values.debug.enabled }}
+   - port: 32585
+     targetPort: 15672
+     nodePort: 32585
+     name: "management"
+     protocol: TCP
+  {{- else }}
    - port: 15672
```
audrium, line 21: I see that we are already exposing `15672` port, why do we need to expose also `32585` ?

ParthS007: You are right that we are exposing `15672` but that's only for internal communication but now we want to expose externally, that's why we need to forward it to [`nodePort`](https://kubernetes.io/docs/concepts/services-networking/service/#nodeport) and according to docs there is range of ports that we can define so I just randomly choose `32585`.

tiborsimko: The convention was to use 30000 as a base and append the usual port, hence 30080/http and 30443/https for web UI (corresponding to ports 80 and 443 by default), 31984 for wdb (corresponding to port 1984 by default)... Using this technique would call for using 35672 (or even 45672, but I guess it's better to stay in the 3xxxxx range) for RabbitMQ.

tiborsimko: Retesting after changes, I get:

```
Error: Service "reana-message-broker" is invalid: spec.ports[1].nodePort: 
Invalid value: 35672: provided port is not in the valid range. 
The range of valid ports is 30000-32767
```

We would have to amend Kubertenes's API server settings to allow higher ports, which I think is not worth the trouble.... So what about changing to 31672?

tiborsimko (approved):

PR #510, opened 2021-06-11
helm: set default `kubernetes_jobs_memory_limit`

### LINKED ISSUE #509: helm: set default memory limit value for user job containers
`REANA_KUBERNETES_JOBS_MEMORY_LIMIT` is being used in `reana-job-controller` to set the limit of the jobs that are submitted. Currently it does not have any default value. To calculate the [workflow complexity](https://github.com/reanahub/reana-server/pull/366) some default value was needed. Ideally, the default values should be set in Helm values.

More context can be found in the discussion:
_Originally posted by @tiborsimko in https://github.com/reanahub/reana-server/pull/366#discussion_r645516762_

mvidalgarcia:
If no value is provided, 4Gi is the default k8s job memory limit.

closes #509

### DIFF
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -60,6 +60,8 @@ components:
     image: reanahub/reana-server:0.8.0-alpha.1
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
+      REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY: 600
+      REANA_SCHEDULER_SECONDS_RETRY_DELAY: 60
       REANA_USER_EMAIL_CONFIRMATION: true
     uwsgi:
       processes: 6
--- a/helm/configurations/values-dev.yaml
+++ b/helm/configurations/values-dev.yaml
@@ -3,6 +3,9 @@
 components:
     reana_server:
       image: reanahub/reana-server
+      environment:
+        REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY: 60
+        REANA_SCHEDULER_SECONDS_RETRY_DELAY: 6
     reana_workflow_controller:
       image: reanahub/reana-workflow-controller
       environment:


helm/reana/values.yaml
```diff
@@ -60,7 +60,6 @@ components:
     image: reanahub/reana-server:0.8.0-alpha.1
     environment:
       REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 30
-      REANA_USER_EMAIL_CONFIRMATION: true
```
audrium, line 63: Can you explain why this was removed? I'm failing to understand, but I guess the default values should be also modified [here](https://github.com/reanahub/reana/blame/2c5df09dc9f6fa02e0321330ec79af9bbe480a0d/helm/reana/README.md#L15)

mvidalgarcia: It was accidental. Fixed.

helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -91,16 +91,12 @@ spec:
           - name: REANA_RUNTIME_SESSIONS_KUBERNETES_NODE_LABEL
             value: {{ .Values.node_label_runtimesessions }}
           {{- end }}
-          {{- if .Values.kubernetes_jobs_memory_limit }}
-          {{- $_ := required "In order to use .Values.kubernetes_jobs_memory_limit safely you should set the label .Values.node_label_runtime and label a node with it." .Values.node_label_runtime }}
```
audrium, line 95: Can you explain what are the implications of removing `node_label_runtime` check? As I understand from [this comment](https://github.com/reanahub/reana/issues/509#issuecomment-859671843), it would be better to keep it? But probably with this check it won't work? @tiborsimko

mvidalgarcia: In principle, memory limits and node labels aren't related at all, so I'm not sure this check belonged here. The main implication is that the deployment wouldn't work by default (contrary to what we have now), as it'd be required to set the node label first.

In addition, the `node_label_runtime` value name was wrong, because now we have `node_label_runtimebatch`, `node_label_runtimejobs` and `node_label_runtimesessions`.

audrium (approved):
LGTM! :rocket:

PR #514, opened 2021-06-23
helm: added scheduler requeue sleep environment variable

audrium:
closes https://github.com/reanahub/reana-server/issues/382


helm/configurations/values-dev.yaml
```diff
@@ -3,9 +3,6 @@
 components:
     reana_server:
       image: reanahub/reana-server
-      environment:
-        REANA_SCHEDULER_SECONDS_TO_WAIT_FOR_REANA_READY: 60
-        REANA_SCHEDULER_SECONDS_RETRY_DELAY: 6
```
tiborsimko, line 8: Not setting anything explicit for the DEV setup here?

mvidalgarcia:
Works well, needs rebase.

helm/configurations/values-dev.yaml
mvidalgarcia: Yeah, it's not nice to wait 15 secs in between `helloworld` runs when in dev mode.

audrium: I have added 2 seconds for DEV setup. After testing with balanced strategy it seems to be enough, works in the same way as 15 seconds

tiborsimko (approved):

PR #523, opened 2021-07-08
Fixed git-clone first time behavior

marcdiazsan:
I modified the reana-dev git-clone command to clone the repositories without looking for the reana one first. It detects if the --target-directory is empty, ask a y/n confirmation to clone there, and aborts if the directory is not empty.


reana/reana_dev/git.py
```diff
@@ -330,8 +330,13 @@ def git_fork(component, exclude_components, browser):  # noqa: D301
 @click.option(
     "--exclude-components", default="", help="Which components to exclude? [c1,c2,c3]",
 )
+@click.option(
+    "--target-directory", default=".", help="In which directory install? [.]"
+)
+
```
tiborsimko, line 336: We are using `black` Pythonic code formatter, and our CI tests are also running black-compliance checks. This is why the CI status is "red". Please plug `black` into your editor so that you could format code automatically before pushing.

reana/reana_dev/git.py
```diff
@@ -371,8 +376,40 @@ def git_clone(user, component, exclude_components):  # noqa: D301
     if exclude_components:
         exclude_components = exclude_components.split(",")
     components = select_components(component, exclude_components)
+
+    instalation_directory = os.path.realpath(target_directory+'/')
+
+    def yes_no_question():
```
tiborsimko, line 382: If we choose the `-- target-directory` option way, then we don't really need to have the Y/N check. The CLI option should be sufficient in that case.

reana/reana_dev/git.py
```diff
@@ -371,8 +376,40 @@ def git_clone(user, component, exclude_components):  # noqa: D301
...
+                return True
+            elif check[0] == 'n':
+                return False
+            else:
+                return yes_no_question()
+        except Exception as error:
+            return yes_no_question()
+
+        
+    if len(os.listdir(instalation_directory)) == 0:
```
tiborsimko, line 395: Please check whether the directory exists, otherwise the users would get a Python traceback which is not nice:

```console
$ reana-dev git-clone -c reana -u tiborsimko --target-directory /this/does/not/exist
  File "/home/tibor/.virtualenvs/reana/bin/reana-dev", line 8, in <module>
    sys.exit(reana_dev())
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/tibor/.virtualenvs/reana/lib/python3.8/site-packages/reana/reana_dev/git.py", line 395, in git_clone
    if len(os.listdir(instalation_directory)) == 0:
FileNotFoundError: [Errno 2] No such file or directory: '/this/does/not/exist'
```

You can simply check whether the directory exists, and, if not, reply something like:

```
[ERROR] The desired directory /this/does/not/exist does not exist. Exiting.
```

and then exit with status code 1 to indicate error. (`sys.exit(1)`)
```

File: reana/reana_dev/git.py [hunk 4]
```diff
@@ -371,8 +376,40 @@ def git_clone(user, component, exclude_components):  # noqa: D301
...
+        clone_answer = yes_no_question()
+    else:
+        now = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
```
tiborsimko, line 398: Perhaps you can use `display_message()` function from `utils.py` that would do the formatting nicely?  So that it is centralised in one place...

reana/reana_dev/git.py
```diff
@@ -393,6 +430,7 @@ def git_clone(user, component, exclude_components):  # noqa: D301
                 run_command(cmd, component)
 
 
+
```
tiborsimko, line 433: Thanks for the PR! Please also add yourself to `AUTHORS.rst` in alphabetical order.  If you don't have ORCID yet, please create one on https://orcid.org/

tiborsimko: Also, please change the commit headline to something like `cli: improved git-clone first time behaviour`.  You can also mention in the commit log message `Closes #123` to indicate which issue number this commit closes.

marcdiazsan: I have already made the changes, and added myself to the `AUTHORS.rst`

reana/reana_dev/utils.py
```diff
@@ -335,7 +335,7 @@ def run_command(cmd, component="", display=True, return_output=False):
         sys.exit(err.returncode)
 
 
-def display_message(msg, component=""):
+def display_message(msg, component="", color="white"):
```
tiborsimko, line 338: Is this change necessary in your setup? It may not work well with certain colour schemes to force white colour... It would be better not to touch `display_message()` since it is an unrelated change to the issue at hand.

reana/reana_dev/git.py
```diff
@@ -328,10 +336,13 @@ def git_fork(component, exclude_components, browser):  # noqa: D301
     help="Which components? [shortname|name|.|CLUSTER|ALL]",
 )
 @click.option(
-    "--exclude-components", default="", help="Which components to exclude? [c1,c2,c3]",
+    "--exclude-components",
+    default="",
+    help="Which components to exclude? [c1,c2,c3]",
 )
+@click.option("--target-directory", default=".", help="In which directory install? [.]")
```
tiborsimko, line 343: Note that the default value cannot be `.` since it would break cases of people cloning new repository onto an existing system. What we should do is simply that when `--target-directory` is not present, the code would fall back to the old behaviour of discovering top-level source directory...

reana/reana_dev/git.py
tiborsimko: See also `flake8` that discovered a warning:

```console
$ flake8 .                                                     
./reana/reana_dev/git.py:414:9: F841 local variable 'error' is assigned to but never used
```

which makes CI to fail.  

We could screen-share about these setup-related things...

marcdiazsan: Done

reana/reana_dev/git.py
marcdiazsan: Done

reana/reana_dev/git.py
marcdiazsan: Done

reana/reana_dev/git.py
marcdiazsan: Done

tests/test_cli.py
```diff
@@ -91,22 +91,22 @@ def test_select_components():
...
         # bad values:
-        (["nonsense",], []),
-        (["nonsense", "reana",], ["reana",]),
+        (["nonsense",], [],),
+        (["nonsense", "reana",], ["reana",],),
         # output uniqueness:
-        (["ALL", "reana",], REPO_LIST_ALL),
-        (["CLUSTER", "reana",], REPO_LIST_CLUSTER),
+        (["ALL", "reana",], REPO_LIST_ALL,),
+        (["CLUSTER", "reana",], REPO_LIST_CLUSTER,),
```
mvidalgarcia, line 109: I think the changes in this file are not needed.

reana/reana_dev/git.py
```diff
@@ -371,8 +372,30 @@ def git_clone(user, component, exclude_components):  # noqa: D301
     if exclude_components:
         exclude_components = exclude_components.split(",")
     components = select_components(component, exclude_components)
+
+    if "--target-directory" in sys.argv:
+        instalation_directory = os.path.realpath(target_directory + "/")
```
mvidalgarcia, line 377:
```suggestion
        installation_directory = os.path.realpath(target_directory + "/")
```

reana/reana_dev/git.py
```diff
@@ -371,8 +372,30 @@ def git_clone(user, component, exclude_components):  # noqa: D301
     if exclude_components:
         exclude_components = exclude_components.split(",")
     components = select_components(component, exclude_components)
+
+    if "--target-directory" in sys.argv:
```
mvidalgarcia, line 376: It's not needed to check against `sys.argv`. `click` already offers the capability to check its value directly, and since the default value is `.` there will always be a `target_directory`.

I believe this check can be better done in a [callback validator](https://click.palletsprojects.com/en/8.0.x/options/#callbacks-for-validation).

reana/reana_dev/git.py
```diff
@@ -371,8 +372,30 @@ def git_clone(user, component, exclude_components):  # noqa: D301
...
+            )
+            display_message(message)
+            sys.exit(1)
+
+        if len(os.listdir(instalation_directory)) != 0:
+            message = "Directory {0} not empty. Cloning aborted".format(
+                instalation_directory
+            )
+            display_message(message)
+            sys.exit(0)
```
mvidalgarcia, line 391: Shouldn't it be exit code 1 here as well?

reana/reana_dev/git.py
mvidalgarcia: We can use `os.path.join` instead of `+`.
Just curious, is it strictly necessary to include the `"/"` at the end?

reana/reana_dev/git.py
```diff
@@ -371,8 +372,30 @@ def git_clone(user, component, exclude_components):  # noqa: D301
...
+
+        if not os.path.isdir(instalation_directory):
+            message = "[ERROR] Directory {0} does not exist. Exiting.".format(
+                instalation_directory
+            )
+            display_message(message)
+            sys.exit(1)
+
+        if len(os.listdir(instalation_directory)) != 0:
+            message = "Directory {0} not empty. Cloning aborted".format(
```
mvidalgarcia, line 387:
```suggestion
            message = "Directory {0} is not empty. Cloning aborted.".format(
```

reana/reana_dev/git.py
```diff
@@ -371,8 +372,30 @@ def git_clone(user, component, exclude_components):  # noqa: D301
...
+                instalation_directory
+            )
+            display_message(message)
+            sys.exit(0)
+
     for component in components:
-        os.chdir(get_srcdir())
+        if "--target-directory" in sys.argv:
```
mvidalgarcia, line 394: Again, no need to check `sys.argv`, we can use `target_directory` directly.

reana/reana_dev/git.py
mvidalgarcia: Agree, the default value could be some falsy value, probably you can even delete the default value to be `None` when `--target-directory` is not passed (I haven't tested it).

reana/reana_dev/git.py
marcdiazsan: I agree, a callback validator is a better option

reana/reana_dev/git.py
marcdiazsan: I have just tested it and it seems to work without the `"/"`, so I just dropped it

reana/reana_dev/git.py
```diff
@@ -330,8 +331,13 @@ def git_fork(component, exclude_components, browser):  # noqa: D301
 @click.option(
     "--exclude-components", default="", help="Which components to exclude? [c1,c2,c3]",
 )
+@click.option(
+    "--target-directory",
+    callback=validate_directory,
+    help="In which directory install? [.]",
```
mvidalgarcia, line 337: * We don't need the `[.]` as there's no default value anymore.
* What about a slightly different wording? Something like "In which directory to clone?" In the end, we're not installing anything but just cloning repositories

reana/reana_dev/utils.py
```diff
@@ -790,3 +790,24 @@ def wrapper(*args, **kwargs):
         return func(*args, **kwargs)
 
     return wrapper
+
+
+def validate_directory(ctx, param, target_directory):
+    """Validate if directory is valid for first time cloning."""
+    if target_directory:
+        instalation_directory = os.path.realpath(target_directory)
```
mvidalgarcia, line 798: Should be `installation_directory` instead (double L).

What about just using the same variable to assign the real directory?

``` python
target_directory = os.path.realpath(target_directory)
```

I see that in the end, we're returning `target_directory` instead of `installation_directory`. Wouldn't be better to return the real path? Then we wouldn't need [this assignment](https://github.com/reanahub/reana/pull/523/files#diff-9953f1f34f7afe5aec70e06c77e2a69473ce7c4cb35bdbb20b3c0985dd858c91R383).

mvidalgarcia (commented):
Looks much better! 👍 
Just a couple of minor things.

PS. Needs rebase.

mvidalgarcia (commented):
I have a problem while testing it. Apparently, if I set a user (`-u`) when running the command I get this:

```console
$ mkdir ~/test_clone 
$ reana-dev git-clone -c reana-server -u mvidalgarcia --target-directory ~/test_clone
[2021-08-20T15:59:54] : git clone git@github.com:mvidalgarcia/reana-server
Cloning into 'reana-server'...
remote: Enumerating objects: 2563, done.
remote: Counting objects: 100% (360/360), done.
remote: Compressing objects: 100% (122/122), done.
remote: Total 2563 (delta 260), reused 311 (delta 236), pack-reused 2203
Receiving objects: 100% (2563/2563), 846.80 KiB | 2.32 MiB/s, done.
Resolving deltas: 100% (1901/1901), done.
[2021-08-20T15:59:56] reana-server: git remote add upstream "git@github.com:reanahub/reana-server"
fatal: not a git repository (or any of the parent directories): .git
  File "/Users/marco/.virtualenvs/r/bin/reana-dev", line 33, in <module>
    sys.exit(load_entry_point('reana', 'console_scripts', 'reana-dev')())
  File "/Users/marco/.virtualenvs/r/lib/python3.8/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/Users/marco/.virtualenvs/r/lib/python3.8/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/Users/marco/.virtualenvs/r/lib/python3.8/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/marco/.virtualenvs/r/lib/python3.8/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/marco/.virtualenvs/r/lib/python3.8/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/Users/marco/code/reanahub/reana/reana/reana_dev/git.py", line 408, in git_clone
    run_command(cmd, component)
  File "/Users/marco/code/reanahub/reana/reana/reana_dev/utils.py", line 326, in run_command
    os.chdir(get_srcdir(component))
  File "/Users/marco/code/reanahub/reana/reana/reana_dev/utils.py", line 147, in get_srcdir
    reana_srcdir = find_reana_srcdir()
  File "/Users/marco/code/reanahub/reana/reana/reana_dev/utils.py", line 122, in find_reana_srcdir
    subprocess.check_output("git rev-parse --show-toplevel", shell=True)
  File "/Users/marco/.pyenv/versions/3.8.2/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/Users/marco/.pyenv/versions/3.8.2/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'git rev-parse --show-toplevel' returned non-zero exit status 128.
$ ls ~/test_clone 
reana-server
$ cd ~/test_clone/reana-server && git remote -v
origin  git@github.com:mvidalgarcia/reana-server (fetch)
origin  git@github.com:mvidalgarcia/reana-server (push)
```

It looks like the cloning itself works fine but when it fails when creating the remote `upstream` ref.

Am I doing something wrong? Can you reproduce it?

reana/reana_dev/git.py
```diff
@@ -371,8 +377,14 @@ def git_clone(user, component, exclude_components):  # noqa: D301
     if exclude_components:
         exclude_components = exclude_components.split(",")
     components = select_components(component, exclude_components)
+
     for component in components:
-        os.chdir(get_srcdir())
+        if target_directory:
+            target_directory = os.path.realpath(target_directory)
```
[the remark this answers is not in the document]
marcdiazsan: You were right, there was an error that came while running the command of creating the upstream repo. It should be fixed now

reana/reana_dev/utils.py
```diff
@@ -790,3 +792,24 @@ def wrapper(*args, **kwargs):
...
+            )
+            click.echo(click.style(message, fg="red"), err=True)
+            ctx.exit(1)
+
+        if len(os.listdir(target_directory)) != 0:
+            message = "[ERROR] Directory {0} is not empty. Cloning aborted.".format(
+                target_directory
+            )
+            click.echo(click.style(message, fg="red"), err=True)
+            ctx.exit(1)
```
mvidalgarcia, line 814: Does the target directory have to be empty?
Imagine that one wants to clone the repo for the first time in `~/my_reana`. First, they would need to clone at least the `reana` repo to `pip install` it and be able to use the `reana-dev` tool. But then, when running `reana-dev git-clone` they would get this error message if they want to clone it in the same directory. Does it make sense?

tiborsimko: They won't actually clone "reana" repository first, since people are instructed to do first:

```console
          $ # install reana-dev developer helper script
          $ pip install git+git://github.com/reanahub/reana.git#egg=reana
```

see `reana-dev --help`.

So, they get the helper script in this way, and then they clone the whole shebang into an empty repository.

(And when they run the script next time, the original helper script will be overwritten by newly cloned "reana" one.)

mvidalgarcia: Mmh I think the CLI is expecting to find the `reana` folder somehow. I'm currently following those docs and I get the following traceback:

```console
$ reana-dev git-clone -c ALL -u mvidalgarcia

fatal: not a git repository (or any of the parent directories): .git
  File "/Users/marco/.virtualenvs/r/bin/reana-dev", line 8, in <module>
    sys.exit(reana_dev())
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/click/core.py", line 1668, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/reana/reana_dev/git.py", line 375, in git_clone
    os.chdir(get_srcdir())
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/reana/reana_dev/utils.py", line 144, in get_srcdir
    reana_srcdir = find_reana_srcdir()
  File "/Users/marco/.virtualenvs/r/lib/python3.9/site-packages/reana/reana_dev/utils.py", line 122, in find_reana_srcdir
    subprocess.check_output("git rev-parse --show-toplevel", shell=True)
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/subprocess.py", line 528, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command 'git rev-parse --show-toplevel' returned non-zero exit status 128.
```

Note the `find_reana_srcdir` call.

marcdiazsan: No, in fact that is the whole initial problem, it looks for reana resources to clone reana resources. It happens because as @tiborsimko mentioned in [this comment](https://github.com/reanahub/reana/pull/523#discussion_r666749724), when there is no `--target-directory` it falls again in the same old behavior (and issue). Maybe we can make `--target-directory` a required value...

mvidalgarcia: > it looks for reana resources to clone reana resources.

This looks wrong to me...

If we don't consider `--target-directory` at all, what is the purpose of calling `find_reana_scrdir` if the command is supposed to clone `reana` repo as well?

marcdiazsan: I agree with you, what do you @mvidalgarcia and @tiborsimko think of the workaround in my last commit?. In that case we won't call `find_reana_scrdir` and simply won't clone `reana` if it is already there

PR #527, opened 2021-07-30
helm: support new paths for workspace directories

audrium:
Temporary solution for a proof of concept to run workflows outside shared path

- Added new `-w (workspace-mounts)` option for `cluster_deploy` command to specify which paths to mount inside reana-workflow-controller, run batches and job pods

Steps to test:
1) Recreate the cluster with latest changes
```
kind delete cluster
reana-dev cluster-create -m /var/reana:/var/reana -m /tmp/reana:/tmp/reana --mode debug 
reana-dev cluster-build --exclude-components=r-ui,r-a-vomsproxy,r-a-krb5 --mode debug
reana-dev cluster-deploy -w /tmp/reana:/tmp/reana --exclude-components=r-a-vomsproxy,r-a-krb5,r-ui --mode debug --admin-email your@email.com --admin-password 123456
```

Note that we specified `-w /tmp/reana:/tmp/reana` in `cluster-deploy`.

2) Change `None` to `/tmp/reana` [here](https://github.com/reanahub/reana-workflow-controller/pull/391/files#diff-c62ab69eb0095073168ae14b880c5b63a61c9f3b2eb8f175df37ed772ba8eaa2R409)

3) Start any workflow

4) Inspect workflow_path:
```
kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SELECT workspace_path FROM __reana.workflow'
```

Workflows should run successfully in `/var/reana` and `/tmp/reana`:
```
                                            workspace_path
--------------------------------------------------------------------------------------------------

 /tmp/reana/93834493-973a-4080-9d44-70f6fa140c68
 /var/reana/users/00000000-0000-0000-0000-000000000000/workflows/158b6f83-b942-45d2-8657-cf43b6581
2aa
(2 rows)
```

PR #531, opened 2021-08-12
helm: mount new local workspace directories for reana-server

marcdiazsan:
Goes on top of #527

Closes reanahub/reana-workflow-engine-yadage#199


tiborsimko (approved):
Fixes the issue nicely :+1: 

We can see later about r-server vs r-w-controller responsibilities

Issue #532, opened 2021-08-17
helm: administrators to choose allowed workspace options


tiborsimko:
When deploying REANA, users can run workflows on several workspaces. Right now, any POSIX filesystem mounted on cluster nodes is supported.

In theory, users could ask for any such mounted filesystem:

```
inputs:
  options:
    workspace_root_path: /usr/local/share
```

In practice, the administrators should limit what is allowed to use.

For example, empty value is allowed and it will use the default shared filesystem (`/var/reana/users/aaaa/workflows`), or `/myshareddisk/` is allowed, etc.

This issue is about allowing REANA cluster administrators to easily choose the available prefix values via Helm.  We need a new variable for `helm/reana/README.md` named for example `workspaces.paths` that would be a list of available path prefixes, the first being the default.

The cluster would then mount automatically these values to all the necessary pods.

P.S. The users would choose which one to use via `reana.yaml`, see another issue https://github.com/reanahub/reana-client/issues/545

PR #533, opened 2021-08-24
helm: expose optional workspace paths while deploying the cluster

marcdiazsan:
Closes reanahub/reana#532


helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -248,3 +263,10 @@ spec:
           path: {{ $workspace_path._0 }}
       {{- end }}
       {{- end }}
+      {{- if .Values.workspaces }}{{ if .Values.workspaces.paths }}
+      {{- range $workspace_paths := .Values.workspaces.paths }}
+      - name: {{ $workspace_paths | quote | replace "/" "" }}
+        hostPath:
+          path: {{ $workspace_paths }}
+      {{- end }}
+      {{- end }}{{ end }}
```
audrium, line 272: Minor: empty line is missing

reana/reana_dev/cluster.py
```diff
@@ -236,7 +236,7 @@ def cluster_build(
     "--workspace-mounts",
     multiple=True,
     help="Which directories from the Kubernetes nodes to mount inside the cluster pods? "
-    "cluster_node_path:job_pod_mountpath, e.g /var/reana/mydata:/mydata",
+    "cluster_node_path:cluster_pod_mountpath, e.g /var/reana/mydata:/mydata",
```
audrium, line 239: Can you explain why this was needed? `/mydata` is `job_pod_mountpath`

helm/reana/templates/reana-server.yaml
```diff
@@ -78,6 +85,10 @@ spec:
             value: {{ .Values.quota.default_disk_limit | default 0 | quote }}
           - name: REANA_KUBERNETES_JOBS_MEMORY_LIMIT
             value: {{ .Values.kubernetes_jobs_memory_limit | default "4Gi" }}
+          {{- if .Values.workspaces }}{{ if .Values.workspaces.paths }}
+          - name: ADMIN_ALLOWED_WORKSPACES
```
audrium, line 89: Maybe `WORKSPACE_PATHS` would be more clear? WDYT?

audrium (commented):
Works nicely, just left couple of minor comments

reana/reana_dev/cluster.py
marcdiazsan: But this paths are mounted not only for the jobs, but also for cluster pods `r-w-controller` and `reana-server`. I thought that `job_pod_mountpath` was not completely correct (but cluster_pod_mountpath might also not be correct, as is not for all cluster pods.....).

helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -248,3 +263,11 @@ spec:
...
+
```
audrium, line 273: Extremely minor: there are some spaces in the last line, that's why it still shows `No new line` error icon

helm/reana/values.yaml
```diff
@@ -11,6 +11,11 @@ debug:
 eos:
   enabled: false
 
+workspaces:
+  paths:
+    - /tmp/reana
+    - /tmp/foo
```
audrium, line 17: This is nice for testing, but perhaps should be removed in the final PR version?

marcdiazsan: Oops, leftovers :sweat_smile:

audrium (approved):
Works nicely!

PR #538, opened 2021-09-15
helm: fix templates to separate cluster_node_path and pod_path while making the cluster deploy

marcdiazsan:
Closes reanahub/reana#539


helm/reana/templates/reana-server.yaml
```diff
@@ -54,9 +54,10 @@ spec:
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
           {{- range $workspace_paths := .Values.workspaces.paths }}
-          {{- if not (eq $workspace_paths "/var/reana")}}
-          - name: {{ $workspace_paths | quote |replace "/" "" }}
-            mountPath: {{ $workspace_paths }}
+          {{- if not (eq $workspace_paths "/var/reana:/var/reana")}}
```
audrium, line 57: Minor: I was checking the deployment without this if statement and it worked without any problems for me. Is this statement here because of the problems you mentioned when mounting "/var/reana" twice?

helm/reana/templates/reana-server.yaml
```diff
@@ -77,15 +78,26 @@ spec:
...
             value: {{ .Values.kubernetes_jobs_memory_limit | default "4Gi" }}
+          {{- $pathsList := list }}
+          {{- range $workspace_paths := .Values.workspaces.paths }}
+          {{- $workspace_path := split ":" $workspace_paths }}
+          {{- $pathsList = append $pathsList $workspace_path._1}}
+          {{- end }}
           - name: WORKSPACE_PATHS
-            value: {{ join "," .Values.workspaces.paths }}
-          {{- if not (eq (index .Values.workspaces.paths 0) "/var/reana")}}
+            value: {{ join "," $pathsList }}
```
audrium, line 87: Syntax of defining  `WORKSPACE_PATHS` and `DEFAULT_WORKSPACE_PATH` seems to be really complicated when we have to exclude `"/var/reana"` and pick the first list item. To do it in Helm is a bit complex, hard to read and maintain. So I would propose to simply expose `WORKSPACE_PATHS` through Helm :
```
- name: WORKSPACE_PATHS
  value: {{ join "," .Values.workspaces.paths }}
  ```
and handle all the complexity in `reana-commons` while defining these envars. WDYT?

File: helm/reana/templates/reana-workflow-controller.yaml [hunk 3]
```diff
@@ -68,10 +69,19 @@ spec:
...
-            value: {{ index .Values.workspaces.paths 0 }}
-          {{- end }}{{ end }}
+            value: {{ $default_path._1 }}
+          {{- else if ( gt (len .Values.workspaces.paths) 1)}}
+          {{- $default_paths := index .Values.workspaces.paths 1 }}
+          {{- $default_path := split ":" $default_paths }}
+          - name: DEFAULT_WORKSPACE_PATH
+            value: {{ $default_path._1 }}
+          {{- end }}
+          - name: MOUNTED_PATHS
```
audrium, line 83: Why do we need `MOUNTED_PATHS`? I guess if we could expose single `WORKSPACE_PATHS` list and take care about the default values in `reana-commons` we wouldn't need this extra list?

helm/reana/templates/reana-server.yaml
```diff
@@ -54,9 +54,10 @@ spec:
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
           {{- range $workspace_paths := .Values.workspaces.paths }}
-          {{- if not (eq $workspace_paths "/var/reana")}}
-          - name: {{ $workspace_paths | quote |replace "/" "" }}
-            mountPath: {{ $workspace_paths }}
+          {{- $workspace_path := split ":" $workspace_paths }}
+          {{- if and (not (eq $workspace_path._0 "/var/reana")) (not (eq $workspace_path._1 "/var/reana"))}}
```
audrium, line 58: Pehaps `{{- if ne $workspace_path._1 "/var/reana" }}` would be enough here? Ideally we could compare with `.Values.shared_storage.shared_volume_mount_path` not to hardcode the values

marcdiazsan: I think the comparison with both `$workspace_path._0` and `$workspace_path._1` is needed, as it would be a problem to mount `/var/reana` either as a `mountPath` or `hostPath`

helm/reana/values.yaml
```diff
@@ -13,7 +13,7 @@ eos:
 
 workspaces:
   paths:
-    - "/var/reana"
+    - "/var/reana:/var/reana"
```
audrium, line 16: Minor: I guess we will have to update `README.md` to reflect these changes

helm/reana/templates/reana-server.yaml
```diff
@@ -54,9 +54,10 @@ spec:
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
           {{- range $workspace_paths := .Values.workspaces.paths }}
-          {{- if not (eq $workspace_paths "/var/reana")}}
-          - name: {{ $workspace_paths | quote |replace "/" "" }}
-            mountPath: {{ $workspace_paths }}
+          {{- $workspace_path := split ":" $workspace_paths }}
+          {{- if and (ne $workspace_path._0 $.Values.shared_storage.shared_volume_mount_path) (ne $workspace_path._1 $.Values.shared_storage.shared_volume_mount_path)}}
```
audrium, line 58: Minor: I guess it should be:
```{{- if and (ne $workspace_path._0 $.Values.shared_storage.hostpath.root_path) (ne $workspace_path._1 $.Values.shared_storage.shared_volume_mount_path)}}```

We should compare first value with `.Values.shared_storage.hostpath.root_path` in case it will be changed

File: helm/reana/values.yaml [hunk 7]
```diff
@@ -13,7 +13,7 @@ eos:
 
 workspaces:
   paths:
-    - "/var/reana"
+    - /var/reana: /var/reana
```
audrium, line 16: I personally find key value pairs for defining workspace paths a bit counterintuitive. I guess it could easily lead to some deployment errors if administrator would specify `/var/eos:/eos` in one line without the space in between. I would propose to use strings here as before. WDYT?

helm/reana/templates/reana-server.yaml
marcdiazsan: I agree with the `$.Values.shared_storage.hostpath.root_path` comparision, but only comparing one them, might lead to try to mount a host path without its volume mount or viceversa

helm/reana/templates/reana-server.yaml
```diff
@@ -54,10 +54,13 @@ spec:
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
           {{- range $workspace_paths := .Values.workspaces.paths }}
-          {{- if not (eq $workspace_paths "/var/reana")}}
-          - name: {{ $workspace_paths | quote |replace "/" "" }}
-            mountPath: {{ $workspace_paths }}
-            mountPropagation: HostToContainer
+          {{- if eq (typeOf $workspace_paths) "map[string]interface {}" -}}
```
audrium, line 57: We could use `{{- if ne (typeOf $workspace_paths) "string" -}}` here, I guess it would be more clear approach

helm/reana/templates/reana-server.yaml
```diff
@@ -54,10 +54,13 @@ spec:
...
+          {{- fail "The workspaces.paths should be a list of strings e.g cluster_node_path:shared_volume_path" -}}
```
audrium, line 58: Minor suggestion: `{{- fail "The workspaces.paths should be a list of strings e.g node_host_path:pod_mount_path" -}}`, WDYT?

helm/reana/templates/reana-server.yaml
```diff
@@ -77,15 +80,13 @@ spec:
             value: {{ .Values.quota.default_disk_limit | default 0 | quote }}
           - name: REANA_KUBERNETES_JOBS_MEMORY_LIMIT
             value: {{ .Values.kubernetes_jobs_memory_limit | default "4Gi" }}
+          {{- $pathsDict := dict -}}
+          {{- range $workspace_paths := .Values.workspaces.paths }}
+          {{- $workspace_path := split ":" $workspace_paths }}
+          {{- $_ := set $pathsDict $workspace_path._0 $workspace_path._1 }}
+          {{- end }}
```
audrium, line 87: I noticed one more problem - we are loosing the order of paths since we are converting the list to dict here. 
If workspace paths have these values:
```yaml
workspaces:
  paths:
    - /var/reana:/var/reana
    - /var/eos:/eos
```
then the default value currently will be `/eos` which is wrong. It should be the first one from this list.
I guess the easiest way to solve it would be to pass a list here:
```yaml
  - name: WORKSPACE_PATHS
    value: {{ .Values.workspaces.paths | toJson | quote }}
```
 and to convert into dict in `reana-commons`

helm/reana/README.md
```diff
@@ -99,4 +99,4 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `quota.default_cpu_limit`                                | Default users CPU quota limit in milliseconds.                                       | None                                            |
-| `workspaces.paths`                                       | List of additional workspace paths, e.g `{cluster_node_path_1, cluster_node_path_2}`. The first one listed different from `/var/reana` will be the default workspace root path. Any POSIX filesystem mounted on cluster nodes is supported | None               |
\ No newline at end of file
+| `workspaces.paths`                                       | JSON list of additional workspace paths as strings. Each mount string is composed by a key cluster_node_path(path to the directory to be mounted from the Kubernetes nodes) and a cluster_pod_mountpath (path inside the cluster containers where the `cluster_node_path` will be mounted) e.g. `cluster_node_path:cluster_pod_mountpath`. The first value listed different from `shared_volume_path` will be the default workspace root path. Any POSIX filesystem mounted on cluster nodes is supported | None               |
```
audrium, line 102: Minor suggestion:
``
List of additional workspace paths as strings. Each mount string is composed of a key `hostPath` (path to the directory to be mounted from the Kubernetes nodes) and a `mountPath` (path inside the cluster containers where the `hostPath` will be mounted) e.g. `hostPath:mountPath`. Any POSIX filesystem mounted on cluster nodes is supported | None 
``
WDYT?

helm/reana/templates/reana-server.yaml
marcdiazsan: Why is it wrong that `/eos` is the default path? I thought the idea was to have the first in the list different from `/var/reana` as the default, or do you think of allowing the admin to choose (my choosing the order) is a better idea?

audrium: It was true before we explicitly included `/var/reana:/var/reana` to the list. Imagine that administrator would like to include `/var/eos:/eos` as an available workspace path without changing the default value. It would not be possible, since the current code would assume the new default value is `/eos` rather than `/var/reana`. If we take the first value from the list as a default one, it's up for administrator to decide which one to put first

PR #542, opened 2021-09-27
helm: explicit number of DB connections

tiborsimko:
Enriches `reana-db` helm template to define explicitly the max number
of allowed DB connections. The default value is 100 which may not be
enough for installations launching many concurrent
workflows. Increasing it to 300. Moreover, having the number
explicitly defined in the template will allow site administrators to
easily customise it further.

### DIFF
--- a/helm/reana/templates/reana-db.yaml
+++ b/helm/reana/templates/reana-db.yaml
@@ -32,6 +32,9 @@ spec:
       containers:
       - name: db
         image: postgres:9.6.2
+        args:
+        - -c
+        - max_connections=300
         ports:
         - containerPort: 5432
         env:


helm/reana/templates/reana-db.yaml
```diff
@@ -32,6 +32,9 @@ spec:
       containers:
       - name: db
         image: postgres:9.6.2
+        args:
+        - -c
+        - max_connections=300
```
tiborsimko, line 37: How to test the PR:

1. Edit the value to your liking, e.g. I was using 1000 for local stress tests.
2. Redeploy REANA cluster.
3. Check the value via:

```console
$ kubectl exec -i -t deployment/reana-db -- psql -U reana -c 'SHOW max_connections'
 max_connections 
-----------------
 1000
```

mvidalgarcia (approved):
Works as expected! 👍 

PS. Thanks for the test instructions.

PR #575, opened 2021-11-02
benchmark: split launch into submit and start commands

### LINKED ISSUE #572: benchmark: allow submission by other means
**Current behaviour**

For pMSSM NoSys tests: 

- The workflows were submitted using `recast submit` tool, which required a little modification to (i) set desired workflow name and (ii) only create and upload, not start.

- The workflows were started using `reana_bench` script, which required a little modification to not create, not upload, and not check for existence, only start workflows.

Here is the quick diff to the bench script:

```diff
@@ -62,6 +62,7 @@


 def _create_workflow(workflow: str, file: str) -> None:
+    return
     reana_specification = load_reana_spec(
         click.format_filename(file),
         access_token=REANA_ACCESS_TOKEN,
@@ -71,6 +72,7 @@


 def _upload_workflow(workflow: str) -> None:
+    return
     upload_cmd = _build_command("upload", workflow)
     subprocess.run(upload_cmd, stdout=subprocess.DEVNULL)

@@ -343,7 +345,7 @@
 def _start_benchmark(
     workflow_name: str, number_of_submissions: int, file: str, workers: int
 ) -> pd.DataFrame:
-    if _workflow_already_exists(workflow_name):
+    if False and _workflow_already_exists(workflow_name):
         raise Exception("Found duplicated workflow name. Please use unique name.")

     _create_and_upload_workflows(workflow_name, number_of_submissions, file, workers)
```

This worked well and allowed pMSSM benchmarking easily.

**Expected behaviour**

We can make this configurable so that no script modification would be necessary. 

The main idea is that we should not assume that the benchmarking tool will be the one always doing everything, from workflow creation to workflow launch.  IOW, currently, `launch = create + upload + start`, and we should allow that some of these steps are done using outside tools.

As an illustration, we could do one of the following:

1. Keep `launch` command and introduce new options `launch --create-only`, `launch --upload-only`, `launch --start-only`.

2. Drop `launch` command and separate it into `create`, `upload`, and `start` commands that will be launched by the user one after another.  (We could later have a `run` command wrapper that does everything.)  And, in the pMSSM case, the create/upload pair would be executed by another tool.

3. Notiing that having separate create and separate upload steps does not perhaps makes too much sense, as it'll be usually one single tool that will be "preparing" workflows for running (by creating them and uploading files), and another tool that will start them and measure performance, we could perhaps introduce only two new steps, `submit` (doing create+upload) and `start` (doing start.)

VMois:
- adds two new commands: submit and start
- launch works as before, performing both submit and start

closes #572

### DIFF
--- a/scripts/reana_bench.py
+++ b/scripts/reana_bench.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
@@ -42,16 +43,28 @@ def cli():
 
     Prerequisites:
 
+        - install reana-client 0.8.x, pandas and matplotlib Python packages
         - set REANA_ACCESS_TOKEN and REANA_SERVER_URL
 
-    How to launch 50 concurrent workflows and collect results:
+    How to launch 50 concurrent workflows and collect results (option 1):
 
         .. code-block:: console
 
         \b
         $ cd reana-demo-root6-roofit  # find an example of REANA workflow
-        $ reana_bench.py launch -w roofit50yadage -n 50 -f reana-yadage.yaml  # starts 50 workflows
-        $ reana_bench.py collect -w roofit50yadage  # collect results
+        $ reana_bench.py launch -w roofit50yadage -n 50 -f reana-yadage.yaml  # submit and start
+        $ reana_bench.py collect -w roofit50yadage  # collect results and save them locally
+        $ reana_bench.py analyze -w roofit50yadage  # analyzes results that were saved locally
+
+    How to launch 50 concurrent workflows and collect results (option 2):
+
+        .. code-block:: console
+
+        \b
+        $ cd reana-demo-root6-roofit  # find an example of REANA workflow
+        $ reana_bench.py submit -w roofit50yadage -n 50 -f reana-yadage.yaml  # submit, do not start
+        $ reana_bench.py start -w roofit50yadage  # start workflows
+        $ reana_bench.py collect -w roofit50yadage  # collect results and save them locally
         $ reana_bench.py analyze -w roofit50yadage  # analyzes results that were saved locally
     """
     pass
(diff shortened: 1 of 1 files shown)


scripts/reana_bench.py
```diff
@@ -379,42 +365,101 @@ def _save_original_results(workflow: str, df: pd.DataFrame):
...
 
 
-@cli.command()
-@click.option("--workflow", "-w", help="Name of the workflow", required=True, type=str)
-@click.option(
-    "--number", "-n", help="Number of workflows to start", required=True, type=int
+def submit(
+    workflow_prefix: str, number_of_workflows: int, file: str, workers: int
+) -> None:
+    """Submit multiple workflows, do not start them."""
```
tiborsimko, line 371: Please add appropriate docs for CLI help, currently the new commands are empty:

```console
$ python ~r/reana/scripts/reana_bench.py --help         
...
Commands:
  analyze  Produce various plots and derive metrics based on launch...
  collect  Collect workflows results, merge them with intermediate...
  launch   Launch benchmark for multiple workflows.
  start
  submit
```

You can also mention there that launch=submit+start so that usage will be clear.

scripts/reana_bench.py
```diff
@@ -438,7 +483,7 @@ def analyze(workflow: str, title: str) -> NoReturn:
 
 
 @cli.command()
```
tiborsimko, line 485: One comment to the beginning of the file. Since the file is executable, please start it with:

```script
#!/usr/bin/env python
```

Otherwise it tries to run in shell when users try `./reana_bench.py --help`...

tiborsimko (commented):
Tested both cases, works well :+1: 

Left one functional and one documentation-related comment.

scripts/reana_bench.py
VMois: Done. I describe `launch` as "Submit and start workflows". This is exactly how it is in the code for the `launch` command.

scripts/reana_bench.py
VMois: Done

scripts/reana_bench.py
VMois: Nice catch with docs. I always forget about them.

tiborsimko (approved):

PR #576, opened 2021-11-02
benchmark: add workflow ranges to submit, start and launch commands

### LINKED ISSUE #573: benchmark: progressive workflow submission in time
Currently, the benchmark script was focusing on a situation of running all N workflows in parallel. The workflows were submitted and then "suddenly" all started at once. This allowed to study the steady-state burn down throughput of how many workflows can be run in the cluster.

The aim of this issue is to build on top of #572 and expand the functionality of "alternative submisssions" to the "delayed submission" scenarios. This will be useful for the test of running 100k pMSSM NoSys workflows.

For example, we shall submit 10k and start them, then in an hour submit another 10k and start them, etc until all 100k are completed.  

The "orchestration" of delayed submission can be done by outer shell script or cronjob. The changes necessary to the benchmarking script are concerned notably with allowing "partial workflow run ranges" for this deployment.

For example:

- The submission will be progressive, e.g. `submit -w mytest -i 200-249` would submit only `mytest-200`, `mytest-201`, ..., `mytest-249`, without concerning itself with the status of other run numbers.

- The start will be progressive, e.g. `start -w mytest -i 210-220` will start only `mytest-210`, `mytest-211`, ..., `mytest-220`.

- The analysis will be progressive, e.g. `analyze -w mytest -i 50-100` will produce the plots only for those run numbers, regardless of the status of the remaining runs and where those 100k workflows might be, e.g. not even submitted yet.

Note that this will also necessitate the presentation changes for plots. Let's take `execution_progress.png` as an example:

Here, x-axis represents the workflow run number as they workflows were submitted/started/etc. But this number won't be representative anymore, because not all workflows are created/uploaded at the same time anymore. So there can be a "large space" between workflow run 49 and 50 and "small space" between workflow run 48 and 49 depending on the submission policy.

It would be therefore desired to better represent "real time arrow" and show when each workflow-N was created, uploaded, started, etc. This will help to create a better representation of the flow of events in each test. For example, in test A everything will be submitted in advance and only started progressively, say 100 workflows each minute; whilst in test B the submission will be done alongside running, say 50 workflows each minute, which will stress the cluster differently. We should therefore capture this on the plots.

For example, we could create one "real time graph" with the overall time flow presentation including submission, and we keep another graph with "execution progress" only that will not concern itself with submission stages, only "consumption" stages; but with "real time" axis, not "artificial run number" axis.

Tasks:

- [x] `submit`, `start`, `launch` commands support range ([PR 1](https://github.com/reanahub/reana/pull/576))
- [ ] `collect` adapted to use with ranges
- [x] `analyze` supports ranges to "zoom in" ([PR 2](https://github.com/reanahub/reana/pull/577))
- [x] `analyze` produces plots that support ranges ([PR 2](https://github.com/reanahub/reana/pull/577))

VMois:
addresses #573

**Important!** in this PR, there is no *range protection*. Meaning you should not cross/mix ranges, etc as it will introduce some strange bugs. This is not a big problem for now but we may want to address it later.

### DIFF
--- a/scripts/reana_bench.py
+++ b/scripts/reana_bench.py
@@ -36,6 +36,9 @@ REANA_ACCESS_TOKEN = os.getenv("REANA_ACCESS_TOKEN")
 # 2 or more workers could hit reana-server API rate limit sometimes
 WORKERS_DEFAULT_COUNT = 1
 
+# common datetime format
+DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S"
+
 
 @click.group()
 def cli():
@@ -52,7 +55,7 @@ def cli():
 
         \b
         $ cd reana-demo-root6-roofit  # find an example of REANA workflow
-        $ reana_bench.py launch -w roofit50yadage -n 50 -f reana-yadage.yaml  # submit and start
+        $ reana_bench.py launch -w roofit50yadage -n 1-50 -f reana-yadage.yaml  # submit and start
         $ reana_bench.py collect -w roofit50yadage  # collect results and save them locally
         $ reana_bench.py analyze -w roofit50yadage  # analyzes results that were saved locally
 
@@ -62,8 +65,8 @@ def cli():
 
         \b
         $ cd reana-demo-root6-roofit  # find an example of REANA workflow
-        $ reana_bench.py submit -w roofit50yadage -n 50 -f reana-yadage.yaml  # submit, do not start
-        $ reana_bench.py start -w roofit50yadage  # start workflows
+        $ reana_bench.py submit -w roofit50yadage -n 1-50 -f reana-yadage.yaml  # submit, do not start
+        $ reana_bench.py start -w roofit50yadage  -n 1-50 # start workflows
         $ reana_bench.py collect -w roofit50yadage  # collect results and save them locally
         $ reana_bench.py analyze -w roofit50yadage  # analyzes results that were saved locally
     """
@@ -99,11 +102,11 @@ def _build_extended_workflow_name(workflow: str, run_number: int) -> str:
 
 def _create_and_upload_workflows(
     workflow: str,
-    n: int,
+    workflow_range: (int, int),
     file: Optional[str] = None,
     workers: int = WORKERS_DEFAULT_COUNT,
 ) -> None:
-    logging.info(f"Creating and uploading {n} workflows...")
+    logging.info(f"Creating and uploading {workflow_range} workflows...")
     with concurrent.futures.ProcessPoolExecutor(max_workers=workers) as executor:
         futures = [
             executor.submit(
@@ -111,7 +114,7 @@ def _create_and_upload_workflows(
                 _build_extended_workflow_name(workflow, i),
                 file,
             )
-            for i in range(0, n)
+            for i in range(workflow_range[0], workflow_range[1] + 1)
         ]
         for future in concurrent.futures.as_completed(futures):
             # collect results, in case of exception, it will be raised here
@@ -119,7 +122,7 @@ def _create_and_upload_workflows(
 
 
 def _get_utc_now_timestamp() -> str:
-    return datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S")
+    return datetime.utcnow().strftime(DATETIME_FORMAT)
 
 
 def _start_single_workflow(workflow_name: str) -> (str, str):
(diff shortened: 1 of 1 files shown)


scripts/reana_bench.py
```diff
@@ -381,39 +383,78 @@ def _save_original_results(workflow: str, df: pd.DataFrame):
 
 
 def submit(
-    workflow_prefix: str, number_of_workflows: int, file: str, workers: int
+    workflow_prefix: str, workflow_range: (int, int), file: str, workers: int
 ) -> None:
     """Submit multiple workflows, do not start them."""
-    if _workflow_already_exists(workflow_prefix):
```
VMois, line 387: Needed to remove this condition as it will require more coding to check workflow and ranges.

scripts/reana_bench.py
```diff
@@ -52,7 +55,7 @@ def cli():
 
         \b
         $ cd reana-demo-root6-roofit  # find an example of REANA workflow
-        $ reana_bench.py launch -w roofit50yadage -n 50 -f reana-yadage.yaml  # submit and start
+        $ reana_bench.py launch -w roofit50yadage -n 1-50 -f reana-yadage.yaml  # submit and start
```
tiborsimko, line 58: Shall we keep starting numbers from zero?  If so, please change to `0-49`. Otherwise let's keep `1-50`, it may actually be more user-friendly.

scripts/reana_bench.py
```diff
@@ -381,39 +376,78 @@ def _save_original_results(workflow: str, df: pd.DataFrame):
...
 
-    if number_of_workflows == 0:
-        raise Exception("Cannot start. Workflow(s) do not exist.")
+    existing_submit_results = pd.DataFrame()
+
+    if submitted_results_path.exists():
+        logging.info("Loading existing submit results. Appending...")
+        existing_submit_results = pd.read_csv(submitted_results_path)
+
+    return existing_submit_results.append(new_submit_results, ignore_index=True)
```
tiborsimko, line 399: Note that appending from several submissions mixes up the `submit_number` in the third column:

```console
$ less test1802_submitted_results.csv                              
name,submit_date,submit_number
test1802-0,2021-11-02T17:03:02,0
test1802-1,2021-11-02T17:03:02,1
test1802-2,2021-11-02T17:05:53,0
test1802-3,2021-11-02T17:05:53,1
test1802-4,2021-11-02T17:05:53,2
test1802-10,2021-11-02T17:06:50,0
test1802-11,2021-11-02T17:06:50,1
```

I guess we don't have to keep it?  The script should be able to deduce "submit numbers" from the `<workflow-name>-<NUMBER>` workflow naming convention already, so keeping `name` and `date` in this CSV file might be enough.

scripts/reana_bench.py
tiborsimko: Skipping is good, we can give all the liberty to the outside user to submit as they wish.  (And we may pay attention during `collect` in case they submitted twice, e.g. if both `myrun-0.1` and `myrun-0.2` are in the run list...)

tiborsimko (commented):
Works well :+1: except for `submit_number`

scripts/reana_bench.py
VMois: It is fine if `submit_number` is mixed - this is known behavior. We first order by `submit_date` and after by `submit_number`. This number exists mainly because our time precision is `seconds` and you can start a couple of workflows in a second (if multiple processes are used), so I added the number to be able to know in which order workflows were started during this second.

*suggestion:* I would prefer not to remove it because it will require unnecessary code changes to a working code (in this PR). I can make it clear in a commit message that `submit number` is only valid within the range it was used in.
Another thing, I can create an issue to investigate later if the submit_number is that useful and if it is not remove it.

What would you prefer? Remove the number or keep it as it is for now?

scripts/reana_bench.py
VMois: I have no preferences. Before it was starting from zero (without justification, just happened like that). It can start from zero or from one, ranges allow that it is up to the user of a script. The question here is what we want to put as an example in docs. I would leave it as it is.

tiborsimko: If `collect` and `analyze` are beginning-number-agnostic, and are happy to take any numbers (even with holes), then we can keep 1 here.

scripts/reana_bench.py
tiborsimko: There is no need to create an issue, we can remove it in a later PR that will be touching `collect` `analyze`  (i.e. adding `-n` ranges there), if you prefer doing it that way.  I think `name` and `submit_date` are all we need to identify when a given run N was submitted. There should be no need to store the extra 0,1 that might look artificial that we we won't even use...

VMois: Okay. Let's do it in later PR. 

In addition, we will need to rename some columns as currently, they are very confusing (like `submit_date`, which in bench CLI means create+upload but in `results` it means "datetime when request to start workflow was sent")

We could have a later PR to harmonize namings + remove `submit_number`.

tiborsimko (approved):
Works well, the format refactoring to come later when `analyze -n N1-N2` is implemented.

PR #577, opened 2021-11-04
benchmark: improve execution progress plot, add workflow range to analyze

VMois:
- improves execution progress plot, now using absolute date values instead of relative time values;
- removes _pre_processed_results function as epoch time is not needed anymore for execution progress plot;
- adds "--interval" parameter to "analyze" command to customize execution progress plot datetime axis;
- renames "*_processed_results.csv" to "*_analyzed_results.csv" to align naming with "analyze" command;
- adds new column "workflow_number" in _derive_metrics function in "analyze" to be able to filter data by workflow range, now plots can be generated with subset of workflows only;
- in case of out of bound workflow range, workflows in the range will be included without throwing an error;
- plots files are generated with a suffix that includes workflow range.

addresses #573

### DIFF
--- a/scripts/reana_bench.py
+++ b/scripts/reana_bench.py
@@ -141,7 +141,7 @@ def _start_workflows_and_record_submit_dates(
     logging.info(f"Starting {workflow_range} workflows...")
     df = pd.DataFrame(columns=["name", "submit_date", "submit_number"])
     submit_number = 0
-    with concurrent.futures.ProcessPoolExecutor(max_workers=workers) as executor:
+    with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
         futures = [
             executor.submit(
                 _start_single_workflow, _build_extended_workflow_name(workflow_name, i)
--- a/scripts/reana_bench.py
+++ b/scripts/reana_bench.py
@@ -9,20 +9,23 @@
 
 """reana_bench script - benchmark script for REANA cluster"""
 
+import concurrent.futures
 import json
 import logging
 import os
 import subprocess
 import time
-import concurrent.futures
 from datetime import datetime
 from pathlib import Path
-from typing import Optional, List, NoReturn
-import urllib3
+from typing import Optional, List, NoReturn, Dict, Tuple
 
 import click
+import matplotlib.dates as mdates
 import matplotlib.pyplot as plt
 import pandas as pd
+import urllib3
+from matplotlib.figure import Figure
+from matplotlib.ticker import MaxNLocator
 
 from reana_client.api.client import start_workflow, create_workflow
 from reana_client.utils import load_reana_spec
@@ -57,7 +60,7 @@ def cli():
         $ cd reana-demo-root6-roofit  # find an example of REANA workflow
         $ reana_bench.py launch -w roofit50yadage -n 1-50 -f reana-yadage.yaml  # submit and start
         $ reana_bench.py collect -w roofit50yadage  # collect results and save them locally
-        $ reana_bench.py analyze -w roofit50yadage  # analyzes results that were saved locally
+        $ reana_bench.py analyze -w roofit50yadage -n 1-50  # analyzes results that were saved locally
 
     How to launch 50 concurrent workflows and collect results (option 2):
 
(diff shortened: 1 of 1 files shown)


scripts/reana_bench.py
```diff
@@ -358,8 +473,8 @@ def _build_submitted_results_path(workflow: str) -> Path:
     return Path(f"{workflow}_submitted_results.csv")
 
 
-def _build_processed_results_path(workflow: str) -> Path:
-    return Path(f"{workflow}_processed_results.csv")
+def _build_derived_results_path(workflow: str) -> Path:
+    return Path(f"{workflow}_analyzed_results.csv")
```
tiborsimko, line 477: When you analyse a subset, e.g. `analyze -n N1-N2`, then the file is still named `analyzed_results.csv`, but it contains only the subset of runs corresponding to N1-N2, not the full set. So it should rather be named `analyzed_results_N1_N2.csv`, quite like the plots?

scripts/reana_bench.py
```diff
@@ -358,8 +473,8 @@ def _build_submitted_results_path(workflow: str) -> Path:
     return Path(f"{workflow}_submitted_results.csv")
```
tiborsimko, line 473: I still see `original_results.csv` when I run the script locally?

```console
$ ls -1 roofit5yadage_*
roofit5yadage_analyzed_results.csv
roofit5yadage_execution_progress_1_5.png
roofit5yadage_execution_status_1_5.png
roofit5yadage_histogram_pending_time_1_5.png
roofit5yadage_histogram_runtime_1_5.png
roofit5yadage_histogram_total_time_1_5.png
roofit5yadage_original_results.csv
roofit5yadage_submitted_results.csv
```

scripts/reana_bench.py
```diff
@@ -218,136 +221,248 @@ def _clean_results(df: pd.DataFrame) -> pd.DataFrame:
...
-    # dummy x axis
-    numbers = [int(i) for i in range(0, len(sorted_df))]
+        # add finished point
+        ax.plot(
+            ended_date,
+            workflow_number,
+            ".",
+            markerfacecolor="lightgreen",
+            markersize=1,
+            color="lightgreen",
```
tiborsimko, line 320: The dark blue reads a bit hard... What about lightblue instead of lightgreen to mark end? Or darkgreen instead of darkblue to mark the start?  (just cosmetics we can change later)

pMSSM-NoSys-1200Mi-10nodeb-56nodej-250max-21k8s-20211104-0sleep-1k_execution_progress_250_500

tiborsimko (commented):
Works well :+1: Please see a note about CSV file naming

scripts/reana_bench.py
VMois: Yep. It should be like that. `original results` are merged from submitted results (when we asked REANA to start workflows) and collect results. Naming is still bad. `original_results` is an output of collect so maybe `collect_results` would be a better name. WDYT?

scripts/reana_bench.py
VMois: I calculate metrics (`runtime`, `pending`, etc.) on the whole file not only subset. I filter later and produce plots with N1_N2. So putting `N1_N2` to `analyzed_results` will be misleading at this point.

scripts/reana_bench.py
VMois: Honestly, I don't have any preferences. I can try the colors that you suggested.

scripts/reana_bench.py
tiborsimko: `collected_results` then... Using  past tense after the action.  (Like for submitted, analyzed.)

scripts/reana_bench.py
tiborsimko: Fully agree, sorry my mistake, I confused files :man_facepalming:

scripts/reana_bench.py
VMois: I could also rename `submitted_results` to `started_results`. WDYT?

scripts/reana_bench.py
VMois: No problems. 

Right now, it is just easier to collect everything together and filter after. We may revisit it later.

scripts/reana_bench.py
VMois: P.S *warning*, you will need to rename already existing results to `collected_results`.

tiborsimko: We might need quite some renaming of functions and variables in the code as well :sweat: 

The logic is submit=create+upload -> submitted, start -> started, collect -> collected, analyse -> analysed. (using British English -ise over American English -ize)

But we could do it after merge as well, the PR works great :+1:

tiborsimko (approved):

PR #579, opened 2021-11-08
helm: use PVC for db deployment

audrium:
closes https://github.com/reanahub/reana-message-broker/issues/41

Uses PVC for db deployment if the storage backend is not `hostpath`:
- once deployed locally it will continue using `hostPath`
- once deployed with e.g. `cephfs` shared storage it will use `reana-shared-volume` PVC for volumes

### DIFF
--- a/helm/reana/templates/reana-db.yaml
+++ b/helm/reana/templates/reana-db.yaml
@@ -60,12 +60,19 @@ spec:
               key: password
         {{- end }}
         volumeMounts:
-          - name: data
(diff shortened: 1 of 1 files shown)


tiborsimko (approved):
Works well on CephFS share :+1: There is a certain slowdown, but notfully noticeable in the standard 1k test. So we can merge. And, for deployments with high number of users, we may envisage to provide later a dedicated node for DB and one for MQ via node labels.

PR #581, opened 2021-11-09
benchmark: allow to plot execution progress from partial results

### LINKED ISSUE #573: benchmark: progressive workflow submission in time

VMois:
- removes "submit_number" column from started_results as it is not used;
- renames "submit_date" column to "asked_to_start_date" in started_results for clarity;
- adds new column "collected_date" to collected_results which has a value of datetime when "collect" command was run;
 - plots runtime and pending lines in execution_progress even if "started" or "ended" datetimes are missing from collected_results, "collected_date" is used instead;
 - do not save analyzed_results to CSV file in "analyze" command as it is not required for anything right now.

closes #573

How to test:

1. Because this PR will bring breaking changes, it will not be possible to analyze results from previous benchmarks out-of-box so the easiest way to test is to run a new short benchmark

2. `submit` some workflows and `start` *some* of them

3. `collect -f` while started workflows are pending/running

4. run `analyze` to produce plots and check if they make sense

### DIFF
--- a/scripts/reana_bench.py
+++ b/scripts/reana_bench.py
@@ -18,6 +18,7 @@ import time
 from datetime import datetime
 from pathlib import Path
 from typing import Optional, List, NoReturn, Dict, Tuple
+from functools import partial
 
 import click
 import matplotlib.dates as mdates
@@ -76,10 +77,6 @@ def cli():
     pass
 
 
-def _build_command(command_type: str, workflow_name: str) -> List[str]:
-    return ["reana-client", command_type, "-w", workflow_name]
-
-
 def _create_workflow(workflow: str, file: str) -> None:
     reana_specification = load_reana_spec(
         click.format_filename(file),
@@ -90,7 +87,7 @@ def _create_workflow(workflow: str, file: str) -> None:
 
 
 def _upload_workflow(workflow: str) -> None:
-    upload_cmd = _build_command("upload", workflow)
+    upload_cmd = _build_reana_command("upload", workflow)
     subprocess.run(upload_cmd, stdout=subprocess.DEVNULL)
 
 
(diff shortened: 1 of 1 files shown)


tiborsimko (approved):
Works well, except for a situation when a user relaunches an aleady-launched workflow. I'll ticketise this separately so that we can progress fast with the Python API ThreadPool issue.

(Also there will be a separate colour cosmetics issue that we could address later.)

PR #585, opened 2021-11-09
benchmark: use reana-client API for upload instead of CLI

### LINKED ISSUE #582: benchmark: use reana-client API to upload files
- use reana-client API to upload files instead of CLI 
- because CLI will not be used anymore, use `ThreadPool` instead of `ProcessPool` for submissions

VMois:
- "submit" command uses reana-client API instead of reana-client CLI and subprocess;
- replaces ProcessPool with ThreadPool in "submit" command to speed up execution.

closes #582


scripts/reana_bench.py
```diff
@@ -87,8 +97,21 @@ def _create_workflow(workflow: str, file: str) -> None:
 
 
 def _upload_workflow(workflow: str) -> None:
-    upload_cmd = _build_reana_command("upload", workflow)
-    subprocess.run(upload_cmd, stdout=subprocess.DEVNULL)
+    reana_spec = get_workflow_specification(workflow, REANA_ACCESS_TOKEN)[
+        "specification"
+    ]
+    filenames = []
```
tiborsimko, line 103: Another area for improving create+upload performance is the memoisation. Since we are uploading the same workflow N times, and since the only thing that changes in between these uploads is the workflow name (`foo-1`, `foo-2`, etc), we can avoid all common preparatory operations such as "parsing reana.yaml specs" or "constructing list of input files and directories to upload" quite aggressively beforehand, and reuse them later. Have you tried to memoise all these manipulations so that the submission loop would have as little work left as possible? (basically just replace workfow name and call API)

VMois: I didn't try.

Some possible quick improvements:

1. `get_workflow_specificiation` calls API, I can replace it with reading `reana.yaml` file locally + using [Python built-in caching](https://docs.python.org/3/library/functools.html#functools.lru_cache) to memoize (assuming reana file will not be modified while the script is running);

2. filename list construction is a fast operation, code can improve a bit (extract `os.getcwd` to a common variable) but memoizing it will be overoptimization. In addition, I cannot modify the internals of `upload_to_server` so not much I can do with files in directories unless I will list filenames by myself but I would better leave it to `upload_to_server`.

Should I include the improvements in this PR?

P.S It is important to remember that if we overoptimize and later we will need to run different workflows together it can create a problem, so all optimizations should not decrease the flexibility of the benchmark submission.

VMois: Okay. Will add the improvements to the current PR.

P.S I mentioned over-optimization as part of memoizing filenames and directories (point 2) not loading `reana.yaml` (point 1). Memoize `reana-yaml` is a good improvement.

tiborsimko (approved):
Nice and sizeable improvements :+1: 

Submission seems as fast as start now:

ana-susy-2018-16-test-d_execution_progress_1_200

PR #589, opened 2021-11-16
benchmark: split script into files and add to entrypoints

### LINKED ISSUE #553: benchmark: migrate script to entrypoints
Migrate `reana-bench` to entrypoints, instead of using `scripts/reana_bench.py`, so that we have a nice consistent r-client, r-dev, and r-bench tools.

[Conversation 1](https://github.com/reanahub/reana/pull/541#discussion_r718276117)
[Conversation 2](https://github.com/reanahub/reana/pull/541#discussion_r724226808)

VMois:
closes #553 

How to test:

No new features were added to the script so testing is mainly about checking if splitting and entry points do not break existing functionality.

Steps:

1. Install `reana-benchmark` via `setup.py` with `pip install -e .` (inside `reana/` repository with this branch)

2. Run some small amount to workflow to test if all commands are available and work as expected. Example (inside, for example, `helloworld` demo):

```console
$ reana-benchmark submit -w test -n 1-3
$ reana-benchmark start -w test -n 1-3
$ reana-benchmark collect -w test
$ reana-benchmark analyze -w test -n 1-3  # -t option
```

and

```console
$ reana-benchmark launch -w test-2 -n 1-3
$ reana-benchmark collect -w test-2
$ reana-benchmark analyze -w test-2 -n 1-3
```


reana/reana_bench/config.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
tiborsimko, line 1: Please remove all coding declarations. We don't use them elsewhere, and I think they are nont really necessary anymore, since [Python documentation](https://docs.python.org/3/reference/lexical_analysis.html#encoding-declarations) says _"If no encoding declaration is found, the default encoding is UTF-8."_

setup.py
```diff
@@ -40,6 +40,8 @@
     "PyYAML>=5.1,<6.0",
     "semver>=2.10.2",
     "packaging>=20.4",
+    "pandas>=1.1.5",
```
tiborsimko, line 43: Please put them in a new target `bench`, so that people who wouldn't like to install benchmark don't have to get pandas etc. And plug it under `all`, so that we could do the usual `pip install '.[all]'`  to have both r-dev and r-bench.

setup.py
```diff
@@ -66,7 +68,12 @@
     url="http://www.reana.io/",
     packages=packages,
     zip_safe=False,
-    entry_points={"console_scripts": ["reana-dev = reana.reana_dev.cli:reana_dev",],},
+    entry_points={
+        "console_scripts": [
+            "reana-dev = reana.reana_dev.cli:reana_dev",
+            "reana-bench = reana.reana_bench.cli:reana_bench",
```
tiborsimko, line 74: Question: What about renaming to `reana-benchmark`?  It's longer, but more precise and unambiguous.

reana/reana_bench/config.py
VMois: I was following files from reana-dev. I will remove the coding declaration from benchmark files.

setup.py
VMois: Changed entrypoints + updated `reana-bench` to `reana-benchmark` accross files.

setup.py
VMois: What will happen if the user didn't install `benchmark` target but entry point `reana-benchmark` is still created?

reana/reana_bench/config.py
VMois: Done.

setup.py
tiborsimko: Actually the benchmarking script could run partially, i.e. launching workflows and collecting results, only the visualisation part would fail when matplotlib is not there... Kind of like yadage uses `viz`...

reana/reana_bench/config.py
VMois: Already removed it from `reana-benchmark`. I think it is a good first move. We can think about removing declarations in other places later.

setup.py
VMois: Got it. If this is not a big issue and we can move this PR forward, I am fine with that.

BTW, added `pandas` and `matplotlib` to `benchmark` target.

audrium (commented):
Works nicely! Left couple of minor comments

reana/reana_benchmark/cli.py
```diff
@@ -0,0 +1,206 @@
...
+from reana.reana_benchmark.start import start
+from reana.reana_benchmark.submit import submit
+from reana.reana_benchmark.utils import logger
+
+urllib3.disable_warnings()
+
+
+@click.group()
+def reana_benchmark():
+    r"""reana-benchmark script - runs single workflow multiple times, collects results, analyzes them.
```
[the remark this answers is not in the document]
VMois: Nice catch. I added `r` because `pydocstyle` throwing an error. I will mute `pydocstyle` then.

reana/reana_benchmark/utils.py
```diff
@@ -0,0 +1,29 @@
...
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Contains common functions shared by reana-benchmark modules."""
+
+import logging
+from datetime import datetime
+
+from reana.reana_benchmark.config import DATETIME_FORMAT
+
+logging.basicConfig(
```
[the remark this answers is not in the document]
VMois: I need `logger` otherwise a lot of unnecessary logs are coming from `reana-client`. Agree, `utils` is not the best place.  I can move `logging.basicConfig` to `cli.py` but I cannot move `logger` there because it will create a circular dependency (other modules like `submit`, etc. need a `logger`)

Some possible solutions:

1. Leave only `logger` in `utils.py`, move `logging.basic` to `cli`

2. Init `logger` in each file separately, move `logging.basic` to `cli`

3. Put `logging.basic` and `logger` to a `config.py`

4. Have a dedicated file for logging like `logging.py` or similar

I am inclined towards solutions `1` or `3`.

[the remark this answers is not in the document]
VMois: After some thinking, solution `1` seems a better option. Having basic config in `cli.py` (entry point) makes more sense.

reana/reana_benchmark/__init__.py
```diff
@@ -0,0 +1,8 @@
+#
```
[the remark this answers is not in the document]
VMois: Done

reana/reana_benchmark/cli.py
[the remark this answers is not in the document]
VMois: Done

reana/reana_benchmark/utils.py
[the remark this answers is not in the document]
VMois: Implemented solution `1`. What do you think?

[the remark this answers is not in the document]
audrium: Looks good

audrium (approved):

PR #592, opened 2021-11-18
benchmark: harmonize colours between plots

### LINKED ISSUE #588: benchmark: harmonise status colours across the plot
It'll be good to harmonise plot colours across the produced set.

**Example 1:** The "pending" status is orange in execution progress plot, but "blue" in histogram of pending times.

Using similar colours may facilitate recognition over various plots.

**Example 2:** Sometimes workflows are failing, such as here:

susy201816-submit200-sleep0-total200_execution_status_1_200

but this is not visible in the progress plot:

susy201816-submit200-sleep0-total200_execution_progress_1_200

It would be good to colour those that fail differently, e.g. in red. Dunno though whether these can be easily seen amongst the sea of blue ones that finished well -- but it would be useful for "zooming in", that is when plotting only results since time T1 until time T2.

**Example 3:** During long-running tests, such as 100k pMSSM NoSys, it'll be useful to make several "snapshots" as the tests progress. So there will be "finished" workflows, "running" workflows, "pending" workflows, "failed" workflows, etc.  It'll be useful to indicate via different colours in the progress plot where we stand.

Could we come up with a consistent colour scheme, e.g. done=green, running=blue, pending=orange, and apply it everywhere across all  the various plots in the consistent manner?

For fun, here is one "colourful" diagram from past tests, showing various statuses appearing along:
pMSSM-NoSys-1200Mi-10nodeb-65nodej-145max_execution_status

VMois:
- adds point to indicate whatever workflow finished or failed in execution_progress plot;
- use the same colours in execution_status and execution_progress plots;
- histogram plots have only one label so colours harmonization is not applied there, colours are generated automatically, usually blue.

closes #588 

How to test:

1. If you don't have already collected results from previous benchmarks you can get new ones:

```console
$ reana-benchmark launch -w color -n 1-10
$ reana-benchmark collect -w color
```

OR, you can use this CSV content and save it under `color_collected_results.csv`:

```csv
color-3,1,2021-11-16T10:24:12,2021-11-16T10:24:39,,running,2021-11-16T10:24:23,2021-11-16T10:24:47
color-2,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:24:44,finished,2021-11-16T10:24:23,2021-11-16T10:24:47
color-1,1,2021-11-16T10:24:11,2021-11-16T10:24:35,2021-11-16T10:24:42,failed,2021-11-16T10:24:23,2021-11-16T10:24:47
color-4,1,2021-11-16T10:24:12,,,queued,2021-11-16T10:24:23,2021-11-16T10:24:47
```

2. Run `reana-benchmark analyze -w color -n 1-10` to get plots.

3. This PR is modifying the style and colors of the two plots so you can check if colors/labels make sense to you, etc.

### DIFF
--- a/reana/reana_benchmark/analyze.py
+++ b/reana/reana_benchmark/analyze.py
@@ -20,7 +20,17 @@ from matplotlib.ticker import MaxNLocator
 
 from reana.reana_benchmark.collect import build_collected_results_path
 from reana.reana_benchmark.utils import logger
-from reana.reana_benchmark.config import DATETIME_FORMAT
+from reana.reana_benchmark.config import DATETIME_FORMAT, WorkflowStatus
+
+
+STATUS_TO_COLOR = {
+    WorkflowStatus.created: "grey",
+    WorkflowStatus.queued: "darkgoldenrod",
+    WorkflowStatus.pending: "darkorange",
+    WorkflowStatus.running: "blue",
+    WorkflowStatus.failed: "firebrick",
+    WorkflowStatus.finished: "forestgreen",
+}
 
 
 def _get_workflow_number_from_name(name: str) -> int:
@@ -106,9 +116,13 @@ def _build_execution_progress_plot(
 
     fig, ax = plt.subplots(figsize=(8, 4), dpi=200, constrained_layout=True)
 
+    collected_date = df["collected_date"].iloc[0]
+    collected_datetime = datetime.strptime(collected_date, DATETIME_FORMAT)
+
     for index, row in df.iterrows():
         workflow_number = row["workflow_number"]
-        collected_date = row["collected_date"]
+        workflow_status = row["status"]
+
         created_date = datetime.strptime(row["created"], DATETIME_FORMAT)
 
         asked_to_start_date_exists = not pd.isna(row["asked_to_start_date"])
@@ -120,9 +134,9 @@ def _build_execution_progress_plot(
             created_date,
             workflow_number,
             ".",
-            markerfacecolor="grey",
-            markersize=2,
-            color="grey",
+            markerfacecolor=STATUS_TO_COLOR[WorkflowStatus.created],
+            markersize=3,
+            color=STATUS_TO_COLOR[WorkflowStatus.created],
             label="1-created",
         )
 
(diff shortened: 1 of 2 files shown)


VMois:
`execution_progress` plot produced by this PR:

color_execution_progress_1_46

`execution_status` plot produced by this PR:

color_execution_status_1_46

VMois:
*thought:* I think, this PR is doing more stuff than the title ("harmonize colors") says (renaming label and adding statuses points to `execution_progress` plot). Remainder for me to keep commits/PRs more separated for the future.

audrium (commented):
Works as expected!
Couple of minor observations:
- Here are the differences between the colors used previously and in this PR. Left - old colors, right - new colors.
Screenshot from 2021-11-23 16-21-01
For me personally the colors now seems a bit too aggressive. I prefer the ones in the left (e.g. `lightblue` instead of `navy`, etc.). Of course it's very personal, it would be interesting to hear what others think.
- Number of workflows is a bit hard to read. Font size could be increased a bit and font weight could be made `bold`?
- Again, very personal, but for me coloring the status labels does not seem necessary 
- Not sure about `awaiting to run`, for me `pending` makes sense with `queued` included. @tiborsimko, WDYT?

> points at the end represent a status of a workflow. reason: having the whole red line to indicate that workflow failed (or finished) is a bit misleading in my opinion. On the other side, I assume, points are not as visible as the whole line (haven't been tested).

 Just a thought: could we increase the dot for failed workflows? So that it wouldn't be lost in a sea of finished ones, WDYT?

VMois:
> For me personally the colors now seems a bit too aggressive. I prefer the ones in the left (e.g. lightblue instead of navy, etc.). Of course it's very personal, it would be interesting to hear what others think.

I don't have preferences with colors, I can try to make them "less aggressive". My only concern is visibility on the `execution_progress` plot.

> Number of workflows is a bit hard to read. Font size could be increased a bit and font-weight could be made bold?

Agree. I can also see that "workflow" can collide with some labels making it even less visible. Will check that too.

> Again, very personal, but for me coloring the status labels does not seem necessary

I added it just because it was part of an example I used. I will remove the coloring of status labels because, honestly, no one really asked for it so no point in adding it :)

> Not sure about awaiting to run, for me pending makes sense with queued included.

Here is also my initiative that no one asked for :) I can revert it back cause even if change is needed there it deserves a separate PR.

> Just a thought: could we increase the dot for failed workflows? So that it wouldn't be lost in a sea of finished ones, WDYT?

Interesting idea, I will try to increase and see how the plot will look like.

audrium (commented):
Works great, just couple of small questions:
- I've noticed that now there are still some inconsistencies between colors in different plot types.
Here are the `execution_progress` and `execution_status` plots:
bsm_execution_progress_1_4
bsm_execution_status_1_4
- You can see that `running` color is light blue in `execution_status` and dark blue in `execution_progress` (other plots uses dark blue too). We could unify these colors across the plots.
- Pending (pending+queued) in `execution_progress` is orange and `Queued` in  `execution_status` is violet. I would expect the same color, since it's almost the same thing. WDYT?

- >> Number of workflows is a bit hard to read. Font size could be increased a bit and font-weight could be made bold?

 > Agree. I can also see that "workflow" can collide with some labels making it even less visible. Will check that too.
 
 Nice that you fixed that! I was actually referring to the white numbers which shows how many workflows falls into certain category in `execution_status` plot. Wanted to suggest to increase the font size there as well, since it's a bit hard to read. Sorry for being unclear

VMois:
> You can see that running color is light blue in execution_status and dark blue in execution_progress (other plots uses dark blue too). We could unify these colors across the plots.

Just in case, `running` workflow status and `running` line in `execution_progress` plot are a bit different things but I think it is a good idea to have the same colors for them.

> Pending (pending+queued) in execution_progress is orange and Queued in execution_status is violet. I would expect the same color, since it's almost the same thing. WDYT?

I can try to find some colors similar to orange for queued and pending statuses to align them with pending line.

> I was actually referring to the white numbers which shows how many workflows falls into certain category in execution_status plot.

Agree. Will make numbers more visible.

Tasks:

- [ ] align running line color with running status

- [ ] align queued and pending statuses with pending line

- [ ] make white numbers on the execution status plot more visible

audrium (approved):
Works nicely!

PR #593, opened 2021-11-19
benchmark: add monitor command

### LINKED ISSUE #574: benchmark: new `monitor` command for DB and K8S statuses
(stems from https://github.com/reanahub/reana/pull/541#discussion_r716529942)

**Current behaviour**

Currently, while running benchmarking script, one can monitor the DB status and the K8S status independently via a script like:

```shell
#!/bin/sh

workflow=$1

while true; do

    echo "$(date +%Y-%m-%dT%H:%M:%S) db pg_stat_activity count:"
    kubectl exec deployment/reana-db -- psql -U reana -c "SELECT COUNT(*) FROM pg_stat_activity;"

    echo "$(date +%Y-%m-%dT%H:%M:%S) db workflow.status count:"
    kubectl exec deployment/reana-db -- psql -U reana -c "SELECT status,COUNT(*) FROM __reana.workflow WHERE name LIKE '$workflow-%' GROUP BY status;"

    echo "$(date +%Y-%m-%dT%H:%M:%S) run-b pods count:"
    kubectl get pods | grep run-b | grep -v STATUS | awk '{print $3}' | sort|uniq -c|sort -rn|head -20|awk '!max{max=$1;}{r="";i=s=60*$1/max;while(i-->0)r=r"#"; printf "%25s %5d %s %s",$2,$1,r,"\n";}'

    echo "$(date +%Y-%m-%dT%H:%M:%S) run-j pods count:"
    kubectl get pods | grep run-j | grep -v STATUS | awk '{print $3}' | sort|uniq -c|sort -rn|head -20|awk '!max{max=$1;}{r="";i=s=60*$1/max;while(i-->0)r=r"#"; printf "%25s %5d %s %s",$2,$1,r,"\n";}'

    echo

    sleep 30

done
```

This gives output as follows, for one particular moment of time:

```
2021-10-31T20:48:35 db pg_stat_activity count:
 count
-------
   182
(1 row)

2021-10-31T20:48:35 db workflow.status count:
 status  | count
---------+-------
 running |   177
 queued  |    89
 pending |   134
(3 rows)

2021-10-31T20:48:36 run-b pods count:
                  Running   190 ############################################################
        ContainerCreating     3 #
2021-10-31T20:48:36 run-j pods count:
                  Running   198 ############################################################
                 Init:0/1    16 #####
          PodInitializing     7 ###

```

and, 30 seconds later:

```
2021-10-31T20:49:07 db pg_stat_activity count:
 count
-------
   212
(1 row)

2021-10-31T20:49:07 db workflow.status count:
 status  | count
---------+-------
 pending |   197
 running |   203
(2 rows)

2021-10-31T20:49:08 run-b pods count:
                  Running   217 ############################################################
2021-10-31T20:49:09 run-j pods count:
                  Running   314 ############################################################
                 Init:0/1    15 ###
          PodInitializing     7 ##
                  Pending     1 #
```

These time snapshots allow to monitor the number of DB connections, the DB statuses vs K8S statuses, the number of "Running" pods vs "Pending" pods, to see how fast the pods terminate, etc, giving a complementary picture of what's happening in the cluster.

The trouble is that this "side" monitoring is a bit "detached" from the main content output of the benchmark scripts. It'll be advantageous to better correlate this information with the workflow burn down plots.

**Expected behaviour**

We can introduce a new command `monitor --sleep 30` which would do the above automatically and which would collect the information in either the textual format above (MVP), or even better in a CSV format that will allow to plot nice DB and K8S status evolution graphs later about how the measured DB and K8S quantities evolve as a function of time.

For example, once #573 is implemented, we shall have a "real time arrow" representation of the workflow burn down in the cluster, and the DB info plots and K8S info plots will nicely complement the overall picture about  what's happening in the cluster. 

They might show graphical insight into "orange hill" and "blue spread" phenomena, such as the transition from "Running -> NotReady -> Terminating" status of workflow pods.

VMois:
This PR introduces `monitor` command to the benchmark. It will collect all defined metrics (check code for details), sleep for a defined time in seconds, and repeat the same process until is terminated *manually*. Results are saved to a JSON file to be used in the `analyze` command later.

closes #574

How to test:

1. (*optional*) Launch some workflows `reana-benchmark launch -w monitor-test -n 1-20` (**not**  `helloworld` as they are too fast, `atlas-demo` is okay);

2. Right after workflows are started, start a monitor command `reana-benchmark monitor --sleep 5 -w monitor-test`;
 
3. When workflows are finished (or around that time), terminate the command using Ctrl+C;

4. Check JSON output in `monitor-test_monitored_results.json`, you will see some metrics (currently 2) and timestamps when they were collected.

*Note:* `-w` in the `monitor` command is not tied to actual workflows started before, and added only as a convention and a way to map monitored results in `analyze` command later.

### DIFF
--- a/reana/reana_benchmark/cli.py
+++ b/reana/reana_benchmark/cli.py
@@ -19,6 +19,7 @@ from reana.reana_benchmark.config import WORKERS_DEFAULT_COUNT
 from reana.reana_benchmark.start import start
 from reana.reana_benchmark.submit import submit
 from reana.reana_benchmark.utils import logger
+from reana.reana_benchmark.monitor import monitor
 
 urllib3.disable_warnings()
 
@@ -98,7 +99,7 @@ workflow_range_option = click.option(
 concurrency_option = click.option(
     "--concurrency",
     "-c",
-    help=f"Number of workers to submit workflows, default {WORKERS_DEFAULT_COUNT}",
+    help=f"Number of workers to submit workflows [default {WORKERS_DEFAULT_COUNT}]",
     type=int,
     default=WORKERS_DEFAULT_COUNT,
 )
@@ -176,7 +177,7 @@ def launch(
 @click.option(
     "--interval",
     "-i",
-    help="Execution progress plot interval in minutes.",
+    help="Execution progress plot interval in minutes [default=10]",
     type=int,
     default=10,
 )
@@ -210,3 +211,20 @@ def collect_command(workflow: str, force: bool) -> NoReturn:
         collect(workflow, force)
     except Exception as e:
         logger.error(f"Something went wrong when collecting results: {e}")
+
+
+@reana_benchmark.command(name="monitor")
+@workflow_option
+@click.option(
+    "--sleep",
+    "-s",
+    help="Sleep in seconds between collecting metrics [default=30]",
+    type=int,
+    default=30,
+)
+def monitor_command(workflow: str, sleep: int) -> NoReturn:
+    """Monitor various metrics and record results."""
+    try:
+        monitor(workflow, sleep)
+    except Exception as e:
+        logger.error(f"Something went wrong during monitoring: {e}")
--- a/reana/reana_benchmark/monitor.py
+++ b/reana/reana_benchmark/monitor.py
@@ -1,0 +1,192 @@
+# This file is part of REANA.
+# Copyright (C) 2021 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""Responsible for monitoring K8s cluster, DB connections."""
+
+import json
+import time
+from pathlib import Path
+from typing import Dict, Any, List, Generator
+from collections import defaultdict
+import subprocess
+from abc import abstractmethod, ABC
+
+from reana.reana_benchmark.utils import get_utc_now_timestamp, logger
+
+
+class BaseMetric(ABC):
+    """Base class for other metrics."""
+
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """Name of the metric."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def _collect(self, parameters: Dict) -> Any:
+        raise NotImplementedError
+
+    def collect(self, parameters: Dict) -> Dict[str, Any]:  # noqa: D102
+        result = self._collect(parameters)
+        return {
+            self.name: result,
+        }
+
+
+class WorkflowDBStatusesMetric(BaseMetric):
+    """Count number of workflows statuses directly from DB."""
+
+    name = "workflow_db_statuses"
+
+    def _collect(self, parameters: Dict) -> Any:
+        workflow_prefix = parameters.get("workflow")
+
+        if not workflow_prefix:
+            logger.warning(
+                f"{self.name} metrics cannot find workflow parameter. Metric will not be collected."
+            )
+            return {}
+
+        cmd = [
+            "kubectl",
+            "exec",
+            "deployment/reana-db",
+            "--",
+            "psql",
+            "-U",
+            "reana",
+            "-c",
+            f"SELECT status,COUNT(*) FROM __reana.workflow WHERE name LIKE '{workflow_prefix}-%' GROUP BY status;",
+        ]
+        output = subprocess.check_output(cmd).decode("ascii")
+        result = {}
+
+        rows = output.splitlines()[2:-2]
+
+        for row in rows:
+            status, count = row.split("|")[0].strip(), int(row.split("|")[1].strip())
+            result[status] = count
+
+        return result
+
+
+class NumberOfDBConnectionsMetric(BaseMetric):
+    """Count number of server processes in REANA DB."""
+
+    name = "db_connections_number"
+
+    def _collect(self, parameters: Dict) -> Any:
+        cmd = [
+            "kubectl",
+            "exec",
+            "deployment/reana-db",
+            "--",
+            "psql",
+            "-U",
+            "reana",
+            "-c",
+            "SELECT COUNT(*) FROM pg_stat_activity;",
+        ]
+        output = subprocess.check_output(cmd).decode("ascii")
+        result = int(output.splitlines()[2].strip())
+        return result
+
+
+class WorkflowPodsMetric(BaseMetric):
+    """Count number of job and batch jobs in different phases."""
+
+    name = "workflows_pods_status"
+
+    @staticmethod
+    def _filter(pods: List[Dict], name_contains: str) -> Generator[Dict, None, None]:
+        for pod in pods:
+            name = pod.get("metadata", {}).get("name", "")
+            if name_contains in name:
+                yield pod
+
+    @staticmethod
+    def _count(pods: List[Dict], name_contains: str) -> Dict[str, int]:
+        statistics = defaultdict(lambda: 0)
+        for pod in WorkflowPodsMetric._filter(pods, name_contains):
+            phase = pod.get("status", {}).get("phase")
+            statistics[phase] += 1
+        return dict(statistics)
+
+    def _collect(self, parameters: Dict) -> Any:
+        kubectl_cmd = ("kubectl", "get", "pods", "-o", "json")
+        output = subprocess.check_output(kubectl_cmd)
+        pods = json.loads(output).get("items", [])
+
+        result = {
+            "batch_pods": self._count(pods, "run-batch"),
+            "job_pods": self._count(pods, "run-job"),
+        }
+
+        return result
+
+
+METRICS = [
+    NumberOfDBConnectionsMetric(),
+    WorkflowPodsMetric(),
+    WorkflowDBStatusesMetric(),
+]
+
+
+def _build_monitored_results_path(workflow: str) -> Path:
+    return Path(f"{workflow}_monitored_results.json")
+
+
+def _save_metrics(workflow: str, results: Dict) -> None:
+    with open(_build_monitored_results_path(workflow), "w") as f:
+        json.dump(results, f)
+
+
+def _collect_metrics(parameters: Dict) -> Dict[str, Any]:
+    collected_metrics = {}
+    for metric in METRICS:
+        try:
+            result = metric.collect(parameters)
+            collected_metrics = dict(collected_metrics, **result)
+        except Exception as error:
+            logger.error(
+                f"Error during collection of {metric.name} metric. Details: {error}"
+            )
+    return collected_metrics
+
+
+def _print_metrics() -> None:
+    logger.info("Following metrics will be collected:")
+    for m in METRICS:
+        logger.info(f"- {m.name}")
+
+
+def monitor(workflow: str, sleep: int) -> None:
+    """Start periodically collect defined metrics and save them to JSON file.
+
+    This function is blocking.
+    """
+    _print_metrics()
+    logger.info("Starting monitoring...")
+
+    all_metrics = {}
+    metrics_parameters = {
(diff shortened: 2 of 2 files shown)


reana/reana_benchmark/monitor.py
```diff
@@ -0,0 +1,151 @@
...
+            name = pod.get("metadata", {}).get("name", "")
+            if name_contains in name:
+                selected_pods.append(pod)
+        return selected_pods
+
+    @staticmethod
+    def _count(pods: List[Dict]) -> Dict[str, int]:
+        statistics = defaultdict(lambda: 0)
+        for pod in pods:
+            phase = pod.get("status", {}).get("phase")
```
VMois, line 80: *question*: The `phase` value is not representing exactly what `kubectl get pods` (`STATUS` column) is displaying. It has `Pending`, `Running` and `Succeed` statuses. In addition, it has `Failed` status which, for some reason, appears for batch pods at the end when workflows are *successfully* finished. 

I guess this happens because batch pods enter `NotReady` state because the workflow engine successfully finished and exits. You cannot easily get `NotReady` state from kubectl JSON output, you need to deduct it from container statuses.

What do you think about the situation? Any ideas on how to get a more accurate status?

audrium: The `phase` values are described [here](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase). I think it's a pretty accurate metric for us to use. But this case with successfully finished `Failed` status could be confusing.. I guess one way how we could solve it is to include the `containerStatuses` to the final output (downsize is that it will make it much more verbose..). Here is an example:
```
kubectl get pod -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.phase}{"\t"}{.status.containerStatuses[*].state}{"\n"}{end}'
...
reana-server-69c494847b-fxldq   Running {"running":{"startedAt":"2021-11-23T08:09:29Z"}} {"running":{"startedAt":"2021-11-23T08:09:29Z"}}
...
```
Then it would be visible that the `Failed` status happens because the containers terminated. But not sure if we need it, could be confusing to have too much output..

reana/reana_benchmark/cli.py
```diff
@@ -210,3 +211,16 @@ def collect_command(workflow: str, force: bool) -> NoReturn:
         collect(workflow, force)
     except Exception as e:
         logger.error(f"Something went wrong when collecting results: {e}")
+
+
+@reana_benchmark.command(name="monitor")
+@workflow_option
+@click.option(
+    "--sleep", "-s", help="Sleep between querying in seconds", type=int, default=30,
```
audrium, line 219: Minor: perhaps we could mention that the default sleep time is 30 seconds?

reana/reana_benchmark/monitor.py
VMois: Example of a JSON output about pod `status` for a single pod from `kubectl get pods -o json` command:

```json
"status": {
      "conditions": [
          {
              "lastProbeTime": null,
              "lastTransitionTime": "2021-11-11T16:15:07Z",
              "status": "True",
              "type": "Initialized"
          },
          {
              "lastProbeTime": null,
              "lastTransitionTime": "2021-11-11T16:15:12Z",
              "status": "True",
              "type": "Ready"
          },
          {
              "lastProbeTime": null,
              "lastTransitionTime": "2021-11-11T16:15:12Z",
              "status": "True",
              "type": "ContainersReady"
          },
          {
              "lastProbeTime": null,
              "lastTransitionTime": "2021-11-11T16:15:07Z",
              "status": "True",
              "type": "PodScheduled"
          }
      ],
      "containerStatuses": [
          {
              "containerID": "containerd://78d3916f2dfca74427f8b0cc38cc344a746a2594c8fd6f746133e7a10b0cfdbb",
              "image": "docker.io/reanahub/reana-workflow-controller:latest",
              "imageID": "sha256:75e1ea2066193703f73e1c71fce92969a7b9c47b8af3e307dc2d5ee812e2d40d",
              "lastState": {},
              "name": "job-status-consumer",
              "ready": true,
              "restartCount": 0,
              "started": true,
              "state": {
                  "running": {
                      "startedAt": "2021-11-11T16:15:10Z"
                  }
              }
          },
          {
              "containerID": "containerd://4b915673af6c64cbec168cc8512969507544365d8ba35b3900f836836d384a26",
              "image": "docker.io/reanahub/reana-workflow-controller:latest",
              "imageID": "sha256:75e1ea2066193703f73e1c71fce92969a7b9c47b8af3e307dc2d5ee812e2d40d",
              "lastState": {},
              "name": "rest-api",
              "ready": true,
              "restartCount": 0,
              "started": true,
              "state": {
                  "running": {
                      "startedAt": "2021-11-11T16:15:11Z"
                  }
              }
          }
      ],
      "hostIP": "172.18.0.2",
      "phase": "Running",
      "podIP": "10.244.0.117",
      "podIPs": [
          {
              "ip": "10.244.0.117"
          }
      ],
      "qosClass": "BestEffort",
      "startTime": "2021-11-11T16:15:07Z"
}
```

As described above `phase` can show pods as `Failed` while they are terminating and it is not so straighforward how to extract more precise info from the above JSON. Some checks can be done on `containerStatuses` or, maybe, in `conditions`.

reana/reana_benchmark/monitor.py
```diff
@@ -0,0 +1,156 @@
...
+            statistics[phase] += 1
+        return dict(statistics)
+
+    def _collect(self) -> Any:
+        kubectl_cmd = ("kubectl", "get", "pods", "-o", "json")
```
audrium, line 85: Minor: have you considered filtering out unnecessary info with something like this?
```
kubectl get pod -o=jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.status.phase}{"\t"}{.status.containerStatuses[*].state}{"\n"}{end}'
```
Perhaps it's harder to parse the output like this and is not worth it.. but just an idea

audrium (commented):
Works nicely!

There was one more metric described in the issue which was not implemented - db workflow status. I think it could be also interesting to check it and compare if it's consistent to information which k8s provides.
```
    kubectl exec deployment/reana-db -- psql -U reana -c "SELECT status,COUNT(*) FROM __reana.workflow WHERE name LIKE '$workflow-%' GROUP BY status;"
```

VMois:
Regarding one missing metric, I can add it. I didn't do it initially because, technically, workflow statuses should be listed by `reana-client list` command (which is used in the `collect` command).

reana/reana_benchmark/monitor.py
VMois: I checked the output, it will be harder to parse. I would prefer to load a JSON.

reana/reana_benchmark/monitor.py
VMois: The current goal for monitoring command is to collect data that later can be visualized along with execution progress plot. With additional `containerStatuses` output it is not clear how to visualize it (there is a `phase` but also one or more containers in `containerStatuses`). I would prefer to reduce it to one field (`Running: 5`, `Failed: 1`, etc.) but the question is how :)

VMois: If I understood correctly when the pod is terminated `SIGKILL` is sent by k8s and apparently `job controller` takes too much time to exit so status is `137` and marked as `Failed`.

Example of JSON:

```json
"containerStatuses": [
      {
        "containerID": "containerd://400831aac7443eba9a41228abf264d1adfa63d8279b3a4b71f34f819af7df420",
        "image": "docker.io/reanahub/reana-job-controller:latest",
        "imageID": "sha256:268ad55769e69f1c41d448fb7b54a95dc074cba5e2e1cfa6a2642d8484756551",
        "lastState": {},
        "name": "job-controller",
        "ready": false,
        "restartCount": 0,
        "started": false,
        "state": {
          "terminated": {
            "containerID": "containerd://400831aac7443eba9a41228abf264d1adfa63d8279b3a4b71f34f819af7df420",
            "exitCode": 137,
            "finishedAt": "2021-11-24T10:28:32Z",
            "reason": "Error",
            "startedAt": "2021-11-24T10:27:44Z"
          }
        }
      },
      {
        "containerID": "containerd://d5e4a82bfe1acd86411058ebaa0302dd8a6d9632b1a07d6f849789071918b1b3",
        "image": "docker.io/reanahub/reana-workflow-engine-serial:latest",
        "imageID": "sha256:5756537cf4d2673d2725e0824f6f6883062c363229d6c81a500b19a1487b8641",
        "lastState": {},
        "name": "workflow-engine",
        "ready": false,
        "restartCount": 0,
        "started": false,
        "state": {
          "terminated": {
            "containerID": "containerd://d5e4a82bfe1acd86411058ebaa0302dd8a6d9632b1a07d6f849789071918b1b3",
            "exitCode": 0,
            "finishedAt": "2021-11-24T10:28:02Z",
            "reason": "Completed",
            "startedAt": "2021-11-24T10:27:44Z"
          }
        }
      }
    ],
```

Check timestamps in `containerStatuses`. `workflow-engine` finished at `10:28:02` so around that time REANA asks k8s to delete pod. Take a look at `job-controller` finish time - `10:28:32`. Exactly 30 seconds after `workflow-engine`, 30 seconds is a **default time** in k8s for graceful shutdown of a pod ([details](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace)).

The same happened in other batch pods - `job controller` is not terminating within 30 seconds when k8s asks for it. @tiborsimko

*suggestion:* I think the above case deserves a separate issue. I can leave the `phase` value in the metric for this PR. I assume when we will figure out why the exit code is `137` instead of `0`, the `phase` value will show `Succeeded` status for batch pods.

VMois: Opened the issue to fix `Failed` phase for batch pods. We can proceed forward with this PR and leave `phase` as it is.

reana/reana_benchmark/cli.py
```diff
@@ -210,3 +211,20 @@ def collect_command(workflow: str, force: bool) -> NoReturn:
...
     except Exception as e:
         logger.error(f"Something went wrong when collecting results: {e}")
+
+
+@reana_benchmark.command(name="monitor")
+@workflow_option
+@click.option(
+    "--sleep",
+    "-s",
+    help="Sleep in seconds between collecting metrics, default 30",
```
audrium, line 221: Very minor: I guess to be consistent with `reana-client` and `reana-dev` commands format we could use 
```
Sleep in seconds between collecting metrics [default=30]
```
`help` description here, WDYT?

reana/reana_benchmark/monitor.py
```diff
@@ -0,0 +1,192 @@
...
+            "job_pods": self._count(pods, "run-job"),
+        }
+
+        return result
+
+
+METRICS = [
+    NumberOfDBConnectionsMetric(),
+    WorkflowPodsMetric(),
+    WorkflowStatuses(),
```
audrium, line 135: Minor: for consistency we could add `Metric` to the end. Also, I think it would be a bit more clear to name it `WorkflowDBStatusesMetric` and accordingly `name = "workflow_db_statuses"`. I think it would be more in line with `workflows_pods_status`

audrium (approved):
Works nicely!
I left couple of small comments. PR also needs a rebase

reana/reana_benchmark/cli.py
VMois: Done.

reana/reana_benchmark/monitor.py
VMois: Done

audrium (commented):
Only thing which is missing for this PR is a rebase ("This branch is out-of-date with the base branch")

PR #596, opened 2021-11-22
benchmark: add datetime range plot parameter to analyze command

### LINKED ISSUE #583: benchmark: similar to workflow range, add time range to analyze command
Currently, there is workflow range option `-n` that allows selecting workflows numbers that will be plotted in the execution progress plot. In some cases, it is beneficial to limit the plot by time range to make it more visible.

We already have the`-t` option for `--title` so a new time range option might be named `--time`.

VMois:
Adding `datetime_range` option to `analyze` command to be able to "zoom" into the `execution_progress` plot. The new option is similar to the `interval` option we already have, it acts as a customizer to a plot.

closes #583

How to test:

1. Save dummy CSV file as `datetime-range_collected_results.csv`:

```csv
color-3,1,2021-11-16T10:24:12,2021-11-16T10:24:39,,running,2021-11-16T10:24:23,2021-11-16T10:25:47
color-2,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:25:44,finished,2021-11-16T10:24:23,2021-11-16T10:25:47
color-1,1,2021-11-16T10:24:11,2021-11-16T10:24:35,2021-11-16T10:25:42,failed,2021-11-16T10:24:23,2021-11-16T10:25:47
color-4,1,2021-11-16T10:24:12,,,queued,2021-11-16T10:24:23,2021-11-16T10:25:47
color-5,1,2021-11-16T10:24:12,,,pending,2021-11-16T10:24:23,2021-11-16T10:25:47
color-12,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:25:44,finished,2021-11-16T10:24:23,2021-11-16T10:25:47
color-13,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:25:44,finished,2021-11-16T10:24:23,2021-11-16T10:25:47
```

You can also edit the while (follow the first point from [here](https://github.com/reanahub/reana/pull/592))

2. Produce plot without changes: `reana-benchmark analyze -w datetime-range -n 1-20`. Check execution progress plot. This is an unmodified plot.

3. Use new `datetime` option, e.g run `reana-benchmark analyze -w datetime-range -n 1-20 --datetime 2021-11-16T10:24:35-2021-11-16T10:25:35`. You need to follow the datetime format. Check execution progress plot, now the timeline (x-axis) is only focused on the datetime range you specified.

### DIFF
--- a/reana/reana_benchmark/analyze.py
+++ b/reana/reana_benchmark/analyze.py
@@ -110,9 +110,10 @@ def _build_plots(df: pd.DataFrame, plot_parameters: Dict) -> List[Tuple[str, Fig
 
 def _build_execution_progress_plot(
     df: pd.DataFrame, plot_parameters: Dict
-) -> (str, Figure):
+) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
     interval = plot_parameters["time_interval"]
+    datetime_range = plot_parameters.get("datetime_range")
 
     fig, ax = plt.subplots(figsize=(8, 4), dpi=200, constrained_layout=True)
 
@@ -221,6 +222,11 @@ def _build_execution_progress_plot(
 
     ax.set_ylim(ymin=df["workflow_number"].min() - 1)
 
+    if datetime_range:
+        left = datetime.strptime(datetime_range[0], DATETIME_FORMAT)
+        right = datetime.strptime(datetime_range[1], DATETIME_FORMAT)
+        ax.set_xlim(left=left, right=right)
+
     ax.grid(color="black", linestyle="--", alpha=0.15)
 
     def _build_legend(axes):
@@ -252,7 +258,7 @@ def _build_execution_progress_plot(
 
 def _build_execution_status_plot(
     df: pd.DataFrame, plot_parameters: Dict
-) -> (str, Figure):
+) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
     total = len(df)
     statuses = list(df["status"].unique())
@@ -279,7 +285,7 @@ def _build_execution_status_plot(
     return "execution_status", fig
 
 
-def _max_min_mean_median(series: pd.Series) -> (int, int, int, int):
+def _max_min_mean_median(series: pd.Series) -> Tuple[int, int, int, int]:
     max_value = series.max()
     min_value = series.min()
     mean_value = series.mean()
@@ -336,7 +342,7 @@ def _build_pending_time_histogram(
 
 
 def _save_plots(
-    plots: List[Tuple[str, Figure]], workflow: str, workflow_range: (int, int)
+    plots: List[Tuple[str, Figure]], workflow: str, workflow_range: Tuple[int, int]
 ) -> None:
     logger.info("Saving plots...")
     for base_name, figure in plots:
@@ -347,7 +353,7 @@ def _save_plots(
 
 
 def analyze(
-    workflow_prefix: str, workflow_range: (int, int), plot_params: Dict
+    workflow_prefix: str, workflow_range: Tuple[int, int], plot_params: Dict
 ) -> None:  # noqa: D103
     results_path = build_collected_results_path(workflow_prefix)
     collected_results = pd.read_csv(results_path)
--- a/reana/reana_benchmark/cli.py
+++ b/reana/reana_benchmark/cli.py
@@ -8,18 +8,22 @@
 """Combines reana-benchmark modules into CLI commands."""
 
 import logging
-from typing import NoReturn
+from typing import NoReturn, Tuple, Optional, Union
 
 import click
 import urllib3
 
 from reana.reana_benchmark.analyze import analyze
 from reana.reana_benchmark.collect import collect
-from reana.reana_benchmark.config import WORKERS_DEFAULT_COUNT
+from reana.reana_benchmark.config import (
+    WORKERS_DEFAULT_COUNT,
+    DATETIME_RANGE_SEPARATOR_POSITION,
+    DATETIME_RANGE_LENGTH,
+)
 from reana.reana_benchmark.start import start
 from reana.reana_benchmark.submit import submit
-from reana.reana_benchmark.utils import logger
 from reana.reana_benchmark.monitor import monitor
+from reana.reana_benchmark.utils import logger, validate_date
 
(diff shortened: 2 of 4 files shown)


reana/reana_benchmark/cli.py
```diff
@@ -165,6 +201,14 @@ def launch(
 @reana_benchmark.command(name="analyze")
 @workflow_option
 @workflow_range_option
+@click.option(
+    "--datetime",
```
VMois, line 205: *thought:* I was thinking maybe splitting the command into `--datetime-start`, `--datetime-end` (or similar like mentioned [here](https://github.com/reanahub/reana/issues/583#issuecomment-975318100)). Having a single long string feels a bit intimidating to me. Let's see what review will yield.

In the case of splitting, we might need to consider a scenario where only one limit is given (or enforce both).

reana/reana_benchmark/analyze.py
```diff
@@ -204,6 +205,11 @@ def _build_execution_progress_plot(
 
     ax.set_ylim(ymin=df["workflow_number"].min() - 1)
 
+    if datetime_range:
+        left = datetime.strptime(datetime_range[0], DATETIME_FORMAT)
+        right = datetime.strptime(datetime_range[1], DATETIME_FORMAT)
+        ax.set_xlim(xmin=left, xmax=right)
```
audrium, line 211: Question: just curious why you chose `xmin` and `xmax` over `left` and `right` parameters? According [docs](https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.set_xlim.html) it's the same. Have you spotted any difference?

reana/reana_benchmark/analyze.py
```diff
@@ -321,7 +327,7 @@ def _save_plots(
 
 
 def analyze(
-    workflow_prefix: str, workflow_range: (int, int), plot_params: Dict
+    workflow_prefix: str, workflow_range: Tuple[int, int], plot_params: Dict
```
audrium, line 330: Nicely solves the annotation error :+1: We should do the same to other annotations in this file, currently there are 7 errors. Could be done later..

reana/reana_benchmark/cli.py
```diff
@@ -95,6 +95,42 @@ def _to_range(workflow_range: str) -> (int, int):
...
+
+def _to_datetime_range(
+    datetime_range: Optional[str],
+) -> Union[Tuple[str, str], None, NoReturn]:
+    """Take datetime range string, split it into left and right datetime limits, validate and return."""
+    if not datetime_range:
+        return None
+
+    # middle position between two datetimes, position value is always the same
+    separator_position = 19
```
audrium, line 108: Minor: perhaps we could declare a constant for this in the config?
Same for `length_of_two_datetimes_plus_separator`. I think we could name it `DATETIME_LENGTH` and `DATETIME_RANGE_LENGTH`? WDYT?

reana/reana_benchmark/cli.py
audrium: I personally like that it's defined in one command, it's pretty easy to simply copy the value from the example you provided and edit it as one needs. I think we could also add `-d` for `--datetime`, WDYT?

audrium (approved):
Works nicely, haven't found any issues :+1:

reana/reana_benchmark/cli.py
```diff
@@ -165,6 +201,14 @@ def launch(
 @reana_benchmark.command(name="analyze")
 @workflow_option
 @workflow_range_option
+@click.option(
+    "--datetime",
+    "datetime_range",
+    help="Datetime range for the execution progress plot, "
```
audrium, line 207: Minor: I think we could specify a bit more what this option does, something like `Filter execution progress plot data by a specific datetime range `

reana/reana_benchmark/analyze.py
VMois: It just happened :) I found an example and it was with `xmin/xmax`. I didn't know about the other option. Nice insight! But, `left/right` looks better (+ I even named unconsciously variables).

reana/reana_benchmark/analyze.py
VMois: Initially, I wanted to amend everything but realized that I will touch code that doesn't belong to the PR/issue. If we are not strict about it, I can update the whole file. It will take a few minutes.

reana/reana_benchmark/cli.py
VMois: Not sure it needs to go to `config`. It is only used in this function. Probably I should uppercase constants in `_to_datetime_range` to make specific that they are constants.

reana/reana_benchmark/cli.py
VMois: I can add `-d`. It will also follow other options.

reana/reana_benchmark/cli.py
VMois: Agree that a better description is needed. But instead of `filter ... data`, I would just put `Filter execution progress plot to contain data within the specified datetime range`.

reana/reana_benchmark/cli.py
VMois: Regarding splitting the command, let's leave it as it is. In case it will bring pain we can always split the command.

reana/reana_benchmark/analyze.py
VMois: I updated a couple of typings. In the next PRs, there are more fixes so will leave it as it is here.

reana/reana_benchmark/analyze.py
VMois: Done

reana/reana_benchmark/analyze.py
VMois: Done

reana/reana_benchmark/cli.py
VMois: It looks like you are not allowed to define constant (uppercase) variables inside the function so I moved it to `config.py`

reana/reana_benchmark/cli.py
VMois: Done

PR #597, opened 2021-11-24
helm: add alembic migration step to installation notes

mvidalgarcia:


helm/reana/templates/NOTES.txt
```diff
@@ -37,4 +37,11 @@ finalise its configuration.
     $ cd reana-demo-root6-roofit
     $ reana-client run -w demo
 
+If you already had REANA deployed and you are upgrading it, there is one step left to
```
tiborsimko, line 40: What about making it shorter as follows: 

```
If you are upgrading from a previous REANA release series, you may need to upgrade database schema now:

   $ kubectl -n {{ .Release.Namespace }} exec -i -t deployment/{{ .Release.Name }}-server -- reana-db alembic upgrade

Please see the dedicated `[upgrade](url)` documentation page with more information.

If you are deploying...
...
...
```

mvidalgarcia: I guess markdown links won't be very effective here. Should we leave it like `[upgrade](url)` regardless?

tiborsimko: Indeed, perhaps just use ASCII-friendly:

```
... please see the dedicated upgrade documentation at <http://example.org/foo/bar>.
```

tiborsimko: Also, please wrap the line so that it'll be <80 characters long. Easier to read on narrow terminals. (The command can remain longer, as the others...)

helm/reana/templates/NOTES.txt
```diff
@@ -1,5 +1,13 @@
 The REANA system has been installed!
 
+If you are upgrading from a previous REANA release series, you may need to
+upgrade database schema now:
+
+   $ kubectl -n {{ .Release.Namespace }} exec -i -t deployment/{{ .Release.Name }}-server -- reana-db alembic upgrade
+
+Please see the dedicated upgrade documentation at
+<https://docs.reana.io/administration/management/migrating-db/>
```
tiborsimko, line 9: One more thing: Does DB migration belongs to "Managing..."? To me it is rather part of "Deploying...", since DB upgrade is a general topic thing, valid all the time for all the deployments, whilst "Managing" has specific topics listed there once the site is up and running, managing users, managing quotas, managing nodes, etc.  

I'd also probably call it "DB upgrade", not "DB migration", which might evoke a feeling of moving from PostgreSQL version N to M, or to another database solution.

Perhaps we can review that one first, settle on a URL there, and come back to this one afterwards.

mvidalgarcia: I called it "Migrating" as this is the term that [alembic](https://alembic.sqlalchemy.org/en/latest/) uses officially.

> Alembic is a lightweight database migration tool for usage with the SQLAlchemy Database Toolkit for Python.

Moved to deployment.

tiborsimko: I see the motivation :+1: However, since people would usually speak of "upgrading REANA", "upgrading client", "upgrading Python" etc from one version to a new version, it seems more end-user-friendly and more consistent to use the term "upgrading DB" as well. (Besides, the Alembic command itself is called `upgrade`...)

tiborsimko (approved):

PR #599, opened 2021-11-26
benchmark: approximate x-axis interval for execution progress plot

### LINKED ISSUE #590: benchmark: single x-axis tick
Sometimes, when workflows run fast, the x-axis has only one label, which is not enough to judge about the runtimes :wink: 

roofit-1i-32b-134j-submit200-sleep0-total200-run20211116k_execution_progress_1_200

roofit-1i-32b-134j-submit200-sleep0-total200-run20211116k_histogram_runtime_1_200

VMois:
- in addition, add one more tick on x-axis if interval is approximated to ensure that all points are covered;
- change time interval from minutes to seconds.

closes #590

How to test:

1. If you don't have already collected results from previous benchmarks you can get new ones:

```console
$ reana-benchmark launch -w time -n 1-10
$ reana-benchmark collect -w time
```

OR, you can use this CSV content and save it under `time_collected_results.csv`:

```csv
time-3,1,2021-11-16T10:24:12,2021-11-16T10:24:39,,running,2021-11-16T10:24:23,2021-11-16T10:24:47
time-2,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:24:44,finished,2021-11-16T10:24:23,2021-11-16T10:24:47
time-1,1,2021-11-16T10:24:11,2021-11-16T10:24:35,2021-11-16T10:24:42,failed,2021-11-16T10:24:23,2021-11-16T10:24:47
time-4,1,2021-11-16T10:24:12,,,queued,2021-11-16T10:24:23,2021-11-16T10:24:47
```

2. Run `reana-benchmark analyze -w time -n 1-10` to get plots.

3. This PR modifies x-axis ticks interval in the `execution_progress` plot. You can take a look if ticks make sense with various date lengths.

4. Add `-i` (time interval option) to check if new updates didn't break the option. Run `reana-benchmark analyze -w time -n 1-10 -i 5`

### DIFF
--- a/reana/reana_benchmark/analyze.py
+++ b/reana/reana_benchmark/analyze.py
@@ -112,7 +112,6 @@ def _build_execution_progress_plot(
     df: pd.DataFrame, plot_parameters: Dict
 ) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
-    interval = plot_parameters["time_interval"]
     datetime_range = plot_parameters.get("datetime_range")
 
     fig, ax = plt.subplots(figsize=(8, 4), dpi=200, constrained_layout=True)
@@ -211,7 +210,10 @@ def _build_execution_progress_plot(
     # force integers on y axis
     ax.yaxis.set_major_locator(MaxNLocator(integer=True))
 
-    ax.xaxis.set_major_locator(mdates.MinuteLocator(interval=interval))
+    interval = plot_parameters.get("time_interval")
+    if interval:
+        ax.xaxis.set_major_locator(mdates.SecondLocator(interval=interval))
+
     ax.xaxis.set_major_formatter(mdates.DateFormatter("%H:%M:%S"))
 
     # rotate dates on x axis
--- a/reana/reana_benchmark/cli.py
+++ b/reana/reana_benchmark/cli.py
@@ -8,7 +8,7 @@
 """Combines reana-benchmark modules into CLI commands."""
 
 import logging
-from typing import NoReturn, Tuple, Optional, Union
+from typing import NoReturn, Optional, Tuple, Union
 
 import click
 import urllib3
@@ -221,16 +221,15 @@ def launch(
 @click.option(
     "--interval",
     "-i",
-    help="Ticks interval in minutes for execution progress plot [default=10]",
+    help="Ticks interval in seconds for execution progress plot",
     type=int,
-    default=10,
 )
 def analyze_command(
     workflow: str,
     workflow_range: Tuple[int, int],
     datetime_range: Tuple[str, str],
     title: str,
-    interval: int,
+    interval: Optional[int],
 ) -> NoReturn:
     """Produce plots based on workflows results collected before."""
     plot_params = {


reana/reana_benchmark/analyze.py
```diff
@@ -210,7 +242,17 @@ def _build_execution_progress_plot(
...
+    # set x axis interval and ticks
+    interval = plot_parameters.get("time_interval")
+    if not interval:
+        interval = _approximate_locator_interval_seconds(df)
+    ax.xaxis.set_major_locator(mdates.SecondLocator(interval=interval))
+    if not plot_parameters.get("time_interval"):
+        locs, _ = plt.xticks()
+        additional_tick = (locs[1] - locs[0]) + locs[-1]
+        locs = np.append(locs, additional_tick)
+        ax.set_xticks(locs)
```
VMois, line 254: *question:* What do you think about adding an additional tick? Sometimes even if the interval is approximated, there won't be enough ticks to cover all points/lines.

audrium: Currently a tick to mark the beginning is missing. Before we could see when workflows have started
Before
hw_execution_progress_1_10 copy 2
Now:
hw_execution_progress_1_10

audrium: Actually, do we even need to approximate the interval by ourselves at all? If this peace of code is commented out (lines 245-257), `matplotlib` will do it for us:
hw_execution_progress_1_10

VMois: Check the discussions in the issue. I would prefer to keep it manual and specify `-i` but this is what was decided. 

*suggestion:*: following on your observation, we can just get rid of most of the PR code and make `-i` just optional instead of the default value. Then matplotlib will do all stuff for us. Sounds nice. Let me test it.

reana/reana_benchmark/analyze.py
```diff
@@ -108,11 +109,42 @@ def _build_plots(df: pd.DataFrame, plot_parameters: Dict) -> List[Tuple[str, Fig
...
+        "ended",
+        "collected_date",
+    ]:
+        sorted_series = df[date_column].sort_values().dropna()
+
+        if len(sorted_series) < 1:
+            continue
+
+        oldest_local_epoch = _convert_str_date_to_epoch(sorted_series.iloc[0])
+        latest_local_epoch = _convert_str_date_to_epoch(sorted_series.iloc[-1])
```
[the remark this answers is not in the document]
VMois: Let's see if another idea below will work out. In this case, we will get rid of this pile :)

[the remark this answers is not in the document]
VMois: But `left_epoch`/`right_epoch` can be a better option.

reana/reana_benchmark/analyze.py
audrium: yeah, that's what I meant, perhaps it would be enough to do:
```python
interval = plot_parameters.get("time_interval")
if interval:
    ax.xaxis.set_major_locator(mdates.MinuteLocator(interval=interval))
  ```

VMois: I tested the above simpler version and it still has some flaws (some ticks are missing at the beginning or end) but not much difference compared to a big chunk of code I wrote. I think it should solve the main issue. Thank you for the suggestion. I got too focused on one solution and didn't realize to check the simpler one. Good to have a second pair of eyes.

Example of "problematic" one with the above simple solution:

histogram_execution_progress_1_5

VMois: I put a second commit with a simpler version to keep the older variant while PR is in review. If we will go with a simpler one I will squash commits at the end.

audrium (approved):
I think this simpler solution solves the problem nicely (even though there are some flows as you spotted). Here are the comparison:
Before:
test_execution_progress_1_10 copy
After:
test_execution_progress_1_10 copy 2

Can you rebase and squash the commits?

PR #600, opened 2021-11-29
benchmark: use only existing datetimes for histograms

### LINKED ISSUE #598: benchmark: include only finished workflows into histogram calculation
`pending`/`queued`/`running` workflows should not affect histogram numbers. We should calculate runtime/pending/total time histogram based on finished workflows only. Maybe include `failed` or put them into a separate histogram.

More context https://github.com/reanahub/reana/issues/583#issuecomment-975470460

VMois:
- do not substitute missing datetimes with collected datetime, use only the existing one to provide the most accurate results;
- use only finished workflows for total_time and runtime histograms.

closes #598

How to test:

1. You can use this CSV content and save it under `histogram_collected_results.csv`:

```csv
histogram-3,1,2021-11-16T10:24:12,2021-11-16T10:24:39,,running,2021-11-16T10:24:23,2021-11-16T10:24:47
histogram-2,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:24:44,finished,2021-11-16T10:24:23,2021-11-16T10:24:47
histogram-5,1,2021-11-16T10:24:11,2021-11-16T10:24:37,2021-11-16T10:24:44,finished,2021-11-16T10:24:23,2021-11-16T10:24:47
histogram-1,1,2021-11-16T10:24:11,2021-11-16T10:24:35,2021-11-16T10:24:42,failed,2021-11-16T10:24:23,2021-11-16T10:24:47
histogram-4,1,2021-11-16T10:24:12,,,queued,2021-11-16T10:24:23,2021-11-16T10:24:47
```

2. Run `reana-benchmark analyze -w histogram -n 1-10` to get plots.

3. Check histograms. Summary (if you didn't modify CSV): `total_time` (contains only finished workflows) = 2 runs, `pending_time` (contains finished + failed + running workflows) = 4 runs, and `runtime` (contains only finished workflows) = 2 runs.

### DIFF
--- a/reana/reana_benchmark/analyze.py
+++ b/reana/reana_benchmark/analyze.py
@@ -10,7 +10,7 @@ import time
 from datetime import datetime
 from functools import partial
 from pathlib import Path
-from typing import List, Dict, Tuple
+from typing import List, Dict, Tuple, Optional
 
 import matplotlib.dates as mdates
 import matplotlib.pyplot as plt
@@ -19,9 +19,8 @@ from matplotlib.figure import Figure
 from matplotlib.ticker import MaxNLocator
 
 from reana.reana_benchmark.collect import build_collected_results_path
-from reana.reana_benchmark.utils import logger
 from reana.reana_benchmark.config import DATETIME_FORMAT, WorkflowStatus
-
+from reana.reana_benchmark.utils import logger
 
 STATUS_TO_COLOR = {
     WorkflowStatus.created: "grey",
@@ -48,11 +47,9 @@ def _derive_metrics(df: pd.DataFrame) -> pd.DataFrame:
         lambda row: _get_workflow_number_from_name(row["name"]), axis=1
     )
 
-    collected_date = df["collected_date"].iloc[0]
-
     def _calculate_difference(
         row: pd.Series, start_column: str, end_column: str
-    ) -> int:
+    ) -> Optional[int]:
         """Calculate difference between two date times in string format."""
         start_date = row[start_column]
         end_date = row[end_column]
@@ -64,14 +61,7 @@ def _derive_metrics(df: pd.DataFrame) -> pd.DataFrame:
             return _convert_str_date_to_epoch(end_date) - _convert_str_date_to_epoch(
                 start_date
             )
-
-        # if only start date exists, take current time as ended
-        if start_date_exists and not end_date_exists:
-            return _convert_str_date_to_epoch(
-                collected_date
-            ) - _convert_str_date_to_epoch(start_date)
-
-        return 0
+        return None
 
     df["pending_time"] = df.apply(
         partial(
@@ -81,13 +71,11 @@ def _derive_metrics(df: pd.DataFrame) -> pd.DataFrame:
         ),
         axis=1,
     )
-    df["pending_time"] = df["pending_time"].astype(int)
 
     df["runtime"] = df.apply(
         partial(_calculate_difference, start_column="started", end_column="ended"),
         axis=1,
     )
-    df["runtime"] = df["runtime"].astype(int)
     return df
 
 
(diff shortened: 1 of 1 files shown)


reana/reana_benchmark/analyze.py
```diff
@@ -307,31 +295,35 @@ def _build_histogram_plot(
 
 def _build_total_time_histogram(
     df: pd.DataFrame, plot_parameters: Dict
-) -> (str, Figure):
+) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
+    data = df[df["status"] == WorkflowStatus.finished]
+    series = data["runtime"] + data["pending_time"]
```
audrium, line 301: Shouldn't we use `.dropna()` here as well?

reana/reana_benchmark/analyze.py
```diff
@@ -307,31 +295,35 @@ def _build_histogram_plot(
...
+        _build_histogram_plot(data, 10, "runtime", title),
     )
 
 
 def _build_pending_time_histogram(
     df: pd.DataFrame, plot_parameters: Dict
-) -> (str, Figure):
+) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
+    data = df["pending_time"].dropna()
```
audrium, line 323: Shouldn't we filter `finished` workflows here as well?

reana/reana_benchmark/analyze.py
```diff
@@ -307,31 +295,35 @@ def _build_histogram_plot(
...
+        _build_histogram_plot(series, 10, "total_time", title),
     )
 
 
-def _build_runtime_histogram(df: pd.DataFrame, plot_parameters: Dict) -> (str, Figure):
+def _build_runtime_histogram(
+    df: pd.DataFrame, plot_parameters: Dict
+) -> Tuple[str, Figure]:
     title = plot_parameters["title"]
+    data = df[df["status"] == WorkflowStatus.finished]["runtime"].dropna()
```
audrium, line 312: Very minor: I like how you filtered `data` and defined `series` in separate lines in `_build_total_time_histogram()` function. I guess same approach could be used here as well (and also `_build_pending_time_histogram()`, WDYT?
Since `_build_histogram_plot()` expects `series` parameter, it would be nicer to pass it instead of `data`

reana/reana_benchmark/analyze.py
VMois: `finished` status always has `runtime` and `pending_time` but it can be confusing so I can add `dropna()`

reana/reana_benchmark/analyze.py
VMois: Agree. Nice catch.

reana/reana_benchmark/analyze.py
VMois: Depends, if you want to have only `finished/failed` or `running` in addition. When a workflow is in `running` status it already has defined `pending_time`. That is why I use `dropna()` here. But it may be confusing. Maybe, I should add some comments or even better filter it in the same way as it is done with `finished` so it will not create confusion later.

reana/reana_benchmark/analyze.py
VMois: Done

reana/reana_benchmark/analyze.py
VMois: I decided to leave it as it is because `dropna()` can be confusing. Someone can assume that the "runtime" and "pending_time" columns could have different data (because `dropna` removed it). But if you have suggestions how to make it more clear maybe, I can do it.

reana/reana_benchmark/analyze.py
VMois: I added filter so it more clear what statuses are include into `pending_time`

reana/reana_benchmark/analyze.py
audrium: Yeah, I agree that it's better not to add it, since as you said `finished` workflows will always have `runtime` and `pending_time`, so there is no need for `dropna()`. But then we need to remove `dopna()` in the `_build_runtime_histogram()` (line 313)?

reana/reana_benchmark/analyze.py
audrium: I find it a bit confusing that other two histogram plots filters `finished` workflows and this one uses more statuses. Perhaps we should include this info in the title or in the legend?
So the only reason `dropna()` is useful is to filter out `failed` workflows which have no `pending_time`? That means it failed before starting? Or I am wrong?

reana/reana_benchmark/analyze.py
audrium: Actually, `dropna()` does not seem very necessary, as `NaN` values are simply omitted from the plot. Or am I missing something?

audrium (commented):
Reminder that rebase is needed

reana/reana_benchmark/analyze.py
VMois: Haven't checked it, maybe. I would prefer to drop them just in case.

reana/reana_benchmark/analyze.py
VMois: `dropna` filters `queued` and `pending`. `failed`, `running` and `finished` all have valid `pending_time` (meaning REANA provided exact start date for workflow) 

I can include it in the title. Not sure if this is important but always good to include.

VMois: Done

audrium (approved):
Thanks, works nicely!

PR #603, opened 2021-12-02
helm: make http logging configurable

mvidalgarcia:
- by default, log all HTTP requests
- allow logging only HTTP error requests (status code 4xx and 5xx)

closes reanahub/reana-server#417

#### To test:
- Force uWSGI (Production mode):
```diff
diff --git a/helm/reana/templates/reana-server.yaml b/helm/reana/templates/reana-server.yaml
index ab43b8b..6be9538 100644
--- a/helm/reana/templates/reana-server.yaml
+++ b/helm/reana/templates/reana-server.yaml
@@ -37,7 +37,7 @@ spec:
         ports:
         - containerPort: 5000
           name: http
-        {{- if .Values.debug.enabled }}
+        {{- if false }}
         command: ["/bin/sh", "-c"]
         args: ["invenio run -h 0.0.0.0 -p 5000"]
         tty: true
```
- Play with different configurations and verify that logs are properly displayed
- To force 4xx, just navigate to /api/foo (`curl -k https://localhost:30443/api/foo`)
- To force 5xx, you can raise an Exception in one of the API endpoints.

### DIFF
--- a/helm/reana/templates/uwsgi-config.yaml
+++ b/helm/reana/templates/uwsgi-config.yaml
@@ -25,9 +25,11 @@ data:
                           # https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html
 
     # Logging
+    {{- if not .Values.components.reana_server.uwsgi.log_all }}
     disable-logging = true
-    log-4xx = true
-    log-5xx = true
+    log-4xx = {{ .Values.components.reana_server.uwsgi.log_4xx }}
+    log-5xx = {{ .Values.components.reana_server.uwsgi.log_5xx }}
+    {{- end }}
 
     # Workers management
     max-requests = 1999
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,14 @@
 Changes
 =======
 
+Version 0.8.1 (UNRELEASED)
+--------------------------
+
+- Administrators:
+    - Adds new configuration option ``reana_server.uwsgi.log_all`` to toggle the logging of all the HTTP requests.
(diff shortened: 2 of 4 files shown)


helm/reana/templates/uwsgi-config.yaml
```diff
@@ -25,9 +25,11 @@ data:
                           # https://uwsgi-docs.readthedocs.io/en/latest/articles/SerializingAccept.html
 
     # Logging
+    {{- if .Values.components.reana_server.uwsgi.log_http_errors_only }}
```
tiborsimko, line 28: One disadvantage of having only one true/false helm value representing both 4xx and 5xx is that some admins wouldn't be easily able to choose only 4xx or only 5xx if they wish so. Moreover, considering that there are more `log-*` options that admins might be interested to log, such as `log-ioerror`, perhaps we could make it a list?  

However I'm not sure it is worth the trouble, because somebody else might wish to `log-slow` which would need an argument... It would not be advantageous to do any complex argument handling here in Helm for those cases. (But rather to offer a way to pass the full textual snippet.) 

Just musing out loud whether one helm value is enough to represent "logging errors only", or whether we introduce more helm values to have more granularity (`reana_server.uwsgi.log_all`, `reana_server.uwsgi.log_4xx`, `reana_server.uwsgi.log_5xxx`).

mvidalgarcia: I had this very same doubt and I decided to go for a simpler approach, but indeed it's not very flexible... Maybe we can just replicate the very same uwsgi values?

However.. I'm just thinking that `uwsgi-config` is a config map so admins should be able to upgrade it? Although, I'm afraid that changes will be lost after an upgrade..

mvidalgarcia: Added the following configuration options and default values to make it more granular:
- `log_all: true`
- `log_4xx: true`
- `log_5xx: true`

audrium (approved):
Works nicely as expected and solves the problem!

Issue #605, opened 2021-12-06
cronjob: calculate CPU quotas


mvidalgarcia:
**Current behavior**

We have a [cronjob](https://github.com/reanahub/reana/blob/36d38c2dec9b376e332665bcea929a9a7a709645/helm/reana/templates/cronjobs.yaml#L111) called `reana-disk-quota-update` that calculates (nightly or following this [schedule](https://github.com/reanahub/reana/blob/36d38c2dec9b376e332665bcea929a9a7a709645/helm/reana/values.yaml#L132)) the disk quota of every user by calling the [`disk_usage_update`](https://github.com/reanahub/reana-db/blob/2f5f1c0f20adde98f1f0854f48292f039ab68473/reana_db/cli.py#L227) function via [`reana-db` CLI](https://github.com/reanahub/reana/blob/36d38c2dec9b376e332665bcea929a9a7a709645/helm/reana/templates/cronjobs.yaml#L130).

The quota update policy is configurable via values.yaml:
- [`termination_update_policy`](https://github.com/reanahub/reana/blob/36d38c2dec9b376e332665bcea929a9a7a709645/helm/reana/values.yaml#L133): Allows calculating Disk and CPU when a workflow is finished. E.g. "disk,cpu"
- [`disk_update`](https://github.com/reanahub/reana/blob/36d38c2dec9b376e332665bcea929a9a7a709645/helm/reana/values.yaml#L132): crontab to update the Disk quota periodically (nightly by default)

**Expected behavior**

`disk_update` should be renamed to be more generic (e.g. `periodic_update_policy`) as well as the cronjob name. This cronjob should also calculate and save the CPU usage besides the Disk usage. To do that, it should query all the workflows and add up all the running times.

PR #607, opened 2021-12-13
helm: generalize disk quota cronjob

mvidalgarcia:
* cover also CPU updates
* replace `disk_update` with `periodic_update_policy`

closes https://github.com/reanahub/reana/issues/605

#### To test:

To check multiple scenarios, edit `values-dev.yaml` to use old or new values (backward compatibility). E.g. 
- Only calculate CPU quota on workflow termination
- Run periodic quota updater every two minutes

```diff
diff --git a/helm/configurations/values-dev.yaml b/helm/configurations/values-dev.yaml
index 01aa44d..c0dcc11 100644
--- a/helm/configurations/values-dev.yaml
+++ b/helm/configurations/values-dev.yaml
@@ -23,3 +23,9 @@ components:
       image: reanahub/reana-message-broker
     reana_ui:
       image: reanahub/reana-ui
+
+quota:
+  disk_update: "*/2 * * * *"
+  termination_update_policy: "CPU"
+  # check new values
+  # periodic_update_policy: ""
+  # workflow_termination_update_policy: "cpu"
```

- Check that quotas are being calculated when configure by checking the CLI `reana-client quota-show` or the UI in the Profile page.


helm/reana/README.md
```diff
@@ -103,8 +103,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `volume_paths.shared_volume_path`                        | Path inside the REANA components where the shared volume will be mounted             | /var/reana                                      |
 | `quota.enabled`                                          | Enable user workflow accounting capabilities.                                        | true                                            |
-| `quota.disk_update`                                      | Cronjob pattern representing how often the users disk quota usage should be updated. Leave it empty to deactivate it | "0 3 * * *"     |
-| `quota.termination_update_policy`                        | Resources to calculate quotas on worflow termination. Possible values: "cpu" and "disk". Leave it empty to deactivate workflow termination accounting. | "cpu,disk"                                   |
+| `quota.cronjob_update_policy`                            | Cronjob pattern representing how often the users Disk and CPU quota usage should be updated. Leave it empty to deactivate it | "0 3 * * *" |
```
tiborsimko, line 106: Whilst a name change would be good to do since the nightly updater can do both CPU and Disk, we should beware of changing the names between 0.8.0 and 0.8.1... If we keep this for `maint-0.8`, then we have to keep the names.  If we change the names, then this must go into 0.9 release series only.

(BTW, speaking of changing names, `quota.termination_update_policy` might read a bit ambiguous to some perhaps, as if we were terminating quotas... Using a fuller name such as `quota.workflow_termination_update_policy` would be more understandable, since more explicit. But these kinds of musings would belong only for 0.9 if we want to change both names.)

(Ditto `quota.periodic_update_policy` might read better than `quota.cronjob_update_policy` for similar ambiguity reasons, in order to avoid an impression that we are updating cronjobs, as it were.)

mvidalgarcia: Please check/test thoroughly the template (`tpl`) mechanism to be able to be backward compatible among values in `values.yaml`.

VMois (approved):
Works well. Most of the comments in the [other PR](https://github.com/reanahub/reana-db/pull/156).

I changed `periodic_update_policy: null` to `periodic_update_policy: ""` as `null` was throwing a Helm error. I assume it was a typo.

PR #608, opened 2021-12-16
cli: add target commit for release to `release-helm`

### LINKED ISSUE #497: cli: fix release-helm tag logic
Currently, almost all `maint-0.7` releases are incorrectly tagged. This happens when following the [Releasing docs](https://github.com/reanahub/reana/wiki/Releasing) tagging a maintenance branch. Tags end up pointing to the latest `master` instead of the correct latest commit in `maint-x.y`.

All Helm releases are fortunately correct so no harm has been done on that side (the `.tar.gz` contain the correct versions).

The actions to be taken are:
- [x] Fix all past release tags to point to the correct release commits. Note this involves force pushing, so we need to warn people.
- [x] Fix the `release-helm` command itself to not behave like this.

mvidalgarcia:
closes #497

#### To test:
It's a bit challenging to test this PR without performing an actual chart release from a `maint` branch. I've added a `--dry-run` flag to `release-helm` so one could first verify the commands to be run.

### DIFF
--- a/reana/reana_dev/release.py
+++ b/reana/reana_dev/release.py
@@ -18,6 +18,7 @@ import click
 
 from reana.reana_dev.docker import docker_push
 from reana.reana_dev.git import (
+    get_current_commit,
     git_clean,
     git_is_current_version_tagged,
     is_last_commit_release_commit,
@@ -189,9 +190,15 @@ def release_pypi(ctx, component, timeout):  # noqa: D301
 
 
 @click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    default=False,
+    help="Display the command without executing it.",
+)
 @release_commands.command(name="release-helm")
 @click.pass_context
-def release_helm(ctx, user):  # noqa: D301
+def release_helm(ctx, user: str, dry_run: bool) -> None:  # noqa: D301
     """Release REANA as a Helm chart."""
     component = "reana"
     version = get_current_component_version_from_source_files(component)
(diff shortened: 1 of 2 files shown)


reana/reana_dev/release.py
```diff
@@ -189,9 +190,12 @@ def release_pypi(ctx, component, timeout):  # noqa: D301
 
 
 @click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option(
+    "--dry-run", is_flag=True, default=False, help="Do not execute any actions."
```
audrium, line 194: Minor: perhaps `Display the command without executing it.` would be more informative?

audrium (approved):
Works nicely! (Tested only the `--dry-run` option to make sure ``--commit` option is added to the `cr upload` command)

reana/reana_dev/utils.py
```diff
@@ -310,16 +317,22 @@ def run_command(cmd, component="", display=True, return_output=False, directory=
     :param component: standard component name
     :param display: should we display command to run?
     :param return_output: shall the output of the command be returned?
+    :param directory: directory whre to run the command
```
audrium, line 320: Super minor: `whre` -> `where`

reana/reana_dev/release.py
mvidalgarcia: I agree, it's more clear. I copied it from https://gitlab.cern.ch/reanahub/deployment/-/blob/master/deployment.py#L53 so we might want to change it there also.

reana/reana_dev/utils.py
mvidalgarcia: Whoops 😅

mvidalgarcia:
> (Tested only the `--dry-run` option to make sure `--commit` option is added to the `cr upload` command)

Yes, it's a bit hard to test... We can revisit it once we deploy from `maint`.

PR #610, opened 2022-01-17
helm: add kubernetes jobs timeout parameters

VMois:
- kubernetes_jobs_timeout_limit, default job timeout for users;
- kubernetes_jobs_max_user_timeout_limit, maximum allowed job timeout for users to set.
- convert parameters to string using "!!str" to make it compatible in the future in case we will decide to use suffixes like "m" to express "minutes".

closes reanahub/reana-job-controller#343


helm/reana/README.md
```diff
@@ -61,6 +61,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `kubernetes_jobs_memory_limit`                           | Maximum default memory limit for user job containers. Exceeding this limit will terminate the container. Please see the following URL for possible values https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory. | 4Gi |
 | `kubernetes_jobs_max_user_memory_limit`                  | Maximum custom memory limit that users can assign to their job containers via `kubernetes_memory_limit` hint in `reana.yaml`. Exceeding this limit will terminate the container. Please see the following URL for possible values https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory. | None |
+| `kubernetes_jobs_timeout_limit`                          | Default timeout for user's job in *seconds*. Exceeding this time will terminate the job. Please see the following URL for more details https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup. | 604800 |
```
mvidalgarcia, line 64:
```suggestion
| `kubernetes_jobs_timeout_limit`                          | Default timeout for user's jobs in *seconds*. Exceeding this time will terminate the job. Please see the following URL for more details https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup. | 604800 |
```

User's job**s** might be better? Otherwise, it seems that a user can only have a single job, but it might be just my impression, what do you think?

mvidalgarcia: Would it make sense to add "(7 days)" next to the default value? Having just a big number is not very meaningful at first sight. As an admin, I'd appreciate knowing how much this is quick.

helm/reana/README.md
```diff
@@ -61,6 +61,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
+| `kubernetes_jobs_max_user_timeout_limit`                 | Maximum custom timeout in *seconds* that users can assign to their job. Please see the following URL for more details https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup. | 1209600 |
```
mvidalgarcia, line 65:
```suggestion
| `kubernetes_jobs_max_user_timeout_limit`                 | Maximum custom timeout in *seconds* that users can assign to their jobs. Please see the following URL for more details https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup. | 1209600 |
```

Same comments from above applies here.

helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -113,6 +113,10 @@ spec:
           {{- end }}
           - name: REANA_KUBERNETES_JOBS_MEMORY_LIMIT
             value: {{ .Values.kubernetes_jobs_memory_limit | default "4Gi" }}
+          - name: REANA_KUBERNETES_JOBS_MAX_USER_TIMEOUT_LIMIT
+            value: !!str {{ .Values.kubernetes_jobs_max_user_timeout_limit | default 1209600 }}  # 1209600 seconds = 14 days
```
mvidalgarcia, line 117: Ahá! here's the value. I still think it would be helpful to have it in the README.

Didn't know `!!str` 👌 but env vars are always transformed to string, aren't they?

mvidalgarcia (commented):
1st superficial review, I haven't tested functionality.

helm/reana/templates/reana-workflow-controller.yaml
VMois: Agree. I only added it in the YAML file but it is hidden there. I will also add it to README.

VMois: > Didn't know !!str 👌 but env vars are always transformed to string, aren't they?

Yes, env variables are strings. But, I think, this is a [matter of Helm templating](https://github.com/helm/helm/issues/4775#issuecomment-430361046). You can input numbers in two ways: `10` and `"10"`. The first one is treated as a number by templating engine (and, actually, needs the` quote` prefix before). The second one treats as a string, but you cannot use `quote` on that. I found `!!str` to convert whatever is input to string.

VMois: I was thinking to allow for strings too. So, in the future, if we will go with "10h" suffixes it will require fewer changes. But, honestly, it was not that important for this PR set but I did what I did :smiley:

helm/reana/README.md
VMois: > Would it make sense to add "(7 days)" next to the default value?

Yes, it would. My only small worry is that it will be clear for admins that the number to enter is "604800" and not "7 days". Cause, as for now, they are not interchangeable. Entering "7 days" will not work. Although, in the description, I mentioned "*seconds*" so should be good.

 I will change it cause, I agree, a big number out of nowhere is not user-friendly.

VMois: *question:* would it make sense to point to our documentation page instead of Kubernetes docs?

helm/reana/templates/reana-workflow-controller.yaml
mvidalgarcia: It's OK for me, we can keep it.

helm/reana/README.md
mvidalgarcia: > > Would it make sense to add "(7 days)" next to the default value?
> 
> Yes, it would. My only small worry is that it will be clear for admins that the number to enter is "604800" and not "7 days". Cause, as for now, they are not interchangeable. Entering "7 days" will not work. Although, in the description, I mentioned "_seconds_" so should be good.
> 
> I will change it cause, I agree, a big number out of nowhere is not user-friendly.

We can add it as part of the "Description" instead of as part of the "Default" column, right after the k8s URL. What do you think?

mvidalgarcia:
I think it's important to add a changelog for this change.

CHANGES.rst
```diff
@@ -4,7 +4,12 @@ Changes
 Version 0.8.1 (UNRELEASED)
 --------------------------
 
+- Users:
+    - Adds support for specifying ``kubernetes_job_timeout`` for Kubernetes compute backend jobs for CWL, Serial, Snakemake, and Yadage workflows.
```
mvidalgarcia, line 8: Since the feature is available in all the workflow engines, perhaps we don't need to mention them?

mvidalgarcia (approved):

PR #615, opened 2022-02-09
helm: add k8s max user memory limit value to reana-server

audrium:
closes https://github.com/reanahub/reana-server/issues/437

### DIFF
--- a/helm/reana/templates/reana-server.yaml
+++ b/helm/reana/templates/reana-server.yaml
@@ -80,6 +80,10 @@ spec:
             value: {{ .Values.quota.default_cpu_limit | default 0 | quote }}
           - name: REANA_DEFAULT_QUOTA_DISK_LIMIT
             value: {{ .Values.quota.default_disk_limit | default 0 | quote }}
+          {{- if .Values.kubernetes_jobs_max_user_memory_limit }}
+          - name: REANA_KUBERNETES_JOBS_MAX_USER_MEMORY_LIMIT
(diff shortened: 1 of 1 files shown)


helm/reana/templates/reana-server.yaml
```diff
@@ -80,6 +80,10 @@ spec:
             value: {{ .Values.quota.default_cpu_limit | default 0 | quote }}
           - name: REANA_DEFAULT_QUOTA_DISK_LIMIT
             value: {{ .Values.quota.default_disk_limit | default 0 | quote }}
+          {{- if .Values.kubernetes_jobs_max_user_memory_limit }}
```
VMois, line 83: *note:* In case of max memory usage is not specified, the memory exceeds error will not show up during `reana-client run` command. The workflow will still fail with a strange error later from `job-controller`:

```
2022-02-09 15:13:40,111 | reana-workflow-engine-snakemake | MainThread | INFO | Environment: reanahub/reana-env-root6:6.18.04
2022-02-09 15:13:40,118 | reana-workflow-engine-snakemake | MainThread | ERROR | Error submitting job gendata: Expecting value: line 1 column 1 (char 0)
2022-02-09 15:13:40,118 | reana-workflow-engine-snakemake | MainThread | INFO | gendata job is failed. job_id: None
```

Because we do not have a default max memory limit it is kinda okay, but not ideal. 
Anyway, the issue with strange error from `job-controller` can be resolved separately later.

VMois (approved):

PR #616, opened 2022-02-15
helm: added scheduler requeue count environment variable

VMois:
closes reanahub/reana-server#436


helm/reana/README.md
```diff
@@ -16,6 +16,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `components.reana_server.environment.REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` | Upper limit on concurrent REANA batch workflows running in the cluster. | 30 |
 | `components.reana_server.environment.REANA_USER_EMAIL_CONFIRMATION` | Enable user to confirm their email address. | true |
 | `components.reana_server.environment.REANA_SCHEDULER_REQUEUE_SLEEP` | Seconds to wait between consuming workflows. | 15 |
+| `components.reana_server.environment.REANA_SCHEDULER_REQUEUE_COUNT` | Numbers of time to requeue workflow before failing it. | 200 |
```
audrium, line 19: Minor: perhaps `The number of times..` would be a bit better

audrium (commented):
- PR needs rebase:
image
- I think we need to add this new envar to `values.yaml` similarly as for `REANA_SCHEDULER_REQUEUE_SLEEP` to be able to easily access and modify the defaults
- Can you also drop a line in `CHANGES.rst` about this feature?

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Version 0.8.1 (UNRELEASED)
```
audrium, line 4: Minor: we should use 0.8.2

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
...
+--------------------------
+
+- Administrators:
+    - Adds new configuration option, ``reana_server.environment.REANA_SCHEDULER_REQUEUE_COUNT``, to set requeue count for submitted workflows in case of scheduling error.
```
audrium, line 8: Very minor: I think it should start with `Adds configuration environment variable` like with other envars (e.g. `REANA_SCHEDULER_REQUEUE_SLEEP`)

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Version 0.8.2 (UNRELEASED)
+--------------------------
+
+- Administrators:
+    - Adds configuration environment variable, ``reana_server.environment.REANA_SCHEDULER_REQUEUE_COUNT``, to set requeue count for submitted workflows in case of scheduling error.
```
audrium, line 8: Minor: we should also mention that cluster could be simply busy for the workflows to be requeued. Similarly as you mentioned [here](https://github.com/reanahub/reana-server/blob/5c046f7dde1d0881a3e0a99aa82881657aa0945b/reana_server/config.py)

Issue #286, opened 2020-04-02
Error message "Cannot start workflow"


alintulu:
After creating a rather large workflow (uploaded around 200 files) I get one of the two following error messages when trying to run it

```
$ reana-client run
Cannot start workflow workflow.6: 
60 per 1 minute

$ reana-client run
Cannot start workflow workflow.7: 
1000 per 1 hour
```

diegodelemos:
Currently, each file upload is a REST API call, which would always hit any reasonable [REST API rate limit](https://github.com/inveniosoftware/invenio-app/blob/867f3c33e8dd1b4fc31ff2ba6761ee580532ada3/invenio_app/config.py#L69-L90) with many files workflows. The idea exposed in https://github.com/reanahub/reana-client/issues/406 about uploading the whole analysis workspace to REANA seems like the long term solution for this issue:

> Uploading too many files at the same might be problematic for the server. A good strategy to alleviate this can be packaging the whole directory in a tarball and upload it.

So to address this issue we have to:
- [ ] In any case, if something like this happens we need to provide a better error message
- [ ] Address https://github.com/reanahub/reana-client/issues/406

tiborsimko:
We could amend the `reana-client` to respect the rate limits a bit better on the client side: 
- when uploading directories with many files, do some `sleep()` between them because we know this can cause troubles;
- even better, when seeing the rate limiter message in the response for any REST API call, do not give up but rather retry after N seconds.

diegodelemos:
Currently, none of the [Invenio App rate-limiting configurations](https://invenio-app.readthedocs.io/en/latest/configuration.html#invenio_app.config.RATELIMIT_APPLICATION) is set (see [here](https://github.com/reanahub/reana-server/blob/master/reana_server/config.py)). This causes the default rate-limiting values to be loaded causing the problem described in this issue.

We could either use Invenio-App's capability of rate-limiting or looking [into Traefik's rate-limiting](https://doc.traefik.io/traefik/middlewares/ratelimit/).

To avoid the problem at hand we could have, apart from the REANA-Client fix, no rate-limiting by default in the server side, setting it only on demand if provided as Helm config.

PR #619, opened 2022-03-09
helm: add rate limit env variable to launch workflow

VMois:
closes reanahub/reana-server#443

### DIFF
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -20,6 +20,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `components.reana_server.environment.REANA_WORKFLOW_SCHEDULING_POLICY` | Define workflow scheduling strategy. Options are "fifo" for first-in-first-out strategy regardless of users and "balanced" for multi-user-aware scheduling strategy. | "fifo" |
 | `components.reana_server.environment.REANA_RATELIMIT_GUEST_USER` | Set API limiter config for guest users. Users using reana-client will be treated as guests. | "20 per second" |
 | `components.reana_server.environment.REANA_RATELIMIT_AUTHENTICATED_USER` | Set API limiter config for authenticated web UI users. | "20 per second" |
+| `components.reana_server.environment.REANA_RATELIMIT_SLOW` | Set API limiter config for slow endpoints that need to be protected e.g. launch endpoint. | "1/5 second" |
 | `components.reana_server.image`                          | [REANA-Server image](https://hub.docker.com/r/reanahub/reana-server) to use          | `reanahub/reana-server:<chart-release-version>` |
 | `components.reana_server.imagePullPolicy`                | REANA-Server image pull policy                                                       | IfNotPresent                                    |
 | `components.reana_server.uwsgi.processes`                | Number of uWSGI processes                                                            | 6                                               |
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,7 @@ Version 0.9.0 (UNRELEASED)
 - Administrators:
     - Adds support for Kubernetes version 1.22 and 1.23 clusters.
     - Removes support for Kubernetes version prior to 1.19.
+    - Adds configuration environment variable ``reana_server.environment.REANA_RATELIMIT_SLOW`` to limit API requests to some protected endpoints e.g launch workflow.
 
 Version 0.8.2 (UNRELEASED)
 --------------------------
--- a/helm/reana/values.yaml
+++ b/helm/reana/values.yaml
@@ -73,6 +73,7 @@ components:
       REANA_WORKFLOW_SCHEDULING_POLICY: "fifo"
       REANA_RATELIMIT_GUEST_USER: "20 per second"
       REANA_RATELIMIT_AUTHENTICATED_USER: "20 per second"
+      REANA_RATELIMIT_SLOW: "1/5 second"
     uwsgi:
       processes: 6
       threads: 4


helm/reana/README.md
```diff
@@ -20,6 +20,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `components.reana_server.environment.REANA_WORKFLOW_SCHEDULING_POLICY` | Define workflow scheduling strategy. Options are "fifo" for first-in-first-out strategy regardless of users and "balanced" for multi-user-aware scheduling strategy. | "fifo" |
 | `components.reana_server.environment.REANA_RATELIMIT_GUEST_USER` | Set API limiter config for guest users. Users using reana-client will be treated as guests. | "20 per second" |
 | `components.reana_server.environment.REANA_RATELIMIT_AUTHENTICATED_USER` | Set API limiter config for authenticated web UI users. | "20 per second" |
+| `components.reana_server.environment.REANA_RATELIMIT_LAUNCH_WORKFLOW` | Set API limiter config for launch workflow endpoint. | "1/30 second" |
```
audrium, line 23: Tested the PR set, works nicely, just a couple of questions:

1) Perhaps it's worth to go for a more generic "fast" and a "slow" endpoint solution as described in the issue? Creating a separate envar only for `launch` endpoint seems to be too specific. WDYT? cc @tiborsimko 

2) "1/30 second" rate limits seems to be too strict for me? Perhaps we should relax it a bit to something like "1/5 second"

3) I haven't looked deeply, but would it be possible to return more user friendly error messages if rate limit is not respected? For example if user hits `Launch on REANA` web page twice in 30 seconds, this is what will be displayed:
Screenshot 2022-03-11 at 13-10-42 REANA

VMois: 2. "1/5 second" means "1 per 5 seconds" which is stricter than "1/30 second" which means "1 per 30 seconds". We can make it "1/15 second" which will mean "1 per 15 seconds". 

3. Good point. There are two ways how to make it nice: change UI or change backend. With the backend, I think, it is possible with `limiter` and `flask-limiter` libraries. But, we rely on `invenio-app` so I will need to check if they allow it somehow.

audrium: "1/30 second" allows a user to launch a workflow only once per 30 seconds
while "1/5 second" allows to do it every 5 seconds, seems less strict to me

VMois: Ops. You are right. My brain is not working.

tiborsimko: **WRT 1** I agree with @audrium, as we'll soon have other possibly "slow" endpoints, such as for validation of docker images that would need fetching, spawning pods, etc. Could create DoS kind of workload if someone overdoes those. So we could have the generic `REANA_RATELIMIT_GUEST_USER` and `REANA_RATELIMIT_AUTHENTICATED_USER` variables for default "fast" endpoints, as up to now, and we could introduce a new `REANA_RATELIMIT_SLOW` variable for "slow" endpoints that need to be protected, and start to applying them e.g. to workflow launcher, to notebook launcher, etc.

We could actually proactively introduce three values already, `*_FAST`, `*_MEDIUM`, `*_SLOW` if we want, and start tagging other endpoints as well.  For example, for ATLAS pMSSM use case, where there are many concurrent workflows running, one may want to go into `1000 per second` for some endpoints. So we could start already tagging via fast/medium/slow those endpoints which are needed for internal system working (and need to be ultra permissive), which are just for regular user consumption wondering about status (and which are medium default), which might be too time consuming (and need slow protection). 

(As for naming, fast/medium/slow might be enough, we could introduce "rocket" or something later if need be. Or we could invent some fancy naming scheme such as cheetah/rabbit/ant/snail that would also be nicely extensible in the future if there will be a need for more fine-grained categorisation than just fast/slow.).

**WRT 2** allowing 1 launch each 5 seconds seems reasonable.

**WRT 3** prettifying the output message would be a nice bonus. See also https://github.com/reanahub/reana/issues/286 how the rate limiter error message looks like in regular `reana-client` usage scenario if user has many input files. Currently it is not easily understandable by users. So improving it for both Web UI users and CLI UI users would be really nice to have.

VMois: 1. Changed to `REANA_RATELIMIT_SLOW`. I think we can do `FAST` and `MEDIUM` later when it is needed. I will create an issue.

2. Change to 1/5

3. Can be a separate issue to improve Web and CLI messages.

helm/reana/README.md
```diff
@@ -20,6 +20,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `components.reana_server.environment.REANA_WORKFLOW_SCHEDULING_POLICY` | Define workflow scheduling strategy. Options are "fifo" for first-in-first-out strategy regardless of users and "balanced" for multi-user-aware scheduling strategy. | "fifo" |
 | `components.reana_server.environment.REANA_RATELIMIT_GUEST_USER` | Set API limiter config for guest users. Users using reana-client will be treated as guests. | "20 per second" |
 | `components.reana_server.environment.REANA_RATELIMIT_AUTHENTICATED_USER` | Set API limiter config for authenticated web UI users. | "20 per second" |
+| `components.reana_server.environment.REANA_RATELIMIT_SLOW` | Set API limit config for slow endpoints. | "1/5 second" |
```
audrium, line 23: We could enhance the message for the user to give a bit more context:
`Set API limiter config for slow endpoints that need to be protected e.g. launch endpoint`
or something similar, WDYT?

Also, we could add a similar message to the changelog

helm/reana/README.md
audrium: Created https://github.com/reanahub/reana-server/issues/455 to address the 3rd point

audrium (approved):
Works nicely!

PR #622, opened 2022-03-23
helm: relax rate limiter value for local developments

audrium:
Issue as described by @mdonadoni : 
`for each launch request  two requests are actually done (OPTIONS followed by POST) and both of them count towards the rate limit. Given that the limit is "1/5 second", the POST request is always blocked and returns an error`
This PR modifies the rate limiter value to be more permissive for local developments


mdonadoni (approved):
Nice! Works as expected

PR #623, opened 2022-03-25
helm: lower job-status-consumer prefetch value

VMois:
- set job-status-consumer RabbitMQ prefetch value to 10;
- this should reduce the risk of getting 406 PRECONDITION errors on message ack.

addresses reanahub/reana-workflow-controller#436


audrium:
Couple minor things:
- CI is blocked, could you please amend your commit and repush to trigger it again?
- We have default prefetch count value also set [here](https://github.com/reanahub/reana-workflow-controller/commit/ca903f4f964af14abe0d85f4fd49f5c478b16fed#diff-5e7c48d2c3a73afca6915c77539bfdd349bca550379c93341656e8475e33dac7R52), not sure if it's very important to change it there as well, but may be good for consistency.
- After reading [the comment](https://groups.google.com/g/rabbitmq-users/c/X-x3DKVJb8k/m/wTKmcEd7BwAJ) you provided in the issue, it seems that the RabbitMQ closes the connection when consumer pod gets restarted. I wonder what would be the way to handle reconnections and would making the queue `durable` help.. Ofc this is outside the PR, just thinking out loud.

VMois:
> I wonder what would be the way to handle reconnections and would making the queue durable help

`workflow-submission` should be durable, `jobs-status` is not durable (according to [r-commons config](https://github.com/reanahub/reana-commons/blob/eb9c132e232ba79a77979cca39868d632c8a00ac/reana_commons/config.py#L191-L200)). In addition, we do not `auto ack` messages, so it doesn't look like we should lose messages on consumer restart.

VMois:
> CI is blocked, could you please amend your commit and repush to trigger it again?

It should get unblocked in this PR - https://github.com/reanahub/reana/pull/624. We can merge this PR after.

tiborsimko (approved):

Issue #627, opened 2022-03-31
reana-dev release-helm: fix github release creation troubles


tiborsimko:
For the release of 0.9.0-alpha.3, the automated release scripts were not working:

```console
$ reana-dev release-helm
...
[2022-03-31T17:12:21] reana: cr upload -o reanahub -r reana --release-name-template '{{ .Version }}' --commit 59d92c9
Error: error creating GitHub release: POST https://api.github.com/repos/reanahub/reana/releases: 422 Validation Failed [{Resource:Release Field:target_commitish Code:invalid Message:}]
```

I had to create the release manually (i.e. create and push tag, attach chart tarball, edit gh-pages entry with proper digest, etc).

Let's look at the issue when time permits.

PR #632, opened 2022-04-14
cli: add target commit to `release-helm` only for maint branches

audrium:
closes https://github.com/reanahub/reana/issues/627

Issue was caused by the `commit` flag while executing `cr upload`. It is only needed for `maint` branches. This PR introduces an additional `--maint` option to conditionally add this flag.

To test:
1. Create a new release commit (`reana-dev git-create-release-commit -c .`)
2. Create a new tag (`reana-dev git-tag -c .`)
3. `reana-dev release-helm --dry-run` should display:
```
...
cr upload -o reanahub -r reana --release-name-template '{{ .Version }}'
...
```
4. `reana-dev release-helm --dry-run --maint` should display:
```
...
cr upload -o reanahub -r reana --release-name-template '{{ .Version }}' --commit b9f51e4
...
```
5. Remove tag (`git tag -d 0.9.0-alpha.5`)
6. Remove release commit (`git reset --hard HEAD~1`)

### DIFF
--- a/reana/reana_dev/release.py
+++ b/reana/reana_dev/release.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2020 CERN.
+# Copyright (C) 2020, 2022 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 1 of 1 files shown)


reana/reana_dev/release.py
```diff
@@ -190,6 +190,12 @@ def release_pypi(ctx, component, timeout):  # noqa: D301
 
 
 @click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option(
+    "--maint",
```
tiborsimko, line 194: Instead of calling the new option `--maint`, I would call it `--commit` (to be conform to `cr` option name) or `--cr-commit` (to be very explicit how it is used) with the help message saying something like "specify target commit for release? [default=no]" and explain in detail in the docstring that `--cr-commit` is to be used only when releasing non-master branches such as `maint-0.9` or `next`.

(Ideally, we should be able to always use this also for master branch, so that there would be no need for such a flag, but this solution would be as an MVP.)

tiborsimko (commented):
Works well, left one cosmetic comment WRT option naming.

tiborsimko (approved):

PR #634, opened 2022-04-21
config: add new DEMO examples

### LINKED ISSUE #633: cli: richer `reana-dev run-example -c DEMO` example set
Enrich a list of runnable demo examples so that the command `reana-dev run-example -c DEMO` would more completely test the ones we now are working well

This requires to amend `REPO_LIST_DEMO_RUNNABLE` and add the following ones:

- reana-demo-alice-lego-train-test-run
- reana-demo-alice-pt-analysis
- reana-demo-bsm-search
- reana-demo-cms-h4l 
- reana-demo-cms-dimuon-mass-spectrum

Some challenges:

- e.g. BSM uses `reana.yaml` but it is Yadage, so it requires similar change as atlas recast example when generating appropriate workflow name;
- e.g. CMS H4L only works well for Snakemake, so use only this example for now (we have an issue for fixing CWL there);
- don't forget to change `EXAMPLE_OUTPUT_FILENAMES` and `EXAMPLE_LOG_MESSAGES` for output checking. The log messages and output files to check for should be "unique enough" to make sure that it works well;
- don't forget to identify which images to pre-fetch in `DOCKER_PREFETCH_IMAGES`.

The ultimate goal is that `reana-dev run-example -c DEMO` will run virtually all of our examples (=hence slow, but exhaustive) so that it can provide complete testing matrix for nightly integrations etc. Whilst individual developers could still use `reana-dev run-example -c .` or `reana-dev run-example -c r-d-helloworld` to run only individual examples (=hence fast, but less covering) when developing some novel workflow features.

VMois:
- because some new workflows take longer to finish, increase the maximum timeout to wait for workflow results to 15 minutes.

closes #633

### DIFF
--- a/reana/config.py
+++ b/reana/config.py
@@ -9,22 +9,22 @@
 """``reana-dev`` CLI configuration."""
 
 REPO_LIST_DEMO_RUNNABLE = [
+    "reana-demo-alice-lego-train-test-run",
+    "reana-demo-alice-pt-analysis",
+    "reana-demo-atlas-recast",
+    "reana-demo-bsm-search",
+    "reana-demo-cms-dimuon-mass-spectrum",
+    "reana-demo-cms-h4l",
     "reana-demo-helloworld",
     "reana-demo-root6-roofit",
     "reana-demo-worldpopulation",
-    "reana-demo-atlas-recast",
 ]
 """All git repositories containing REANA runnable demos."""
 
 REANA_LIST_DEMO_ALL = REPO_LIST_DEMO_RUNNABLE + [
-    "reana-demo-alice-lego-train-test-run",
-    "reana-demo-alice-pt-analysis",
-    "reana-demo-bsm-search",
     "reana-demo-cdci-crab-pulsar-integral-verification",
     "reana-demo-cdci-integral-data-reduction",
-    "reana-demo-cms-h4l",
     "reana-demo-cms-reco",
-    "reana-demo-cms-dimuon-mass-spectrum",
     "reana-demo-fcchh-fullsim",
     "reana-demo-lhcb-d2pimumu",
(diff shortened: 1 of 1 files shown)


reana/config.py
```diff
@@ -13,6 +13,11 @@
     "reana-demo-root6-roofit",
     "reana-demo-worldpopulation",
     "reana-demo-atlas-recast",
+    "reana-demo-bsm-search",
+    "reana-demo-cms-h4l",
+    "reana-demo-cms-dimuon-mass-spectrum",
+    "reana-demo-alice-pt-analysis",
+    "reana-demo-alice-lego-train-test-run",
```
tiborsimko, line 20: You may want to remove them from `REANA_LIST_DEMO_ALL` below so that they won't appear twice there.

reana/config.py
```diff
@@ -13,18 +13,18 @@
...
```
tiborsimko, line 20: Cosmetics: please sort lines in `REPO_LIST_DEMO_RUNNABLE` alphabetically.

reana/config.py
```diff
@@ -217,13 +231,25 @@
         "reanahub/reana-demo-bsm-search:1.0.0",
         "reanahub/reana-env-root6:6.18.04",
     ],
+    "reana-demo-cms-h4l": [
+        "cmsopendata/cmssw_5_3_32",
+    ],
+    "reana-demo-cms-dimuon-mass-spectrum": [
+        "cmsopendata/cmssw_5_3_3",
```
tiborsimko, line 238: Should be `5_3_32` as well.

VMois: Good catch!

tiborsimko (approved):
Works nicely :+1:

PR #643, opened 2022-05-27
helm: Kubernetes 1.24

tiborsimko:
Declares support for Kubernetes 1.24 that was successfully tested
locally using Kind 0.14.


mdonadoni (approved):
Tested locally with kind 0.13 and everything is working well!

Issue #647, opened 2022-06-21
Unable to deploy cluster locally


arm2arm:
Hello,
I was following the steps described here: https://docs.reana.io/administration/deployment/deploying-locally/
In the helm install I am getting an error on versions mismatch
```
helm install reana reanahub/reana --namespace reana --create-namespace --wait
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.13.0-0 < 1.22.0-0 which is incompatible with Kubernetes v1.24.0
```
I try to downgrade the kubectl but it is still bringing the server version above it
```
/usr/local/bin/kubectl version --short 
Client Version: v1.22.11
Server Version: v1.24.0
WARNING: version difference between client (1.22) and server (1.24) exceeds the supported minor version skew of +/-1
```
I appreciate any workaround.

Thank you beforehand.

tiborsimko:
Hi @arm2arm:

Kubernetes v1.24 clusters are supported only in REANA's bleeding-edge `master` branch sources. It works well, but we have not released the new Helm chart yet. Hence it does not work out of the box yet.

Are you just trying out REANA locally using Kind or are you thinking about deploying on a "proper" Kubernetes v1.24 cluster later?

If mostly the former, the easiest solution to test locally would be to force Kind to use an older Kubernetes cluster version. Assuming you are using the latest stable Kind v0.14 release:

```
$ kind version
kind v0.14.0 go1.18.2 linux/amd64
```

you can use the following `--image` option value to bring up Kubernetes v1.21 instead of the default v1.24:

```
$ kind delete cluster
$ kind create cluster --config kind-localhost-30443.yaml --image=kindest/node:v1.21.12
```

This is the only change necessary to install REANA 0.8.1 stable release locally on your Kind v0.14.

Can you please give it a try?

P.S. Another useful trick, if you are interested in testing out our REANA 0.9.0-alpha.* developer preview versions from our [release list](https://reanahub.github.io/reana/), would be the following: 

```
$ helm install ... --version 0.9.0-alpha.5
```

This would bring up REANA 0.9.0 latest alpha release instead of the above 0.8.1 latest production release.

(And, once we release `0.9.0-alpha.6` officially, Kind v0.14 will be supported out of the box, so the above `kind create ... --image` option would not be necessary anymore.)

arm2arm:
Hi @tiborsimko,
Thank you for the excellent answer! The kind option: **--image=kindest/node:v1.21.12** did the trick.
[...]

PR #649, opened 2022-06-24
Pin K8s version in Kind config

mzur:
The Helm chart of v0.8 requires K8s <1.22.0 which will break with the most recent version of kindest/node.

References: https://github.com/reanahub/reana/blob/ed48d7eedf92ff64683a2436c890aa49df70f056/helm/reana/Chart.yaml#L31

### DIFF
--- a/etc/kind-localhost-30443.yaml
+++ b/etc/kind-localhost-30443.yaml
@@ -12,3 +12,4 @@ nodes:
           kubeletExtraArgs:
             node-labels: "ingress-ready=true"
     role: control-plane
+    image: kindest/node:v1.21.12@sha256:f316b33dd88f8196379f38feb80545ef3ed44d9197dca1bfd48bcb1583210207


etc/kind-localhost-30443.yaml
```diff
@@ -12,3 +12,4 @@ nodes:
           kubeletExtraArgs:
             node-labels: "ingress-ready=true"
     role: control-plane
+    image: kindest/node:v1.21.12@sha256:f316b33dd88f8196379f38feb80545ef3ed44d9197dca1bfd48bcb1583210207
```
tiborsimko, line 15: Hi @mzur, this fix would be OK for Kind 0.14, but will it work well with older Kind versions? Have you tried?

Note that I suggested to use `kind create ... --image=kindest/node:v1.21.12` without the SHA1 part in https://github.com/reanahub/reana/issues/647#issuecomment-1161689205 simply to be more welcoming to older Kind versions, such as 0.13, which may perhaps prefer older SHA1 builds.

In other words, I'm concerned that if we force a particular SHA1 image version here in YAML, it may cause troubles for people using older Kind clients. Especially if somebody uses Kind 0.11 still, they will get Kubernetes 1.21 out of the box, so they wouldn't really need to change anything in the YAML template -- and if we do force this SHA1 here, perhaps their deployment may not work? (To be tested.)

TL;DR if this change could cause troubles for people using older Kind versions, I would rather leave Kubernetes version pinning from the YAML template, and work with this on the CLI level, advising people to install 1.21 depending on which Kind version they use.

mzur: I'm no expert with K8s/Kind and was following the [documentation](https://kind.sigs.k8s.io/docs/user/quick-start/#setting-kubernetes-version) that recommended to specify the checksum. 

I totally missed https://github.com/reanahub/reana/issues/647 which is about the exact issue that I wanted to solve with this PR, sorry. I agree that this can also be fixed with an update of the documentation to pin the K8s version. Please feel free to close this PR.

PR #650, opened 2022-06-24
helm: add possibility to generate self-signed certificate

mdonadoni:
When opening or closing interactive sessions, `Ingress` objects are
either created or deleted to correctly route requests to the responsible
containers. Each time this happens, Traefik reloads its configuration,
and its default self-signed certificate gets regenerated.

This causes problems with Jupyter notebooks, in particular when trying
to create a new WebSocket. Self-signed certificates are not considered
valid by browsers, unless users explicitly trust them. This means that,
when the certificate is regenerated, browsers do not consider it valid,
given that it was never trusted by the user. This causes the creation of
a new WebSocket connection to fail.

This commit makes it possible to generate a self-signed certificate on
deploy, which will then be used by Traefik, overriding its default one.

Closes reanahub/reana-workflow-controller#457

How to test:
1. Deploy REANA
2. Launch some workflows
3. Open and close some sessions
4. Open a new Jupyter notebook

Expected result: the notebook works correctly and there aren't any connection errors


helm/reana/templates/secrets.yaml
```diff
@@ -60,3 +60,18 @@ data:
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "reana.prefix" . }}-tls-secret
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/resource-policy": keep
+type: kubernetes.io/tls
+data:
+  {{- $cert := genSelfSignedCert (.Values.reana_hostname | default "localhost") nil nil 365 }}
```
tiborsimko, line 74: To be on the safe side, we can perhaps use 10 years instead of 1 year for self-generated certificates, just in case somebody might use it on a longer-lasting DEV server somewhere?

mdonadoni: I have changed it to 10 years. We should however consider that, even though this is a self-signed certificate, once the user trusts it in the browser it will remain trusted forever. This can become a security issue: if somebody gets access to the certificate it can impersonate the server for a very long time, given the very long duration of the certificate. What do you think?

helm/reana/values.yaml
```diff
@@ -113,6 +113,8 @@ ingress:
     kubernetes.io/ingress.class: traefik
     traefik.frontend.entryPoints: "http,https"
     ingress.kubernetes.io/ssl-redirect: "true"
+  tls:
+    self_signed_cert: true
```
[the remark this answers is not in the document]
tiborsimko: ... not applicable; we'll switch it off for proper Let's Encrypt certificate use case scenarios in local `values.yaml` that would override this default. I should multi-task less before commenting :sweat_smile:

PR #658, opened 2022-07-27
helm: change default value for retention days to "forever"

### LINKED ISSUE #653: helm: let administrators disable the default retention period
[...]
2. I think it should be possible to disable the maximum retention period, as we do for the disk/cpu quota limit. WDYT?
[...]

_Originally posted by @mdonadoni in https://github.com/reanahub/reana/issues/652#issuecomment-1172474025_

In case we decide to provide a way to disable the default retention period, we would also need to modify how we handle the `WORKSPACE_RETENTION_PERIOD` env variable in reana-server.

VMois:
closes #653


mdonadoni (approved):
LGTM!

PR #659, opened 2022-08-15
helm: add Kubernetes job timeouts to reana-server template

VMois:
closes reanahub/reana-client#622


tiborsimko (commented):
LGTM. I'm a bit worried that we have two occurrences where these values appear for different components, so we don't have a "single source of truth" where to edit the defaults, as it were -- but we already have other variables of this kind, so we'll just have to make sure to propagate value edits everywhere they appear.

tiborsimko (approved):

PR #691, opened 2023-01-23
uwsgi: enable keep-alive support

mdonadoni:
Enable support for keep-alive in uWSGI's configuration. This fixes an
issue with Traefik sporadically returning 500 or 502 status codes, with
logs reporting errors such as `http: server closed idle connection` and
`write: broken pipe`.

See
https://uwsgi-docs.readthedocs.io/en/latest/HTTP.html#http-keep-alive
for more information about keep-alive in uWSGI.

How to test:
1. Create a new workflow: `reana-client create -w workflow`
2. Run many concurrent uploads: ``for i in `seq 50`; do reana-client upload -w workflow.1 &; done; wait``

### DIFF
--- a/helm/reana/templates/uwsgi-config.yaml
+++ b/helm/reana/templates/uwsgi-config.yaml
@@ -8,7 +8,7 @@ data:
   uwsgi.ini: |
     [uwsgi]
     stats = /tmp/stats.socket
-    http-socket = 0.0.0.0:5000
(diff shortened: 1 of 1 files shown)


tiborsimko (approved):
Works well locally with 6 uswgi workers, including BSM input files :+1: 

Shall we give it a try on reana-dev too to test with CERN K8s's traefik?

PR #692, opened 2023-02-09
auth: add values for Keycloak SSO to Helm Chart

goseind:
This PR adds values for the Keycloak SSO Sign-in option to the Helm Chart.

Refer to https://github.com/reanahub/reana-server/issues/513 for more details.

Fixes https://github.com/reanahub/reana-server/issues/513


helm/reana/README.md
```diff
@@ -95,6 +95,14 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `secrets.gitlab.REANA_GITLAB_OAUTH_APP_ID`               | GitLab OAuth application id                                                          | None                                            |
 | `secrets.gitlab.REANA_GITLAB_OAUTH_APP_SECRET`           | **[Do not use in production, use secrets instead]** GitLab OAuth application secret  | None                                            |
 | `secrets.reana.REANA_SECRET_KEY`                         | **[Do not use in production, use secrets instead]** REANA encryption secret key      | None                                            |
+| `sso.enabled`                                            | Enable and configure your own third-party authentication service                     | false                                           |
```
tiborsimko: What about keeping `sso.*` values for the old `cern_openid` technique, and introducing `keycloak_*` values for the new Keycloak based way?

(And we'll deprecate `sso.*` in the future.)

goseind: I changed the naming to `keycloak*` and moved the key and secret to `secrets.keycloak` as well.

helm/reana/values.yaml
```diff
@@ -49,6 +49,20 @@ secrets:
...
+    consumer_secret: ""
+
+# Keycloak configuration (also refer to secrets.keycloak.*)
+keycloak_config:
+  enabled: false
+  title: ""
+  desciption: ""
+  base_url: ""
+  auth_url: ""
+  token_url: ""
```
mdonadoni, line 63: I think we should also add the `OAUTHCLIENT_KEYCLOAK_REALM_URL`, as I think that the URL generated by the helper is probably going to be wrong with deployments of keycloak that do not use realms

helm/reana/values.yaml
```diff
@@ -49,6 +49,20 @@ secrets:
...
+  userinfo_url: ""
+  realm: ""
```
mdonadoni, line 65: I see that we give administrators the possibility to provide the realm. Does this mean that you would like to support both providing `base_url` + `realm` or providing each URL separately?

mdonadoni (commented):
Looks good! I will take a look again when all the PRs will be ready, but in the meantime I have left some small comments.

Regarding `values.yaml`, we can name `keycloak_config` in a couple different ways:
1. If we don't plan on supporting other SSO providers and do not plan on deploying keycloak as part of the Helm Chart, then `keycloak.*` should be enough. Otherwise, settings like `keycloak.enabled` might be confusing (does this mean that logging in with keycloak is supported, that keycloak is being deployed or both?)
2. If we want to be more future-proof, we can already add `keycloak` inside a more general section, for example:

    ```yaml
    login:
      keycloak:
        title: ...
        base_url: ...
        ...
    ```

    This will let us add more providers in an easier way:

    ```yaml
    login:
      keycloak:
        ...
      github:
        ...
    ```

    However, this might only become useful in a very far future and it does not support expressing multiple login providers of the same type anyway (e.g. multiple keycloak instances), so maybe it is not worth it.

3. To support having multiple authentications of the same type, we would need something similar to this:

    ```yaml
    login:
      - name: escape-iam
        type: keycloak
        ...
      - name: another-keycloak-instance
        type: keycloak
        ...
    ```

    but this requires some work on reana-server to parse this and generate the correct configuration, and some work on reana-ui to show multiple login buttons. We would also need to investigate whether invenio-oauthclient supports this kind of flexibility.
    
    

@tiborsimko @audrium what do you think about this?

helm/reana/values.yaml
```diff
@@ -49,6 +49,20 @@ secrets:
...
   reana: {}
+  keycloak:
+    consumer_key: ""
+    consumer_secret: ""
+
+# Keycloak configuration (also refer to secrets.keycloak.*)
+keycloak_config:
+  enabled: false
+  title: ""
+  desciption: ""
```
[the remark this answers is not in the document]
goseind: It is required by the helper atm, but as discussed it would make sense to disregard the helper and set everything manually, then this value would become obsolete.

helm/reana/values.yaml
goseind: Probably not. Best to set everything manually, but for that, I still need to do some changes to the server.

tiborsimko:
> 3\. To support having multiple authentications of the same type, we would need something similar to this:

This technique looks the most forward-looking, so I would tend to prefer that. Even if for now we would only implement support only one element in the `login` list.

@goseind I think you only need one ESCAPE IAM button, is that right? (Even though https://github.com/reanahub/reana-ui/pull/311 shows two.)

goseind:
> > 3. To support having multiple authentications of the same type, we would need something similar to this:
> 
> This technique looks the most forward-looking, so I would tend to prefer that. Even if for now we would only implement support only one element in the `login` list.
> 
> @goseind I think you only need one ESCAPE IAM button, is that right? (Even though [reanahub/reana-ui#311](https://github.com/reanahub/reana-ui/pull/311) shows two.)

We only need one button that's correct, but the way I understand the third option described by Marco and also judging from your comment [here](https://github.com/reanahub/reana-ui/pull/311#discussion_r1114300525) in the end there would be multiple login buttons depending on how many are defined in the values list. Before changing anything I just wanted to confirm whether there would always just be **one** login button or multiple ones dynamically created. The second option would require a lot more changes to the template and the server of course.

helm/reana/values.yaml
```diff
@@ -53,6 +53,10 @@ secrets:
   cern:
     sso: {}
   reana: {}
+  login: {}
+
+# Enable and configure SSO authentication via a third-party Keycloak identity provider
+login: {}
```
mdonadoni, line 59:
```suggestion
login: []
```
The default value should be `[]` instead of `{}`, what do you think?

helm/reana/README.md
```diff
@@ -96,6 +96,8 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
 | `secrets.gitlab.REANA_GITLAB_OAUTH_APP_SECRET`           | **[Do not use in production, use secrets instead]** GitLab OAuth application secret  | None                                            |
 | `secrets.reana.REANA_SECRET_KEY`                         | **[Do not use in production, use secrets instead]** REANA encryption secret key      | None                                            |
+| `login`                                                  | Enable and configure SSO authentication via a third-party Keycloak identity provider ([configuration details](https://docs.reana.io/administration/configuration/configuring-access/#keycloak-single-sign-on-configuration)) | None |
+| `secrets.login`                                          | **[Do not use in production, use secrets instead]** Third-party Keycloak identity provider consumer key and secret ([configuration details](https://docs.reana.io/administration/configuration/configuring-access/#keycloak-single-sign-on-configuration)) | None |
```
mdonadoni, line 100: Let's change `None` with the actual default value (`[]` for `login`, `{}` for `secrets.login`). 

Could you also move them so that they are in alphabetical order?

helm/reana/templates/secrets.yaml
```diff
@@ -23,6 +23,19 @@ data:
   CERN_CONSUMER_KEY: {{ .Values.secrets.cern.sso.CERN_CONSUMER_KEY | default "cern_consumer_key" | b64enc }}
   CERN_CONSUMER_SECRET: {{ .Values.secrets.cern.sso.CERN_CONSUMER_SECRET | default "cern_consumer_secret" | b64enc }}
 ---
+{{- if .Values.login }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "reana.prefix" . }}-login-provider-secrets
```
mdonadoni, line 30: extremely minor: let's call this `...-login-providers-secrets` for consistency with `LOGIN_PROVIDERS_SECRETS`?

helm/reana/templates/secrets.yaml
```diff
@@ -23,6 +23,19 @@ data:
...
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/resource-policy": keep
+type: Opaque
+data:
+  PROVIDER_SECRETS: {{ .Values.secrets.login | toJson | b64enc }}
```
mdonadoni, line 36: as before, let's call this `LOGIN_PROVIDERS_SECRETS` for consistency with the rest?

PR #693, opened 2023-02-09
helm: add cronjob to autoclose interactive sessions

audrium:
closes https://github.com/reanahub/reana-workflow-controller/issues/498


audrium:
One possible addition could be to pass `maximum_duration` value set by administrators down to `reana-ui` and to enrich  the current notification with a message indicating that the session will be closed after a specified number of days. I've quickly checked that it's not feasible to show the notification in the actual jupyter notebook itself, but we could at least inform users in REANA.

helm/reana/values.yaml
```diff
@@ -20,6 +20,10 @@ workspaces:
   paths:
     - /var/reana:/var/reana
 
+interactive_sessions:
+  cronjob_schedule: "0 3 * * *"  # everyday at 3am
+  maximum_duration: forever
```
mdonadoni, line 25: I think this might be a little bit misleading, as an administrator might think that this is the maximum duration of an interactive session since the time it was started. What about naming it `maximum_inactivity_period` or something similar to this?

The description of the value in README.md and in CHANGES.rst would also need to be changed

helm/reana/templates/cronjobs.yaml
```diff
@@ -257,3 +257,82 @@ spec:
           - name: reana-code
             hostPath:
               path: /code/reana-server
+---
+{{- if not (contains "forever" (.Values.interactive_sessions.maximum_duration | quote)) }}
```
mdonadoni, line 261: We are repeating the definition of cronjobs many times, resulting in a lot of copy-paste. Should we investigate whether we can reduce duplication, perhaps by defining new templates?

We can do this in another issue/PR.

audrium: Created an issue for this https://github.com/reanahub/reana/issues/695

mdonadoni (approved):

PR #694, opened 2023-02-16
helm: add new ingress resource that redirects http requests to https

audrium:
Latest k8s templates at CERN are using traefik v2 version which requires to define a new ingress resource to redirect HTTP requests through traefik middleware.

More information: https://kubernetes.docs.cern.ch/docs/networking/services/


helm/reana/Chart.yaml
```diff
@@ -31,7 +31,7 @@ version: 0.9.0
 kubeVersion: ">= 1.19.0-0 < 1.26.0-0"
 dependencies:
   - name: traefik
-    version: 10.3.4
+    version: 10.6.2
```
audrium, line 34: Updated to the same version used by the latest `kubernetes-1.25.3-3` CERN k8s template

helm/reana/values.yaml
```diff
@@ -124,7 +124,7 @@ ingress:
   enabled: true
   annotations:
     kubernetes.io/ingress.class: traefik
-    traefik.frontend.entryPoints: "http,https"
+    traefik.ingress.kubernetes.io/router.entrypoints: web,websecure
     ingress.kubernetes.io/ssl-redirect: "true"
```
mdonadoni, line 128: Is `ingress.kubernetes.io/ssl-redirect` still needed with traefik v2?

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Version 0.9.1 (UNRELEASED)
 - Administrators:
     - Adds new configuration option ``interactive_sessions.maximum_inactivity_period`` to set a limit in days for the maximum inactivity period of interactive sessions after which they will be closed.
     - Adds new configuration option ``interactive_sessions.cronjob_schedule`` to set how often interactive session cleanup should be performed.
+    - Adds new Ingress resource that redirects HTTP requests to HTTPS.
```
mdonadoni, line 10: Can you add that this is needed to support traefik v2?

helm/reana/values.yaml
audrium: Not sure about this one, I also wanted to remove it as it seems it's used for traefik 1.7, but after testing it on DEV it didn't work without it. It could be that I have tested not the final version of the new ingress resource, so we could re-test it again with custom Docker image on DEV

PR #700, opened 2023-03-02
helm: set minimum k8s version to 1.21

mdonadoni:


tiborsimko (approved):
Note: needed for cron jobs.

PR #714, opened 2023-06-06
helm: add missing env variables to cronjobs

### LINKED ISSUE #711: deployment: wrong database hostname in cronjobs
When choosing a release name different from "reana" (e.g. "reanadev"), the cronjobs incorrectly use `reana-db.default.svc.cluster.local` instead of `reanadev-db.default.svc.cluster.local` to connect to the database. Cluster components (e.g reana-server) use the correct hostname/domain.

The [`REANA_COMPONENT_PREFIX`](https://github.com/reanahub/reana-commons/blob/c7ed1bb45be61655b18a9c5b16337e381126c183/reana_commons/config.py#L48) environment variable should already take care of this, so we should check whether it is set correctly for cronjobs.

mdonadoni:
Add missing environment variables that caused cronjobs to fail to
connect to the database when REANA is deployed with a non-default
namespace or prefix.

Closes #711

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Version 0.9.1 (UNRELEASED)
     - Adds new configuration option ``notifications.email_config.smtp_starttls`` to use the STARTTLS command to enable encryption after connecting to the SMTP email server.
     - Changes uWSGI configuration to add vacuuming of generated files and sockets.
     - Fixes uWSGI memory consumption on systems with very high allowed number of open files.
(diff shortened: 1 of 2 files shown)


mdonadoni:
How to test:
1. Deploy REANA locally with non-default namespace and/or prefix
2. Trigger `resource-quota-update` cronjob

The cronjob should run successfully, instead of having database connection issues

giuseppe-steduto (approved):
LGTM! Fixes the error as expected.

PR #715, opened 2023-06-08
helm: fix enabling/disabling Ingresses

mdonadoni:
Fix ``ingress.enabled`` option to correctly enable or disable the
creation of Ingresses.


audrium (approved):
Works nicely

PR #717, opened 2023-06-09
reana-dev: fix deploy/undeploy with custom instance name/namespace

### LINKED ISSUE #716: reana-dev: cluster-deploy waits forever when providing custom instance name
How to reproduce:
1. Deploy the cluster, providing a custom instance name:

    ```
    reana-dev cluster-deploy --admin-email john.doe@example.org --admin-password mysecretpassword --instance-name reana-xyz
    ```
3. The script `scripts/create-admin-user.sh default reana-xyz john.doe@example.org mysecretpassword` will wait forever: `Waiting for deployment/reana-xyz-db to be ready...`

mdonadoni:
Closes #716

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,7 +18,7 @@ Version 0.9.1 (UNRELEASED)
     - Fixes ``ingress.enabled`` option to correctly enable or disable the creation of Ingresses.
     - Adds new configuration options ``login`` and ``secrets.login`` for configuring Keycloak SSO login with third-party authentication services.
 - Developers:
-    - Fixes ``cluster-deploy`` and ``cluster-undeploy`` commands when using custom instance name or kubernetes namespace.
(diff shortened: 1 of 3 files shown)


giuseppe-steduto (approved):
LGTM and works as expected!

Side note: even though `cluster-deploy` and `cluster-undeploy` with custom instance name now look fixed, there are several other commands in `reana-dev` that use the default `reana` instance name, such as `client-setup-environment` and some of the `git-` related ones. 
Probably it's a low priority task now, especially because `reana-dev` is mainly used internally, but we may want to fix those and make them more general-purpose. What do you think?

mdonadoni:
> Side note: even though `cluster-deploy` and `cluster-undeploy` with a custom instance name now look fixed, the `client-setup-environment` command still uses the default `reana` instance name.

Fixed!

PR #722, opened 2023-06-29
cronjobs: add secret key and admin access token to environment

mdonadoni:
Add `REANA_SECRET_KEY` and `REANA_ADMIN_ACCESS_TOKEN` environment
variables to cronjobs. The former is needed to decrypt the tokens in the
database, the latter is needed to execute `reana-admin` commands.

Closes reanahub/reana-server#489


giuseppe-steduto (approved):
Works well!

PR #726, opened 2023-07-06
helm: fix SSL/STARTTLS configuration

### LINKED ISSUE #725: helm: cannot set `smtp_starttls` to false
Even when `notifications.email_config.smtp_starttls` is set to `false`, it will be set to `true` in the pods' environments.

https://github.com/reanahub/reana/blob/0a0c4c69ad6adfe46b954fde927ec605c23edd95/helm/reana/templates/reana-server.yaml#L196

The `default` function treats `false` as en empty value, thus setting the environment variable to `true`.

Note that Helm suggests to put all the default values in `values.yaml`, and not to use `default`:
_In an actual chart, all static default values should live in the values.yaml, and should not be repeated using the default command (otherwise they would be redundant)_ ([source](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function))

mdonadoni:
Avoid using `default` when setting boolean values, as `false` is
interpreted as an empty value.

Closes #725

How to test:
1. Set `smtp_starttls` and `smtp_ssl` to different values, like:

    ```yaml
    notifications:
      enabled: true
      email_config:
        smtp_starttls: false
    ```
2. Check that the values are correctly set inside the pods:
    ```
    kubectl exec deployment/reana-server -- env | grep -i smtp
    ```

### DIFF
--- a/helm/reana/templates/cronjobs.yaml
+++ b/helm/reana/templates/cronjobs.yaml
@@ -61,13 +61,13 @@ spec:
             - name: REANA_EMAIL_SMTP_PORT
               value: "{{ .Values.notifications.email_config.smtp_port | default "30025" }}"
             - name: REANA_EMAIL_SMTP_SSL
-              value: {{ .Values.notifications.email_config.smtp_ssl | default "false" | quote }}
+              value: {{ .Values.notifications.email_config.smtp_ssl | quote }}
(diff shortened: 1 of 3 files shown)


tiborsimko (approved):
Works nicely

PR #738, opened 2023-08-24
helm: add r-server env variable for enabled notifications

### LINKED ISSUE #736: helm: sending emails fails if notifications are disabled
REANA sends emails even if `notifications.enabled` is set to false, for example after a user signs up or after a token request. However, the email sending fails as the necessary env variables are not set:

```
reana_commons.errors.REANAEmailNotificationError: Cannot send email, missing server and port configuration. Please provide the following environment variables:
REANA_EMAIL_SMTP_SERVER
REANA_EMAIL_SMTP_PORT
```
https://github.com/reanahub/reana/blob/e59b98288b2d98855ca1ef9bd05d5d3af483b419/helm/reana/templates/reana-server.yaml#L184-L210

We should discuss whether we should not send emails in this case, or whether we should always set the email-related env variables

giuseppe-steduto:
Closes #736.
Closes reanahub/reana-server#618.

### DIFF
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -87,7 +87,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `notifications.email_config.smtp_ssl`                    | Use SSL to connect to SMTP server                                                    | false                                           |
 | `notifications.email_config.smtp_starttls`               | Use STARTTLS command of SMTP to upgrade to an encrypted connection                   | true                                            |
 | `notifications.email_config.smtp_port`                   | SMTP email server port                                                               | None                                            |
-| `notifications.enabled`                                  | Enable REANA system events notifications                                             | false                                           |
+| `notifications.enabled`                                  | Enable REANA system events notifications. For more information, visit the [documentation page](https://docs.reana.io/administration/configuration/configuring-access/) on user sign up. | false |
 | `notifications.system_status`                            | Cronjob pattern representing how often the system status notification should be sent. Leave it empty to deactivate it | "0 0 * * *"    |
 | `reana_hostname`                                         | REANA hostname (e.g. reana.example.org)                                              | None                                            |
 | `namespace_runtime`                                      | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | `.Release.Namespace`                            |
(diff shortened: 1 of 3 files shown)


helm/reana/README.md
```diff
@@ -6,126 +6,126 @@ This Helm automatically prefixes all names using the release name to avoid colli
...
+| `notifications.email_config.smtp_port`                   | SMTP email server port                                                                                                                                                                                                                                                                                                                                                                                                               | None                                            |
+| `notifications.enabled`                                  | Enable REANA system events notifications. Note that if this is set to `false`, users will not receive an email to confirm their email address after signup, so you should either disable email confirmation (`components.reana_server.environment.REANA_USER_EMAIL_CONFIRMATION`) or add/approve them manually.                                                                                                                      | false                                           |
```
mdonadoni, line 90: Instead of adding this explanation, let's link to the docs? What do you think?

helm/reana/README.md
mdonadoni, line 1: Could you revert back to the original file, without reformatting?

helm/reana/templates/reana-server.yaml
```diff
@@ -181,7 +181,11 @@ spec:
                 name: {{ include "reana.prefix" . }}-db-secrets
                 key: password
           {{- end }}
+          - name: REANA_NOTIFICATIONS_ENABLED
+            value: "{{ .Values.notifications.enabled }}"
           {{- if .Values.notifications.enabled }}
+          - name: REANA_NOTIFICATIONS_EMAIL_RECEIVER
+            value: {{ .Values.notifications.email_config.receiver }}
```
mdonadoni, line 188: We should add these two variables also to the system-status cronjob

helm/reana/templates/cronjobs.yaml
```diff
@@ -54,6 +54,10 @@ spec:
                   name: {{ include "reana.prefix" . }}-db-secrets
                   key: password
             {{- end }}
+            - name: REANA_NOTIFICATIONS_ENABLED
+              value: "{{ .Values.notifications.enabled }}"
+            - name: REANA_EMAIL_RECEIVER
+              value: { { .Values.notifications.email_config.receiver } }
```
mdonadoni, line 60:
```suggestion
              value: {{ .Values.notifications.email_config.receiver }}
```

mdonadoni (approved):

PR #743, opened 2023-09-12
cli: add git-upgrade-requirements

### LINKED ISSUE #741: reana-dev: add command to update `requirements.txt`
As part of the release process, the `requirements.txt` file needs to be updated with `pip-compile`.

We can create a new command to do this automatically for all the cluster components.

mdonadoni:
Closes #741

### DIFF
--- a/reana/reana_dev/utils.py
+++ b/reana/reana_dev/utils.py
@@ -32,6 +32,8 @@ from reana.config import (
     HELM_VERSION_FILE,
     JAVASCRIPT_VERSION_FILE,
     OPENAPI_VERSION_FILE,
+    PYTHON_DOCKER_IMAGE,
+    PYTHON_REQUIREMENTS_FILE,
     PYTHON_VERSION_FILE,
     REPO_LIST_ALL,
     REPO_LIST_CLIENT,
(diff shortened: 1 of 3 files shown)


mdonadoni:
What should we do regarding feature branch creation/management? Currently `git-upgrade-shared-modules` and `git-upgrade-requirements` need to be run in a repository where the current branch is a feature branch, and they will fail if the branch is master.

Instead `git-create-release-commit` creates a new branch `release-1.2.3`, but this command is usually executed as last when making a new release: first the shared modules are updated, then all the dependencies in requirements.txt are updated (if needed), and finally the release commit is created.

Should we change `git-upgrade-shared-modules`/`git-upgrade-requirements` to also create a new feature branch? Which name should they use? Or should we create a new `git-create-release-branch` command?

reana/reana_dev/utils.py
```diff
@@ -578,6 +580,42 @@ def update_module_in_cluster_components(
...
+
+    if not pip_compile_cmd:
+        display_message(
+            f"File {PYTHON_REQUIREMENTS_FILE} does not contain a valid pip-compile command.",
+            component,
+        )
+        return False
+
+    if "annotation-style" not in pip_compile_cmd:
+        executable, options = pip_compile_cmd.split(maxsplit=1)
```
giuseppe-steduto, line 607: Just a note: probably quite unlikely, but if `pip_compile_cmd` is only composed by the string `"pip-compile"`, this will raise a `ValueError` because there aren't enough values to unpack. Do you think it's an edge case worth handling?

giuseppe-steduto (commented):
Looks good!

giuseppe-steduto:
> we could maybe group all the release-related commands in an (yet) another git-prepare-release command, that invokes all the others in the correct order?

Sounds good to me too! Do you want to address this here or in another issue?

mdonadoni:
> Sounds good to me! Do you want to address this here or in another issue?

Let's do it in another issue/PR, this one is already quite big!

giuseppe-steduto (approved):

PR #765, opened 2023-12-14
fix(reana-dev): update container image labels when releasing (#765)

tiborsimko:
Changes `reana-dev git-create-release-commit` behaviour to also update the container image label version when releasing cluster components.

### DIFF
--- a/reana/reana_dev/utils.py
+++ b/reana/reana_dev/utils.py
@@ -13,6 +13,7 @@ import functools
 import importlib.util
 import json
 import os
+import re
 import subprocess
 import sys
 from concurrent import futures
@@ -28,6 +29,7 @@ from reana.config import (
     CLUSTER_DEPLOYMENT_MODES,
     COMPONENTS_USING_SHARED_MODULE_COMMONS,
     COMPONENTS_USING_SHARED_MODULE_DB,
+    DOCKER_VERSION_FILE,
     GIT_DEFAULT_BASE_BRANCH,
     HELM_VERSION_FILE,
     JAVASCRIPT_VERSION_FILE,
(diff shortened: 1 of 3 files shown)


reana/reana_dev/utils.py
mdonadoni, line 1: How do we want to update the date in the `Dockerfile`, also considering that we are switching to release-please?

mdonadoni (approved):
Looks good! Just one small discussion point

reana/reana_dev/utils.py
tiborsimko: I guess we'd have to update it ourselves at the day we are releasing and compiling the user-facing summary to be put into CHANGELOG before the regular news items generated by Release Please.

Might be nice to script it further, we would probably need to read `upstream/release-please--branches--master` branch content, enrich it and push back. Perhaps a job for a new command?

mdonadoni: But the user-facing summary would only be for the general `reana` repository, wouldn't it? I would avoid having an user-facing summary for each component, as I think that 1) that would not be very useful to the users in any case 2) it would cancel the benefit of using conventional commits and release-please by introducing a manual step in what could otherwise be a fully automatic release.

Regarding the second point, if we want to script it then I guess we could remove the label from the Dockerfile itself, and then set i from the CLI when building the image (either with reana-dev or with our custom GitHub action). In this way w would not need a new command, and we could still use release-please PRs without having to modify them.

tiborsimko: WRT user-facing summaries, fully agree about doing it only for the user-facing repositories, such as `reana` and perhaps some like `reana-client`, `reana-client-go`.  We may not even need them for every patchlevel release, only for "big" releases (major, and perhaps minor).

WRT updating also date, (a) I could amend `git-create-release-commit`  command to do this too; (b) removing this label from `Dockerfile` could also be interesting at a slight con of not having full labels in the source code repository; (c) BTW we also do need to be aware of updating the date in the Release Please's made pull request, if the release is not happening the same day. We can see IRL about pros/cons?

PR #767, opened 2024-01-10
ci(commitlint): addition of commit message linter

tiborsimko:
Adds commitlint to check the commit message style against agreed conventional commits configuration.

Changes script argument values to always use linter names (e.g. shellcheck).

Changes argument handling to allow only one checking action that can now accept further optional arguments.


mdonadoni (approved):
Two comments:
- The commits refer to the wrong PR (676 instead of 767)
- Should we investigate how to update the Helm Chart now or we do it as part of another PR?

tiborsimko:
> * Should we investigate how to update the Helm Chart now or we do it as part of another PR?

We can do this later, together with updating the OpenAPI versioning, the alpha-beta decision, etc.

PR #774, opened 2024-02-05
feat(helm): add value to customise PostgreSQL docker image (#774)

### LINKED ISSUE #744: helm: upgrade to PostgreSQL 14
We are currently using PostgreSQL 12 in our Helm chart:
https://github.com/reanahub/reana/blob/9cb65c8302290e699c4e23273b54bab1a4bcf376/helm/reana/templates/reana-db.yaml#L34

Given that we have already tested (and we are already using) PostgreSQL 14,  we should update the internal database to the same version, so that local/dev environments use the same database version as REANA deployed at CERN.

Note that this represents a big issue for already-deployed instances of REANA that use the internal database, as a data migration is needed to move from PostgreSQL 12 to PostgreSQL 14. See the related docs PR https://github.com/reanahub/docs.reana.io/pull/170.

We should discuss what to do, as we have a few options:
- Update to PostgreSQL 14 without making it possible to customise the version with Helm values, as we have done up until now. This is good to keep all deployments up to date and using the same version, but it means that REANA admins are forced to migrate their database when upgrading to a new version of REANA. This means we should also advertise in some way if a database migration is needed when releasing a new version of REANA.
- Make the version of PostgreSQL customisable via Helm values, making PostgreSQL 14 as default. This means that admins can keep their old version of PostgreSQL without having to upgrade, but this also means that they might encounter incompatibilities if they keep the old version for too long. It might also happen that admins choose a newer version of PostgreSQL, which might also cause troubles.

### LINKED ISSUE #773: helm: add value to customise PostgreSQL docker image
Currently the PostgreSQL Docker image used by REANA is hardcoded in the Helm chart. We should add a new value to the chart to make it customisable.

This is the first step towards #744

mdonadoni:
- feat(helm): add value to customise PostgreSQL docker image (#774)
- feat(helm): use PostgreSQL 14.10 in local dev deployment (#774)

Closes #773 
Closes #744

### DIFF
--- a/helm/configurations/values-dev.yaml
+++ b/helm/configurations/values-dev.yaml
@@ -1,6 +1,8 @@
 # REANA components pointing to `latest`, locally built master branch
 
 components:
+    reana_db:
+      image: docker.io/library/postgres:14.10
     reana_server:
       image: docker.io/reanahub/reana-server
       environment:
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -9,6 +9,7 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | Parameter                                                | Description                                                                          | Default value                                   |
 |----------------------------------------------------------|--------------------------------------------------------------------------------------|-------------------------------------------------|
 | `components.reana_db.enabled`                            | Instantiate a PostgreSQL database inside the cluster                                 | true                                            |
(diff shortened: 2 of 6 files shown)


mdonadoni:
Note: PRs to update the version of PostgreSQL to 14.10 in the other repos are still missing

giuseppe-steduto (approved):

PR #777, opened 2024-02-08
fix(reana-dev): create commits that conform to conventional style (#777)

mdonadoni:


reana/reana_dev/git.py
```diff
@@ -1365,7 +1371,8 @@ def git_upgrade_requirements(ctx, component, exclude_components):  # noqa: D301
         if upgrade_requirements(component):
             run_command(f"git add {PYTHON_REQUIREMENTS_FILE}", component)
             run_command(
-                'git commit -m "installation: bump all dependencies"', component
+                f'git commit -m "build(deps): bump all dependencies{get_commit_pr_suffix(component)}"',
```
tiborsimko, line 1374: `build(python)` perhaps, so that we restrict the meaning of "all" here? (alternatively, we could use "bump all Python packages")

reana/reana_dev/git.py
```diff
@@ -1280,7 +1286,7 @@ def git_upgrade_shared_modules(
 
     def _create_commit_or_amend(components):
         for c in components:
-            commit_cmd = 'git commit -m "installation: bump shared modules"'
+            commit_cmd = f'git commit -m "build(deps): bump shared modules{get_commit_pr_suffix(component)}"'
```
tiborsimko, line 1289: `build(python)` perhaps, by analogy with `build(docker)` and friends?

reana/reana_dev/git.py
```diff
@@ -199,8 +201,11 @@ def git_create_release_commit(
     if modified_files:
         run_command(f"git add {' '.join(modified_files)}", component)
 
+    commit_msg = (
+        f"chore({base}): release {next_version}{get_commit_pr_suffix(component)}"
```
tiborsimko, line 205: BTW one meta comment related to the release news.

Currently this PR would advertise:

```
fix(reana-dev): create commits that conform to conventional style
```

This does not expose the `reana-dev` sub commands that this change improves.

If we would like to include those (for analogy with #765), perhaps the commit should be split in two, advertising something like:

```
fix(reana-dev): `git-create-release-commit` use conventional commit style
fix(reana-dev): `git-upgrade-shared-modules` use conventional commit style
```

But we are probably hitting the line length limits here...

mdonadoni: From my point of view this was a "general" change to the whole reana-dev to conform to the conventional commit style we have agreed on. Not sure how to also include the name of the command in the commit title while also staying under the character limit, unfortunately.

tiborsimko (approved):

PR #778, opened 2024-02-09
feat(helm): add value to customise URL of privacy notice (#778)

mdonadoni:
Closes reanahub/reana-ui#392

### DIFF
--- a/helm/reana/templates/reana-config.yaml
+++ b/helm/reana/templates/reana-config.yaml
@@ -1,4 +1,7 @@
 ---
+{{- if .Values.components.reana_ui.cern_ropo }}
+  {{- fail  "`components.reana_ui.cern_ropo` is deprecated, please use `components.reana_ui.privacy_notice_url` instead." }}
+{{- end }}
 apiVersion: v1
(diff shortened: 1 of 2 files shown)


helm/reana/templates/reana-config.yaml
```diff
@@ -12,9 +15,9 @@ data:
     client_pyvenv: {{ .Values.components.reana_ui.client_pyvenv | quote | default "null" }}
     forum_url: {{ .Values.components.reana_ui.forum_url | quote | default "null" }}
```
giuseppe-steduto, line 16: If we update the default value for the forum and docs URLs in the README, shouldn't we update them here as well?

mdonadoni: They are already set in the values file, I am just keeping the docs up to date: 
https://github.com/reanahub/reana/blob/79aedb9ef2ba0c8a933fe2cac334bd51cd32dd85/helm/reana/values.yaml#L76-L77

giuseppe-steduto (approved):

PR #781, opened 2024-02-22
feat(helm): add value to customise env vars of job controller and workflow engines (#781)

mdonadoni:
- feat(helm): add values to customise env vars of workflow engines (#781)
- feat(helm): add value to customise env vars of job controller (#781)


giuseppe-steduto (approved):
LGTM!

PR #783, opened 2024-02-28
fix(reana-dev): add PR number in commit to bump shared modules (#783)

mdonadoni:


tiborsimko (approved):
Works nicely.

A suggestion for the commit headline:

```
fix(reana-dev): add PR number in commits bumping shared modules (#783)
```

(otherwise people could read "to bump..." as a reason for the change)

mdonadoni:
I have also improved reana-dev to detect whether a PR was already open or not

PR #789, opened 2024-03-07
feat(reana-dev): add `git-aggregate-changelog` (#789)

mdonadoni:


reana/reana_dev/git.py
```diff
@@ -1586,4 +1590,201 @@ def git_tag(component, exclude_components):  # noqa: D301
...
+        changelog_lines[:idx_insert] + new_lines + changelog_lines[idx_insert:]
+    )
+
+    with open(changelog_path, "w") as changelog_file:
+        changelog_file.writelines(new_changelog)
+
+
+@git_commands.command(name="git-aggregate-changelog")
+@click.option(
+    "--previous-reana-client",
```
tiborsimko, line 1725: We could enrich the CLI later so that users could pass `-c` as to the other `reana-dev` commands. This might be useful to generate quickly an overview of all the changes from version X to version Y for example.

Such a CLI interface could look something like:

```
-c r-client -- since-version 0.9.2
-c r-client --since-date=2023-01-01
```

Or, by analogy with `git-checkout-pr`, we should probably have a tuple of values to allow assembling several components:

```
-b r-client 0.9.1 -b r-j-controller 0.9.0 
```

Just some wild musings for later, not for now.  

That said, if we do decide to go for this one day, to allow aggregating news from more than one release, then the CLI option name could change, so perhaps we could forward-name the option to be something like `--since-reana-client-version 0.9.2` instead of `--previous-reana-client`? :thinking:

reana/reana_dev/git.py
```diff
@@ -1586,4 +1590,201 @@ def git_tag(component, exclude_components):  # noqa: D301
...
+    help="Which is the version of reana-client that was released "
+    "for the last REANA release?",
+    required=True,
+)
+def get_aggregate_changelog(previous_reana_client):  # noqa: D301
```
tiborsimko, line 1730: Works nicely :+1: 

One functionality comment that would be good to do already now  is to present the aggregation of components following their alphabetical naming order, so that the order is predictible. 

(Currently I'm seeing r-w-e-yadage being printed before r-w-e-snakemake for example.)

tiborsimko (approved):

Issue #784, opened 2024-02-29
build: drop Python 3.6 and 3.7


mdonadoni:
Drop Python 3.6 and 3.7 in the next major release, as they are both EOL.

Clean up `setup.py` in all the cluster components to remove workarounds that are only needed for Python<3.8.

PR #793, opened 2024-03-26
build(python): drop support for Python 3.6 and 3.7 (#793)

mdonadoni:
BREAKING CHANGE: drop support for Python 3.6 and 3.7

Closes reanahub/reana#784

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 #
 # This file is part of REANA.
-# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023 CERN.
+# Copyright (C) 2017, 2018, 2019, 2020, 2021, 2022, 2023, 2024 CERN.
 #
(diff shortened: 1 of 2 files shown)


setup.py
mdonadoni, line 1: Many components do not set the `python_requires` field in `setup.py`. Should we add it everywhere, even to cluster components that are not published to PyPI?

mdonadoni: Also, what about the classifiers `Programming Language :: Python :: 3.xx`? Should we put 3.8...3.12 in all the components, or only in the ones published to PyPI?

mdonadoni:
Many components do not set the `python_requires` field in `setup.py`. Should we add it everywhere, even to cluster components that are not published to PyPI?

Also see related comment https://github.com/reanahub/reana-db/pull/235#discussion_r1538929806

PR #795, opened 2024-04-16
feat(helm): add value to customise interactive session images (#795)

mdonadoni:
Closes reanahub/reana-workflow-controller#569


helm/reana/values.yaml
```diff
@@ -22,7 +22,12 @@ workspaces:
 
 interactive_sessions:
   cronjob_schedule: "0 3 * * *"  # everyday at 3am
-  maximum_inactivity_period: forever
+  environments:
+    jupyter:
+      recommended:
+        - image: "docker.io/jupyter/scipy-notebook:notebook-6.4.5"
```
tiborsimko, line 28: It may be perhaps good to have also an alias (or a list of aliases) for an image, because currently we would allow opening of `docker.io/jupyter/scipy-notebook:notebook-6.4.5` but not the opening of `jupyter/scipy-notebook:notebook-6.4.5` which is essentially the same image.  Motivation: people often leave out `docker.io` from the fully-qualified image names.

This could be an issue during the use case of opening a session from the `reana-client`, such as:

```console
$ rc open -w root6-roofit-snakemake-kubernetes -i jupyter/scipy-notebook:notebook-6.4.5
==> ERROR: Interactive session could not be opened:
Custom container image jupyter/scipy-notebook:notebook-6.4.5 is not allowed.
```

Since the `docker.io` prefix is kind of unique, due to Docker history, there may not be many more examplees calling for image aliases out there... So perhaps the best solution is not to alter the YAML configuration file with aliases after all, but rather to solve this command-line use case on the application level, e.g. try to add  leading `docker.io/` if the image name supplied by the user does not start with it, when trying to match the user-supplied value against the list of available images.

mdonadoni: Done in reana-workflow-controller, the alias prefixes I have considered are:
- `docker.io/`
- `docker.io/library/`
- `library/`

mdonadoni: This means that `ubuntu:24.04`, `library/ubuntu:24.04`, `docker.io/library/ubuntu:24.04` are all going to work well

tiborsimko (approved):

PR #804, opened 2024-06-18
feat(helm): allow cluster administrator to configure ingress host (#804)

Alputer:
Closes reanahub/reana-workflow-controller#587

### DIFF
--- a/helm/reana/templates/reana-workflow-controller.yaml
+++ b/helm/reana/templates/reana-workflow-controller.yaml
@@ -86,6 +86,8 @@ spec:
             value: {{ .Values.namespace_runtime | default .Release.Namespace }}
           - name: REANA_INGRESS_ANNOTATIONS
             value: {{ .Values.ingress.annotations | toJson | quote }}
+          - name: REANA_INGRESS_HOST
+            value: {{ .Values.reana_hostname | default "" | quote  }}
           - name: REANA_DEFAULT_QUOTA_CPU_LIMIT
             value: {{ .Values.quota.default_cpu_limit | default 0 | quote }}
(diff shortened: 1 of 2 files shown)


helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -91,6 +91,8 @@ spec:
           - name: REANA_DEFAULT_QUOTA_DISK_LIMIT
             value: {{ .Values.quota.default_disk_limit | default 0 | quote }}
           {{- if .Values.quota.enabled }}
+          - name: REANA_INGRESS_HOST
```
tiborsimko, line 94: The `REANA_INGRESS_HOST` should not be located inside the "if quota..." clause, because it does not depend on quotas. Please move it above, next to where `REANA_INGRESS_ANNOTATIONS` lives.

helm/reana/values.yaml
```diff
@@ -135,6 +135,7 @@ notifications:
 # Accessing the cluster from outside world
 ingress:
   enabled: true
+  host: ""
```
tiborsimko, line 138: Hmm, do we need to add a new field `ingress.host` for this functionality? I'm wondering because we already have `reana_hostname` field defined elsewhere, so if people set it up, we could perhaps reuse that directly.   (And if it is not set up, then use the empty string as default value.)

(Another possibility would be to hardcode `""` as default value inside `helm/reana/templates/reana-workflow-controller.yaml`, without even playing with `reana_hostname`. )

In this way we don't need to introduce the new `ingress.host` field.... Thinking of making changes to 0.9.3 as minimal as possible.

tiborsimko (approved):
Works nicely locally and on DEV :+1:

Issue #654, opened 2022-07-06
Helm install of reana cluster locally fails due to kubeVersion incompatibility (required < 1.22.0-0)


egazzarr:
Following the documentation (https://docs.reana.io/administration/deployment/deploying-locally/) to install reana from helm chart on my personal laptop. Locally I have the following kubectl versions:

Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-02-16T12:38:05Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-19T15:39:43Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}

When running:
$ helm install reana reanahub/reana --namespace reana --create-namespace --wait

I get the error: 
Error: INSTALLATION FAILED: chart requires kubeVersion: >= 1.13.0-0 < 1.22.0-0 which is incompatible with Kubernetes v1.24.0

tiborsimko:
Hi @egazzarr, if you have latest Kind 0.14 installed locally, then it defaults to Kubernetes v1.24 which REANA supports only in the latest bleeding-edge development sources. Hence this compatibility trouble.

A quick fix is to bring up an older Kubernetes version, such as v1.21, using:

```console
$ kind delete cluster
$  kind create cluster --config kind-localhost-30443.yaml --image=kindest/node:v1.21.12
```

Please see https://github.com/reanahub/reana/issues/647

We'll update the documentation in this respect.

PR #811, opened 2024-07-26
feat(reana_dev): customize kind image version (#811)

jlemesh:
Similar [issue](https://github.com/reanahub/reana/issues/654) happened to me using `reana-dev` tool with fresh `kind` install. `kind` pulls the latest image version, but the version does not match `kubeVersion` in Helm chart, which causes error on Helm chart deployment.

Though Helm chart is now upgraded and accepts newest image version, at some point in time discrepancy will happen again.  `kindest/node` image version can just be pinned (should be updated together with Helm chart) and can be allowed to customize as a CLI option.

### DIFF
--- a/reana/reana_dev/cluster.py
+++ b/reana/reana_dev/cluster.py
@@ -71,8 +71,14 @@ def cluster_commands():
     is_flag=True,
     help="Disable default CNI and use e.g. Calico.",
 )
+@click.option(
+    "--kind-node-version",
+    help="Which kindest/node image version to use?",
+)
 @cluster_commands.command(name="cluster-create")
-def cluster_create(mounts, mode, worker_nodes, disable_default_cni):  # noqa: D301
+def cluster_create(
+    mounts, mode, worker_nodes, disable_default_cni, kind_node_version
+):  # noqa: D301
     """Create new REANA cluster.
 
     \b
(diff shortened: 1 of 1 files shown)


reana/config.py
```diff
@@ -266,6 +266,9 @@
 TIMEOUT = 900
 """Maximum timeout to wait for results when running demo analyses in CI."""
 
+KIND_IMAGE_VERSION = "v1.30.2"
+"""Kind Docker image version. Should be compatible with kubeVersion in HELM_VERSION_FILE."""
```
mdonadoni, line 270: nitpicking:
```suggestion
"""Kind node Docker image version. Should be compatible with kubeVersion in HELM_VERSION_FILE."""
```

PS: we will not need this default value if we decide to make the flag optional, see other comment

reana/reana_dev/cluster.py
```diff
@@ -71,8 +75,15 @@ def cluster_commands():
     is_flag=True,
     help="Disable default CNI and use e.g. Calico.",
 )
+@click.option(
+    "--kind-version",
```
mdonadoni, line 79:
```suggestion
    "--kind-node-version",
```
To make it clear that it's not the version of kind itself, but it's the one used for the node image?

reana/reana_dev/cluster.py
```diff
@@ -71,8 +75,15 @@ def cluster_commands():
...
+    default=KIND_IMAGE_VERSION,
```
mdonadoni, line 80: Instead of providing a default, let's make this flag optional? If not provided, kind will decide on its own which version to use.

I'm saying this in particular because on kind releases page they even recommend to specify the hash of the images, as different images might not be compatible with all versions of kind ([source](https://github.com/kubernetes-sigs/kind/releases)):
_NOTE: You must use the sha256 digest to guarantee an image built for this release, until such a time as we switch to a different tagging scheme. Even then we will highly encourage digest pinning for security and reproducibility reasons._

So having a single default image might not work on every `kind` version.

What do you think?

jlemesh: Agree, will make it optional.

mdonadoni (approved):
Nice job, looks good! Could you squash all the commits into a single one? Then I will merge :D

mdonadoni:
Rebased and merged. I have also slighly changed the commit message: `reana_dev` -> `reana-dev`.

PR #817, opened 2024-08-05
feat(reana-dev): change run-example to use gherkin-parser module (#817)

ajclyall:
Run-example now uses the test files in the specification file to check the correctness of the results (by running the test command).

### DIFF
--- a/reana/reana_dev/run.py
+++ b/reana/reana_dev/run.py
@@ -9,6 +9,7 @@
 """`reana-dev`'s run commands."""
 
 import os
+import re
 import subprocess
 import sys
 import time
@@ -18,9 +19,7 @@ import click
 
 from reana.config import (
     COMPUTE_BACKEND_LIST_ALL,
-    EXAMPLE_LOG_MESSAGES,
     EXAMPLE_NON_STANDARD_REANA_YAML_FILENAME,
-    EXAMPLE_OUTPUT_FILENAMES,
     TIMECHECK,
     TIMEOUT,
     WORKFLOW_ENGINE_LIST_ALL,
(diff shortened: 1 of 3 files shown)


reana/reana_dev/run.py
```diff
@@ -604,18 +555,21 @@ def _return_missing_output_files(component: str, workflow_name: str) -> List[str
...
+                                f"reana-client test -w {workflow_name}",
+                                component,
+                                return_output=True,
+                            )
+
+                            if re.findall(r".*failed$", tests, re.MULTILINE):
+                                run_statistics["failed"].append(workflow_name)
+                            else:
+                                run_statistics["passed"].append(workflow_name)
+                        except SystemExit:
```
mdonadoni, line 569: Mmmh catching SystemExit is (probably) not best practice, but improving this would require refactoring `run_command` which is used everywhere, so let's keep it like this for now.

However, I think that if the command fails, then we should put the workflow in the list of the failed ones?

mdonadoni:
PS: this PR needs to be rebased

reana/reana_dev/run.py
mdonadoni: Otherwise this can happen:
```
Test matrix:
  - 1 demo example(s): reana-demo-root6-roofit
  - 4 workflow engine(s): cwl, serial, snakemake, yadage
  - 1 compute backend(s): kubernetes

Test results:
  - 0 submitted
  - 0 running
  - 0 pending
  - 0 queued
  - 0 passed
  - 0 failed

OK
```
Four workflows were run, but the final statistics all report 0 workflows

mdonadoni (commented):
Looks good to me!

PS: Please add yourself to `AUTHORS.md`! ;)

mdonadoni (approved):

PR #818, opened 2024-08-06
feat(helm): add support for pgbouncer (#818)

mdonadoni:
Closes reanahub/reana#819

### DIFF
--- a/helm/reana/README.md
+++ b/helm/reana/README.md
@@ -98,6 +98,12 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `reana_hostname`                                         | REANA hostname (e.g. reana.example.org)                                              | None                                            |
 | `namespace_runtime`                                      | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | `.Release.Namespace`                            |
 | `naming_scheme`                                          | REANA component naming scheme                                                        | None                                            |
+| `pgbouncer.enabled`                                      | Instantiate PgBouncer inside the cluster to pool database connections                | false                                           |
+| `pgbouncer.image`                                        | [PgBouncer image](https://hub.docker.com/r/bitnami/pgbouncer/) to use                | `bitnami/pgbouncer:1.23.1`                      |
+| `pgbouncer.pool_mode`                                    | Pool mode to use (session, transaction, statement)                                   | transaction                                     |
+| `pgbouncer.max_client_conn`                              | Maximum number of client connections allowed                                         | 1000                                            |
+| `pgbouncer.max_db_connections`                           | Maximum number of server connections allowed                                         | 100                                             |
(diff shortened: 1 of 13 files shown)


helm/reana/README.md
```diff
@@ -98,6 +98,12 @@ This Helm automatically prefixes all names using the release name to avoid colli
 | `reana_hostname`                                         | REANA hostname (e.g. reana.example.org)                                              | None                                            |
 | `namespace_runtime`                                      | Namespace in which the REANA runtime pods (workflow engines, jobs etc...) will run   | `.Release.Namespace`                            |
 | `naming_scheme`                                          | REANA component naming scheme                                                        | None                                            |
+| `pgbouncer.enabled`                                      | Instantiate PgBouncer inside the cluster to pool database connections                | false                                           |
```
tiborsimko, line 101: When `pgbouncer` is enabled, shall we act on `reana-job-controller`'s `
REANA_DB_CLOSE_POOL_CONNECTIONS` configuration value?

helm/reana/templates/database-config.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
```
tiborsimko, line 1: Hmm, I wonder whether it would be possible to amend the setup to automatically delete / refresh `reana-server` and `reana-workflow-controller` pods when the admin changes PgBouncer from enabled to disabled and vice versa? (Currently in addition to `helm diff upgrade...` one has to do it manually.) 

This comment is a sort of a "bonus", the things work nicely :+1: and I'm not sure whether admins would attempt to play with enabling/disabling PgBouncer dynamically in reality... Well we could always simply explain this in the documentation.

helm/reana/README.md
mdonadoni: No, it's actually the opposite, if pgbouncer is not enabled then the two options are:
- increase the number of connection slots of the database
- close the connections as soon as they are returned to SQLAlchemy's connection pool
I have kept the default value of `REANA_DB_CLOSE_POOL_CONNECTIONS` as to have the best performance possible, but it is customisable so that we can revert back the change in case of troubles with PgBouncer. 

What do you think we should do regarding this? Shall we add this to the documentation?

helm/reana/templates/database-config.yaml
mdonadoni: I have updated the PR to also roll the deployments of pgbouncer/r-server/r-w-controller when the database config (configmap or secrets) change. This can be used as a blueprint to do the same for the other config/secret that might change over time.

helm/reana/README.md
tiborsimko: Yes, we can later improve the scalability-related documentation regarding PgBouncer.

tiborsimko (approved):
Works nicely :+1: Including deployment rollout in case of DB config chnges

Issue #824, opened 2024-08-21
REANA job logs collection with FluentBit DaemonSet and OpenSearch


jlemesh:
Update REANA Helm chart to add FluentBit DaemonSet and OpenSearch for job logs collection.

jlemesh:
Production security configuration for OpenSearch: https://opensearch.org/docs/latest/security/configuration/index/

jlemesh:
Related code changes: https://github.com/reanahub/reana-workflow-controller/pull/602 https://github.com/reanahub/reana-job-controller/pull/468

PR #827, opened 2024-08-26
feat(helm): add live logs deployment (#827)

jlemesh:
Closes https://github.com/reanahub/reana/issues/824

## OpenSearch

Optionally deploy OpenSearch alongside REANA (`enabled: false` by default).

In dev environment, deploy one instance with security features off (i. e. no TLS, no authn/authz)

In prod, also only one OpenSearch node is configured. To add more nodes, each of these nodes needs to have its own `PersistentVolumeClaim` and `PersistentVolume` as each node writes to a directory with the same name, but different contents - this will not work with current `reana-shared-persistent-volume` or `reana-infrastructure-persistent-volume`. It is possible to  use `volumeClaimTemplates` with the `StorageClass` of choice (configured in OpenSearch Helm chart `persistence.storageClass`).

Volumes configuration tested with [nfs-ganesha-server-and-external-provisioner/nfs-server-provisioner](https://artifacthub.io/packages/helm/nfs-ganesha-server-and-external-provisioner/nfs-server-provisioner), for one and multiple nodes.

OpenSearch allows deploying nodes of different types, e. g. `master` and `data`. This configuration would not allow to do it, we need to add the second `opensearch` dependency, e. g. `opensearch-master`, and use the same Helm chart with different configuration. I think it is an overkill at this stage, as this is needed for bigger clusters.

TLS certificates for OpenSearch nodes are generated automatically by a Helm function, similar to how it is done for Ingress, but without automatic rotation on each `helm upgrade`. It is possible for a user to supply his own certificates, by putting them in a secret and mounting to an OpenSearch pod (in `opensearch.secretMounts`). Admin TLS certificates provide superadmin permissions to whoever uses them, and are stored in the same secret - there is no specific admin user.

Two users are configured - `reana` and `fluentbit`. Their passwords need to be prepared by first spinning up environment with Opensearch and running `hash.sh` script (described in [docs](https://opensearch.org/docs/latest/security/configuration/yaml/)):

```
./plugins/opensearch-security/tools/hash.sh -p somepassword
```

 After getting the hashes for the passwords, they should be supplied to Helm with `--set opensearch.customSecurityConfig.internalUsers.reana.hash='$So$Me$pASsWOrD.HasH' --set opensearch.customSecurityConfig.internalUsers.fluentbit.hash='$So$Me$pASsWOrD.HasH'`.

Also enabled SSL cert reload, which allows updating SSL certs without restarting OpenSearch instance. Link to docs provided in the comments for those who will want to use it.

## FluentBit

Optionally deploy FluentBit alongside REANA (`enabled: false` by default). Collects job and workflow logs with `tail` input plugin and pushes logs to OpenSearch. Allows configuring custom TLS certificates for OpenSearch connection.

## `reana-dev`

The command breaks Helm configuration if it contains long (multiline) strings. PyYAML library first reads the YAML config file, converts multiline strings to one long string delimited with `"` and then wraps it while `dump`ing, i. e.:

```
<...>
key: "value\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\n"\
"value\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\n"\
"value\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\nvalue\n"
<...>
```

If the string is used with Helm template values (`tpl` function), it adds whitespaces in places where `\` was used, which results in malformed configuration for OpenSearch and FluentBit.

`dump(width=100000)` will add `\` only if the string is longer than 100000 characters.

It also requires user to escape certain characters in YAML values like `$` and `\`. PyYAML converts `|` strings to `""` and Helm command is supplied with bad values.

To avoid errors when loading/dumping `values.yaml` file, skip its loading altogether as there is no need to load it explicitly as it is always loaded by Helm itself. In case `releasehelm` mode is used for `reana-dev cluster-deploy`, just use an empty dict for values and add overrides into it, `values.yaml` will be loaded by default by Helm.

## How to test

Checkout https://github.com/reanahub/reana-workflow-controller/pull/602 and https://github.com/reanahub/reana-job-controller/pull/468 to retrieve live logs in REANA components. Checkout and install https://github.com/reanahub/reana-client/pull/731 to watch live logs via CLI.

### Dev setup

Deploy REANA as usual and run:

```
reana-dev run-example -c r-d-helloworld -w serial --submit-only
# forward opensearch port
curl -XGET "http://localhost:9200/fluentbit-job_log/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "match": {
      "kubernetes.labels.job-name.keyword": "<reana run job ID>"
    }
  },
  "sort": [
    {
      "@timestamp": {
        "order": "desc"
      }
    }
  ]
}' | jq
reana-client logs --workflow helloworld-serial-kubernetes
reana-client logs --workflow helloworld-serial-kubernetes --follow --filter step=hello1
```

### Prod setup

Deploy REANA as usual then run:

```
helm upgrade reana reana/helm/reana --wait --debug --force \
  --set components.reana_workflow_controller.image=docker.io/reanahub/reana-workflow-controller \
  --set opensearch.customSecurityConfig.internalUsers.reana.hash='$2y$12$e4RWvWxRW0QO8QtnDd.70eV/fJDJ2F7uUo2ox1P07bIpRArYDSUWK' \
  --set opensearch.customSecurityConfig.internalUsers.fluentbit.hash='$2y$12$e4RWvWxRW0QO8QtnDd.70eV/fJDJ2F7uUo2ox1P07bIpRArYDSUWK' \
  --set fluent-bit.outputConfig.httpPasswd='ReanaOS1=' \
  --set opensearch.enabled=true \
  --set fluent-bit.enabled=true \
  --set components.reana_workflow_controller.environment.REANA_OPENSEARCH_ENABLED=true \
  --set components.reana_workflow_controller.environment.REANA_OPENSEARCH_PASSWORD='ReanaOS1=' \
  --set components.reana_db.image=docker.io/library/postgres:14.10
reana-dev run-example -c r-d-helloworld -w serial --submit-only
# forward opensearch port
curl -u "reana:reana" --insecure -XGET "https://localhost:9200/fluentbit-job_log/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
    "match": {
      "kubernetes.labels.job-name.keyword": "<reana run job ID>"
    }
  },
  "sort": [
    {
      "@timestamp": {
        "order": "desc"
      }
    }
  ]
}' | jq
reana-client logs --workflow helloworld-serial-kubernetes
reana-client logs --workflow helloworld-serial-kubernetes --follow --filter step=hello1
```


helm/configurations/values-dev.yaml
```diff
@@ -31,3 +33,36 @@ components:
...
+    generate: false
+  singleNode: true
+  config:
+    opensearch.yml: |
+      cluster.name: opensearch-cluster
+      network.host: 0.0.0.0
+      plugins.security.disabled: true
+  extraEnvs:
+    - name: OPENSEARCH_INITIAL_ADMIN_PASSWORD
+      value: ReanaOS1=
```
tiborsimko, line 49: It would be better not to hard-code any password values in `values-dev.yaml`.

See also how we handle e.g. cache password in the 0.9.4:

```console
$ rg -C 3 REANA_CACHE_PASSWORD
• helm/reana/templates/reana-cache.yaml
49-      - name: cache
50-        image: docker.io/library/redis:5.0.5
51-        env:
52:          - name: REANA_CACHE_PASSWORD
53-            valueFrom:
54-              secretKeyRef:
55-                name: {{ include "reana.prefix" . }}-cache-secrets
56-                key: password
57-        args:
58-          - "--requirepass"
59:          - "$(REANA_CACHE_PASSWORD)"
60-        ports:
61-        - containerPort: 6379
62-      {{- if .Values.node_label_infrastructure }}

• helm/reana/templates/reana-server.yaml
151-              secretKeyRef:
152-                name: {{ include "reana.prefix" . }}-secrets
153-                key: REANA_SECRET_KEY
154:          - name: REANA_CACHE_PASSWORD
155-            valueFrom:
156-              secretKeyRef:
157-                name: {{ include "reana.prefix" . }}-cache-secrets
```

helm/reana/templates/priority-classes.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if index .Values "fluent-bit" "enabled" }}
+---
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: {{ include "reana.prefix" . }}-fluent-bit-priority-class
+value: 1000000
```
tiborsimko, line 7: Can we keep the value 1000000 hard-coded, or do we need to make it more easily configurable by admins at the deployment time?

reana/reana_dev/cluster.py
```diff
@@ -314,6 +315,7 @@ def cluster_deploy(
     admin_email,
     admin_password,
     instance_name,
+    live_logs,
```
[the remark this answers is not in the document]
jlemesh: Hm, I thought `values-dev.yaml` is a defaults file that is not meant to be modified (as the changes in this file are tracked by git). Also developers need to know to make three specific changes to enable live logs (enable opensearch, enable fluentbit, and enable opensearch fetcher env var in r-workflow-controller), I then need to add this to some development documentation, not sure where exactly, or maybe just adding comments near these values in `values-dev.yaml` would do?

helm/reana/templates/priority-classes.yaml
jlemesh: Hm, maybe still worth making it configurable and add 1000000 as a default, though I doubt anyone would ever need to change this.

helm/reana/values.yaml
```diff
@@ -189,3 +194,354 @@ quota:
...
+# OpenSearch chart values.yaml
+opensearch:
+  enabled: false
+  tls:
+    generate: true
+    ca:
+      cn: "reana.io"
+      ttl: 365
+    cert:
+      cn: "opensearch-cluster-master.default.svc.cluster.local"
```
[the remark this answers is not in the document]
jlemesh: `-master` or other postfix is mandatory. So we can have `reana-opensearch-*`, I suggest leaving `-master` as it is a default.

tiborsimko (approved):
Works nicely :+1: 

I have squashed the two OpenSearch commits, slightly rephrased the `reana-dev` PyPAML commit, and switched off OpenSearch/FluentBit for now in `reana-dev` until we put a fix for the `/var/reana/opensearch` directory ownership (just in case anyone tries master).

Issue #798, opened 2024-05-10
helm: fix linting errors


tiborsimko:
**Current behaviour**

`helm lint` currently reports an error for our chart:

```
$ helm lint helm/reana -f helm/configurations/values-dev.yaml
engine.go:189: [INFO] Fail:

 ERROR: You must have at least networking.k8s.io/v1beta1 to use ingressClass
==> Linting helm/reana

1 chart(s) linted, 0 chart(s) failed
```

**Expected behaviour**

Our charts should pass linting error-free.

**Bonus**

Introduce chart linting to the `run-tests.sh` and the CI process. 

See also https://github.com/helm/chart-testing

mdonadoni:
The error is coming from Traefik's subchart ([source](https://github.com/traefik/traefik-helm-chart/blob/8a7f7833ce14ece56e914b1b7236a8208ffa6869/traefik/templates/ingressclass.yaml#L9)). Disabling the subchart makes the linter pass without warnings or errors:

```console
$ helm lint helm/reana -f helm/configurations/values-dev.yaml --set traefik.enabled=false
==> Linting helm/reana

1 chart(s) linted, 0 chart(s) failed
```

Alputer:
Handled by https://github.com/reanahub/reana/pull/832

PR #832, opened 2024-09-10
feat(helm): upgrade traefik (#832)

Alputer:
I have realized that our `traefik` dependency is from 3 years ago. Updating the treaefik dependency resolved the linting error. I tested it with 5 different workflows and it seems to be working fine.

closes (#798)

### DIFF
--- a/helm/reana/Chart.yaml
+++ b/helm/reana/Chart.yaml
@@ -30,7 +30,7 @@ version: 0.9.3
 kubeVersion: '>= 1.21.0-0 < 1.32.0-0'
 dependencies:
   - name: traefik
-    version: 10.6.2
+    version: 31.1.0
     repository: https://helm.traefik.io/traefik
     condition: traefik.enabled
     tags:


mdonadoni:
Works well!

By moving from the Traefik helm chart v10.6.2 to v31.0.0 we are also moving from Traefik v2.5.4 ([source](https://github.com/traefik/traefik-helm-chart/blob/8a7f7833ce14ece56e914b1b7236a8208ffa6869/traefik/Chart.yaml#L6)) to v3.1.2 ([source](https://github.com/traefik/traefik-helm-chart/blob/f14ba440aa55b806ba024be27db7d29d91df7fa9/traefik/Chart.yaml#L7)). 

I have checked the [migration guide](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3/) and the [changes](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3-details) and it seems like we are not affected by any of it, even more so because we are using k8s `Ingress`es and not traefik's `IngressRoute`s. The only change that might have affected us is the change of behaviour of `PathPrefix`, which does not support regexes anymore ([source](https://doc.traefik.io/traefik/v3.0/migration/v2-to-v3-details/#new-v3-syntax-notable-changes)); we are not using regexs in router rules, so we are good.

Just two small comments:
- Could you rebase the PR?
- Could you chage the commit message to highlight that we are moving to traefik v3? Something like `build(helm): upgrade Traefik to v3 (#832)`

Alputer:
I rebased and updated the commit message. I also tested the version `31.1.0` and it seems to be working fine too. So updated the version from `31.0.0` to `31.1.0`.

mdonadoni (approved):

Issue #836, opened 2024-10-09
ci: introduce `yamllint` checks


tiborsimko:
It would be useful to enrich our CI check suite in order to validate YAML files.

We are using YAML in Helm charts, in REANA demo examples, but also in all the repositories with respect to commitlint.

In our deployment repository (using GitLab), we do:

(1) In GitLab CI:

```yaml
lint-yamllint:
  image: registry.gitlab.com/pipeline-components/yamllint:latest
  script:
    - ./run-tests.sh --check-yamllint
```

(2) In `run-tests.sh`:

```bash
...

check_yamllint () {
    yamllint .
}

check_all () {
    check_black
    check_flake8
    check_pydocstyle
    check_shellcheck
    check_yamllint
}

for arg in "$@"
do
    case $arg in
        --check-black) check_black;;
        --check-flake8) check_flake8;;
        --check-pydocstyle) check_pydocstyle;;
        --check-shellcheck) check_shellcheck;;
        --check-yamllint) check_yamllint;;
        *) echo "invalid argument $arg" && exit 2;;
    esac
done
```

(3) We have the following `.yamllint.yaml` configuration file so that starting "---" is optional:

```yaml
extends: default

rules:
  document-start: disable
  line-length: disable
```

The goal of this issue is to introduce `yamllint` in a similar way to all our GitHub repositories. This will consist of:

- enriching `run-tests.sh` so that the checking would work locally;
- finding out an appropriate GitHub action and enriching `.github/workflows/ci.yml` so that the checking would be always executed during pull request time.

tiborsimko:
... and, when we are at enriching CI for all repositories, we could also introduce in the same PR set:

- `markdownlint` to check our `*.md` files that we also use everywhere in all the repositories;
- `.editorconfig` to introduce common editing style (tabs vs spaces, indentation length, etc).

I'll try to add some examples for the Markdown linting configuration and for the editor configuration later on.

PR #838, opened 2024-10-10
ci(yamllint): add YAML linting (#838)

jlemesh:
Related to https://github.com/reanahub/reana/issues/836

Add YAML linting for all files except Helm chart files, which are linted by `helm lint` and `ct lint` (`yamllint` does not understand templating language and fails with invalid syntax errors).Update `values-dev.yaml` file to pass linter checks.

Add `.editorconfig` file with rules for various files. Adding also .go as I remember we were talking that it would be good if the file could be used everywhere, we'll be able to copypaste it to other repositories and there will be no need for changes if new files are introduced to the repository. Let me know if it is better to only add config only for the files that currently exist in the repository.

For markdownlint, some rules were disabled because there will be to many changes to existing files (e. g. line length). Some rules do not work due to some specific contents in files, e. g. we have cron schedule defined (* * * * *) and markdownlint thinks this is emphasis (**bold**) and throws warning that there is a space between emphasis asterisks.


.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*.{yaml,yml}]
```
tiborsimko, line 1: Can you take away the `.editorconfig` into its own commit?

It would be useful to add rules for the other usual suspects we have: Go files, Markdown files, Python files, shell scripts, Dockerfiles, etc.  

And you can also add a general section about trimming trailing white space, inserting final new line, etc.

For inspiration, please see what our Invenio/Indico/Zenodo colleagues use:

- https://github.com/inveniosoftware/invenio-records/blob/master/.editorconfig
- https://github.com/indico/indico/blob/master/.editorconfig
- https://github.com/zenodo/zenodo/blob/master/.editorconfig

We should however modernise the rules, e.g. for Python we are using `black` code formatting, so the `.editorconfig` rule might look as simply as:

```
[*.py]
profile = black
```

So, can you please look around at the best practices, pick a configuration that is very close to our current code formatting in nature (so that our code base would not have to change much), and assemble all these editor rules into a new separate commit in order to make a separate release note announcement? (Perhaps under the `style` scope of the conventional commits fame, see https://github.com/reanahub/reana/wiki/Tips-for-Git)

.markdownlint.yaml
```diff
@@ -0,0 +1,11 @@
+# MD013/line-length : Line length : https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md013.md
+MD013: false
+
+# MD037/no-space-in-emphasis : Spaces inside emphasis markers : https://github.com/DavidAnson/markdownlint/blob/v0.35.0/doc/md037.md
```
tiborsimko, line 4: Some of these checks are not necessary as they would catch only a handful of cases. I'll amend the sources and remove them and repush.

PR #848, opened 2024-11-23
feat(helm): release check on most-supported Kubernetes version

tiborsimko:
Since Kubernetes 1.22, there was never a big need to actually upgrade REANA source code to make it work on new Kubernetes releases. (This is thanks to the fact that REANA uses pretty stable Kubernetes API internally.) This commit therefore releases the upper boundary check on the supported Kubernetes versions in order to be more welcoming to future Kubernetes upgrades, alleviating the need to manually bump the supported version number.

### DIFF
--- a/helm/reana/Chart.yaml
+++ b/helm/reana/Chart.yaml
@@ -27,7 +27,7 @@ keywords:
   - reusable-science
 type: application
 version: 0.9.3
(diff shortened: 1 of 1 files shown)


mdonadoni (approved):
LGTM!

PR #867, opened 2025-01-31
Harmonize default value for REANA_HOSTNAME

### LINKED ISSUE #865: helm: harmonise the treatment of `reana_hostname`
### Current behaviour

In the REANA Helm chart, we have `reana_hostname` commented out, and a somewhat inconsistent handling of `reana_hostname`'s default value in case the admin does not configure it:

```console
$ rg reana_hostname
helm/reana/templates/secrets.yaml
98:  {{- $cert := genSelfSignedCert (.Values.reana_hostname | default "localhost") nil nil 90 }}

helm/reana/templates/reana-workflow-controller.yaml
111:            value: {{ .Values.reana_hostname | default "" | quote  }}

helm/reana/templates/NOTES.txt
32:    $ firefox https://{{ default "localhost:30443" .Values.reana_hostname }}
39:    $ export REANA_SERVER_URL=https://{{ default "localhost:30443" .Values.reana_hostname }}
```

Analogously, in the REANA component source code base, the environment variable `REANA_HOSTNAME` reading is handled a bit differently in different components:

```console
$ rg REANA_HOSTNAME | grep 'os\.'
reana-server/reana_server/config.py:REANA_HOSTNAME = os.getenv("REANA_HOSTNAME")
reana-workflow-controller/reana_workflow_controller/config.py:REANA_HOSTNAME = os.getenv("REANA_HOSTNAME", "CHANGE_ME")
```

This causes certain URLs to not look pretty, for example notice the `CHANGE_ME` in the Dask dashboard URL:

```console
$ kubectl exec -i -t deployment/reana-db -- psql -U reana
psql (14.10 (Debian 14.10-1.pgdg120+1))
Type "help" for help.

reana=# select * from __reana.service;
                 id_                  |         name          |                                   uri                                   | status  | owner_id | type_
--------------------------------------+-----------------------+-------------------------------------------------------------------------+---------+----------+-------
 6eb43922-3c90-44b3-b21c-d555a4f8725f | dask-service-f24f0a14 | https://CHANGE_ME/f24f0a14-b59c-40eb-8104-53d77286ce5a/dashboard/status | running |          | dask
(1 row)
```

### Expected behaviour

It would be good to harmonise the treatment of `reana_hostname` across the code base. For example, we could require it, and fill some convenient default:

- e.g. switch everything to use `localhost` as default, provided that this value would be OK for various local development deployment scenarios (such as Docker Desktop VM on macOS).

- e.g. switch everything to use something else (say `127.0.0.1`) as default, if that is better for various deployment scenarios;
 
- but we should not have other default values such as `CHANGE_ME`  across components.

Alputer:
This PR is part of harmonizing the treatment of REANA_HOSTNAME accross all REANA components. You can refer to other PRs below.

https://github.com/reanahub/reana-server/pull/717
https://github.com/reanahub/reana-workflow-controller/pull/630

Closes #865


reana/reana_dev/cluster.py
```diff
@@ -283,6 +302,12 @@ def cluster_build(
     callback=validate_mode_option,
     help="In which mode to run REANA cluster? (releasehelm,releasepypi,latest,debug) [default=latest]",
 )
+@click.option(
+    "--port",
```
tiborsimko, line 306: **suggestion:** ‏Might be good to call the variable `--hostport` for full compatibility with the Helm chart value naming.

helm/reana/values.yaml
```diff
@@ -2,9 +2,13 @@
...
 # Declare variables to be passed into your templates.
 
-# reana_hostname: reana.cern.ch
+reana_hostname: localhost
 # `reana_hostname` should be set for third party integrations to work and for
-# production deployments to be secure.
+# production deployments to be secure. The default `localhost` value is used
+# in local deployment scenarios.
+
+reana_hostport: 30444
```
tiborsimko, line 10: **todo:** ‏Shouldbe 443 by default.

helm/reana/values.yaml
```diff
@@ -2,9 +2,13 @@
...
+# `reana_hostport` should be set to same port you expose in your ingress controller.
```
tiborsimko, line 11: **suggestion:** ‏You can rather explain that "... should be set to the port number where the REANA service will be exposed for incoming SSL connections."  (Kind of using outward user-oriented language, not inward ingress-oriented language.)

helm/reana/values.yaml
```diff
@@ -170,7 +174,7 @@ traefik:
     web:
       nodePort: 30080
     websecure:
-      nodePort: 30443
+      nodePort: 30444
```
tiborsimko, line 177: **todo:** ‏You should revert back to 30443 after testing is done.

reana/reana_dev/cluster.py
```diff
@@ -86,6 +92,7 @@ def cluster_create(
        $ reana-dev cluster-create -m /var/reana:/var/reana
                                   -m /usr/share/local/mydata:/mydata
                                   --mode debug
+                                  --extra-ports 30080 30443 5000
```
tiborsimko, line 95: **todo:** ‏I'm note sure port 5000 would work, Kind/K8s may need ports above 30000.  You can use a realistic exampe of 30444 for running a second REANA instance in the documentation pages.

reana/reana_dev/cluster.py
```diff
@@ -109,32 +145,15 @@ def add_volume_mounts(node):
                 '    node-labels: "ingress-ready=true"\n'
             )
         ],
-        "extraPortMappings": [
-            {"containerPort": 30080, "hostPort": 30080, "protocol": "TCP"},  # HTTP
-            {"containerPort": 30443, "hostPort": 30443, "protocol": "TCP"},  # HTTPS
-        ],
+        "extraPortMappings": extra_port_mappings,  # Only user-specified ports
```
tiborsimko, line 148: **question:** ‏Before, 30080 and 30443 were automatically added. Now, they would not be, so developers would have to pass `--extra-ports 30080 30443 manually.  Perhaps you can check the argument, and if nothing was specified, fall back to 30080 and 30443, so that old behaviour would work out of the box?

reana/reana_dev/run.py
```diff
@@ -288,6 +294,7 @@ def run_ci(
                           -c r-d-helloworld
                           --exclude-components=r-ui,r-a-krb5,r-a-rucio,r-a-vomsproxy
                           --mode debug
+                          --port 30500
```
tiborsimko, line 297: **todo:** ‏Please use realistic examples, i.e. 30443 default value. The help would be more easily copy-pasteable.

scripts/create-admin-user.sh
```diff
@@ -42,7 +43,7 @@ kubectl -n "${kubernetes_namespace}" create secret generic "${instance_name}"-ad
 # Success!
 echo "Success! You may now set the following environment variables:"
 echo ""
-echo "  $ export REANA_SERVER_URL=https://localhost:30443  # or use your URL"
+echo "  $ export REANA_SERVER_URL=https://localhost:${reana_hostport} # or use your URL"
```
tiborsimko, line 46: **suggestion:** ‏`localhost` still stays hard-coded here. If we want to pass the host port, we should probably pass the host name too, for consistency.  Alternatively, let's just keep the old content, i.e. not introduce any new argument, since the localhost:30443 part is only printed out as a help.  I guess I would prefer the latter, so that we don't change the number of arguments to this script.  (It does not seem necessary to know the port number when creating an admin user...)

tiborsimko (approved):

PR #868, opened 2025-02-03
fix(reana-dev): use chore commit scope when bumping dependencies

tiborsimko:
For commits created by `reana-dev git-upgrade-shared-modules` and `reana-dev git-upgrade-requirements` commands, use "chore" rather than "build" conventional commit scope, so that these commits won't appear in the release news message.

The bug fixes or new features that are brought by upgrades are announced in the release news in another way, such as when a feature needs the new shared package or an upgraded version of dependency. The pre-release version bumps can therefore go without a release news item.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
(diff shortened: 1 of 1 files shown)


Alputer (approved):
LGTM!

PR #884, opened 2025-03-25
feat(reana-dev): add `--create-branch` option to `git-upgrade` command

Alputer:
This commit adds a `--create-branch` option to the `git-upgrade` command. When used, it creates the branch locally from the upstream branch and pushes it to the origin, if the branch exists in upstream. This option is especially useful when upgrading maintenance branches that don't exist locally or in your fork.


CameronMcClymont:
Taken over and replaced by https://github.com/reanahub/reana/pull/957.

PR #885, opened 2025-03-25
feat(helm): introduce `node_label_infrastructurewc` Helm value

Alputer:
Closes reanahub/reana-workflow-controller#641


Alputer:
What do you think about using `node_label_infrastructure` for workflow controller when `node_label_infrastructure` exists and `node_label_infrastructurewc` does not? @tiborsimko 
We can fall back to `node_label_infrastructure` in that scenario.

tiborsimko:
> What do you think about using `node_label_infrastructure` for workflow controller when `node_label_infrastructure` exists and `node_label_infrastructurewc` does not? 

Yes, exactly, we should behave just like for `infrastructuremq` or `infrastructuredb`. They should be optional. If they are set, then use them; otherwise use the generic `infrastructure` label.

.github/ct.yaml
```diff
@@ -2,3 +2,4 @@ validate-maintainers: false
 check-version-increment: false
 chart-dirs:
   - helm
+target-branch: master
```
tiborsimko, line 5: Here, in the `maint-0.9` branch, the `target-branch` should be `maint-0.9`, isn't it? (And we keep it on `master` when merging upstream.)

tiborsimko (approved):

PR #887, opened 2025-03-26
feat(helm): update uWSGI config map

Alputer:
Add new uWSGI parameters after upgrading to version 2.0.28.

- `idle`: Restarts the worker after it has been idle for the specified number of seconds.
- `max-worker-lifetime`: Restarts the worker after running for the specified number of seconds.


CameronMcClymont:
Replaced by https://github.com/reanahub/reana/pull/953 which incorporates these changes.

PR #889, opened 2025-04-03
feat(helm): add k8s resource request/limits for infra components

Alputer:
This PR adds optional Kubernetes resource requests and limits for infrastructure component pods. It complements #888, which introduced resource configuration for user job pods.

All fields default to `{}` (no constraints applied), so this is fully backwards compatible. Operators can set per-component limits by overriding the relevant resources key in their `values.yaml`:

```
components:
  reana_server:
    rest_api:
      resources:
        requests:
          cpu: "500m"
          memory: "1Gi"
        limits:
          cpu: "2"
          memory: "4Gi"
```

### Scope

The resource knobs are applied to the long-lived infrastructure pods:

- `reana-ui`, `reana-db`, `reana-cache`, `reana-message-broker`
- `reana-server` (both REST API and scheduler containers)
- `reana-workflow-controller` (REST API and job-status-consumer containers)
- `pgbouncer` (optional connection pooler `Deployment`)

The following are intentionally out of scope:

- **reana-job-controller and workflow-engine containers**: these run in dynamically created per-workflow run-batch pods rather than Helm-managed long-lived infrastructure workloads, so they are intentionally out of scope. PR #888 configures the user job pods spawned by reana-job-controller, not the run-batch containers themselves.
- **Admin/cleanup cronjobs** (`system-status`, `resource-quota-update`, `retention-rules-apply`, `interactive-session-cleanup`): these are short-lived, periodic jobs so requests/limits pinning isn't really needed
- **reana-mail** and **reana-wdb**: debug-only dev tools that never run in production, so production resource configuration doesn't apply


michaelbuchar:
This PR still leaves some infrastructure workloads without configurable resources, so it does not fully satisfy “infra components” unless that scope is narrowed. The PR covers the always-on core pods, but optional infra workloads like PgBouncer and the admin cronjobs still have no resources hook (pgbouncer.yaml (line 35), cronjobs.yaml (line 24), cronjobs.yaml (line 149), cronjobs.yaml (line 235), cronjobs.yaml (line 330)). If those are intentionally out of scope, the PR title/body and issue linkage should say so and let me know!

michaelbuchar:
Also, there are some merge conflicts here unfortunately again :/

CameronMcClymont:
> This PR still leaves some infrastructure workloads without configurable resources, so it does not fully satisfy “infra components” unless that scope is narrowed. The PR covers the always-on core pods, but optional infra workloads like PgBouncer and the admin cronjobs still have no resources hook (pgbouncer.yaml (line 35), cronjobs.yaml (line 24), cronjobs.yaml (line 149), cronjobs.yaml (line 235), cronjobs.yaml (line 330)). If those are intentionally out of scope, the PR title/body and issue linkage should say so and let me know!

I think this is fine, but Alp is the original owner of the PR. I had reviewed his work and made some changes. I guess it makes sense to only have configurable resources for the permanent pods? But maybe @tiborsimko knows and could comment?

michaelbuchar:
This looks like a useful step forward, but I think the current scope/title still overstates what is covered. The PR adds configurable requests/limits for several infrastructure pods, but `pgbouncer` still has no corresponding resources hook, so operators using PgBouncer still cannot tune that workload. If the intent is to cover infrastructure components broadly, I think pgbouncer should be included as well. Otherwise, I’d suggest narrowing the PR title/body to make the scope explicit.

michaelbuchar:
I think the current wording still implies broader coverage than the implementation provides. The new resource knobs cover the main long lived infrastructure pods, but the periodic admin/cleanup cronjobs are still outside this mechanism. That means operators still cannot configure requests/limits for those workloads. If those cronjobs are meant to stay out of scope, I’d suggest narrowing the PR title/body to make that explicit. Otherwise, it may be worth extending the same pattern to those jobs as well.

CameronMcClymont:
Added configurable resources for pgbouncer now and clarified about the cronjobs in the PR description.

michaelbuchar:
Final comment: This PR still omits the debug-only long-lived infra deployments reana-mail and reana-wdb. When debug.enabled=true, both become persistent Deployments, but neither template has a resources hook (helm/reana/templates/reana-mail.yaml, helm/reana/templates/reana-wdb.yaml). If the intended scope is “infra components” broadly, these should either get the same treatment or be explicitly excluded in the PR scope.

CameronMcClymont:
> Final comment: This PR still omits the debug-only long-lived infra deployments reana-mail and reana-wdb. When debug.enabled=true, both become persistent Deployments, but neither template has a resources hook (helm/reana/templates/reana-mail.yaml, helm/reana/templates/reana-wdb.yaml). If the intended scope is “infra components” broadly, these should either get the same treatment or be explicitly excluded in the PR scope.

Yes, since they're debug only I don't think it really makes sense to have them be configurable like this. I will update the commit/PR description and rebase instead.

michaelbuchar (dismissed):
LGTM, just the commitlint needs fixing but moving this to in-merge

helm/reana/README.md
```diff
@@ -130,15 +133,20 @@ collisions.
...
+| `components.reana_workflow_controller.environment.REANA_RUNTIME_SESSIONS_SUPPLEMENTAL_GROUPS`             | Comma-separated supplemental groups for interactive-session pods. Defaults to `100` for Jupyter Docker Stacks compatibility. Set to an empty string to omit `supplementalGroups` from session pods.                                                                                                                                                                                                                          | `100`                                                                                                                                                                                                                                                                 |
```
tiborsimko: **🤖 R889-02: Workflow-controller README row moved between workflow-engine sections (NIT)**

The diff moves the `components.reana_workflow_controller.environment.REANA_RUNTIME_SESSIONS_SUPPLEMENTAL_GROUPS` row out of the workflow-controller block and re-inserts it between the CWL and Serial workflow-engine blocks (`helm/reana/README.md:149`):

```
| `components.reana_workflow_engine_cwl.image`                                                   | REANA-Workflow-Engine-CWL image to use ...
| `components.reana_workflow_controller.environment.REANA_RUNTIME_SESSIONS_SUPPLEMENTAL_GROUPS`  | Comma-separated supplemental groups for interactive-session pods ...
| `components.reana_workflow_engine_serial.environment`                                          | [REANA-Workflow-Engine-Serial] environment variables ...
```

A workflow-controller environment parameter now sits between the CWL image row and the Serial engine section, breaking the table's per-component grouping — readers scanning the `reana_workflow_controller.environment.*` block will not find it there. The other relocated workflow-controller rows (`image`, `imagePullPolicy`, `pod_security_context`, and the new `rest_api.resources`/`job_status_consumer.resources`) all landed in correct positions; this is the only stray.

Suggested fix — move the row back to its pre-PR position among the other `components.reana_workflow_controller.environment.*` rows, i.e. between `...environment.IMAGE_PULL_SECRETS` (line 135) and `...environment.WORKSPACE_DISPLAY_FILE_LIMIT` (line 136).

tiborsimko (commented):
**🤖 R889-01: PR description misstates how reana-job-controller resources are controlled (LOW)**

The PR description's scope section says:

> **reana-job-controller**: its resources are controlled via the user job pod configuration introduced in #888.

That conflates the spawner with the spawned. PR #888 expanded support for the `kubernetes_jobs_*` resource parameters, which are forwarded as `REANA_KUBERNETES_JOBS_*` environment variables. These govern the **user job pods that reana-job-controller spawns**, not the job-controller container itself. The job-controller container runs inside the per-workflow run-batch pod built by reana-workflow-controller's `workflow_run_manager.py`, where both the workflow-engine and job-controller `V1Container`s are constructed with no `resources` argument, and `components.reana_job_controller` in `values.yaml` has no `resources` key either. So neither #888 nor this PR lets an operator configure requests/limits for the job-controller (or workflow-engine) containers, and an operator reading the description would look for a knob that does not exist.

Excluding the run-batch pod is perfectly defensible — it is a dynamically created runtime workload rather than a Helm-managed deployment — but the description should say that, rather than pointing at #888.

Suggested fix — reword the bullet along these lines:

> **reana-job-controller and workflow-engine containers**: these run in dynamically created per-workflow run-batch pods rather than Helm-managed long-lived infrastructure workloads, so they are intentionally out of scope. PR #888 configures the user job pods spawned by reana-job-controller, not the run-batch containers themselves.

helm/reana/README.md
tiborsimko: Fixed during merge.

tiborsimko:
> **🤖 R889-01: PR description misstates how reana-job-controller resources are controlled (LOW)**

Fixed during merge.

tiborsimko (approved):

PR #898, opened 2025-06-26
ci(commitlint): fix local running of commit linter on macOS

tiborsimko:
Replaces GNU-specific `grep` command-line option in the commitlint check so that the linter would work locally on both macOS and Linux operating systems.


michaelbuchar (approved):
Looks good on my end! 🚀

PR #900, opened 2025-07-11
feat(helm): add configurable workspace listing limits

### LINKED ISSUE #644: workspace: ensure smooth performance of file listings for workspaces containing 250k files
## Current behaviour

When a workspace contains many produced output files, the `reana-workflow-controller` pod starts to use too much CPU and too much memory. It could happen that the whole system "freezes". This was observed for a workflow that was producing over 250k files, and when the researcher clicked on the "Workspace" tab of the REANA web interface.

(When this happened, the REANA web interface was slow to serve other users as well. The solution was to delete the workflow controller pod that was "stuck" in calculating the workspace listing response.)

## Expected behaviour

We should improve the file serving behaviour for workspaces containing large number of files so that the workflow controller would not "freeze". This could be achieved in several manners:

1. Simply refuse to serve the full file content if there are more than N files in the workspace, and serve only N files. (N should be configurable by the REANA cluster admins, because some deployments may have worker nodes with a lot of memory, some with smaller memory. For example, N could be set to 10k.)
2. Allow specifying filtering, such as `reana-client ls -w myanalysis --filter name=mydirectory`, both from CLI and from Web UI, and restrict serving N files _after_ the filtering. If we do this, then in theory the researcher could repeat downloading all files by issuing many requests, each with desired filtering.
3. Improve pagination so that only parts of the workspace files are returned, i.e. process at most 10 files using some pre-defined sorting order (e.g. creation timestamps), and then move onto next N, so that all the files do not have to be processed in advance. This could have an advantage in case the workspace is modified afterwards manually, since the pagination could be "lost".

Either of these solutions would prevent "freezing" of the workflow controller component.

## Notes

This issue is related to #641

michaelbuchar:
Add WORKSPACE_DISPLAY_FILE_LIMIT and FORCE_GARBAGE_COLLECTION
environment variables for reana-workflow-controller to the
Helm chart README.

Closes https://github.com/reanahub/reana-workflow-controller/issues/644


helm/reana/values.yaml
```diff
@@ -125,6 +125,8 @@ components:
       REANA_OPENSEARCH_USE_SSL: true
       REANA_OPENSEARCH_CA_CERTS: "/code/certs/ca.crt"
       REANA_OPENSEARCH_USER: reana
+      WORKSPACE_DISPLAY_FILE_LIMIT: 100000
```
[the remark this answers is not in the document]
michaelbuchar: Let's do it now so that we don't forget about it :)

helm/reana/values.yaml
```diff
@@ -127,6 +127,8 @@ components:
...
+      FORCE_GARBAGE_COLLECTION: ""
```
tiborsimko: `FORCE_GARBAGE_COLLECTION` can be removed from `values.yaml` since `reana-workflow-controller` already defaults to `""` when the env var is not set:

```python
_gc_env = os.getenv("FORCE_GARBAGE_COLLECTION", "")
```

The README documentation added in this PR is sufficient for discoverability.

(Same applies to `REANA_QUOTA_MANAGEMENT_SECRET` which could be cleaned up separately — `reana-server` already defaults to `""` in its `config.py`.)

helm/reana/values.yaml
```diff
@@ -127,6 +127,8 @@ components:
       REANA_OPENSEARCH_USE_SSL: true
       REANA_OPENSEARCH_CA_CERTS: "/code/certs/ca.crt"
       REANA_OPENSEARCH_USER: reana
+      WORKSPACE_DISPLAY_FILE_LIMIT: 100000
```
tiborsimko: Same for `WORKSPACE_DISPLAY_FILE_LIMIT` — `reana-workflow-controller` already defaults to `100000` in its `config.py`:

```python
WORKSPACE_DISPLAY_FILE_LIMIT = int(os.getenv("WORKSPACE_DISPLAY_FILE_LIMIT", "100000"))
```

So this line can be removed from `values.yaml` as well, keeping only the README documentation.

tiborsimko:
Nit: consider rewording the commit message for release notes readability. The current headline "add workspace listing defaults" is a bit vague. Suggestion:

```
feat(helm): add configurable workspace listing limits (#900)

Add WORKSPACE_DISPLAY_FILE_LIMIT and FORCE_GARBAGE_COLLECTION
environment variables for reana-workflow-controller to the
Helm chart README.

Closes reanahub/reana-workflow-controller#644
```

helm/reana/values.yaml
michaelbuchar: Fixed

helm/reana/values.yaml
michaelbuchar: Fixed

tiborsimko (approved):

PR #902, opened 2025-07-16
feat(helm): add value to customise list of vetted container images

### LINKED ISSUE #728: validate: check workflow specification for unathorised container images
In some deployment scenarios we may want to allow running only those workflows which use vetted authorised container images.

(This is similar to how we allow to run only certain vetted Jupyter notebook images by users, see `interactive_sessions.environments.jupyter.recommended` Helm value.)

To address these deployment scenarios, it would be useful to:

- allow REANA cluster administrators to specify which images can be used by users in their runtime workflows;
- enrich `reana-client info` to report to users which images are authorised;
- amend `reana-client validate` to report any troubles early in case a user tries to use an unauthorised image;
- make strong check on the REANA server side not to accept user workflows if the workflow uses an unauthorised image for some of its steps.

The implementation could be similar to how we handle Jupyter images, i.e. via a Helm value. However, the list of authorised images may grow, so we may have to be ready to handle potentially large number of allowed items in the vetted container image list.

CameronMcClymont:
Add a `vetted_container_images` Helm value to toggle container image vetting and set a list of allowed images that user workflows are allowed to run on the cluster.

Closes reanahub/reana-server#728


helm/reana/templates/reana-server.yaml
```diff
@@ -121,6 +121,8 @@ spec:
             value: {{ .Values.interactive_sessions.maximum_inactivity_period | default "forever" | quote }}
           - name: REANA_INTERACTIVE_SESSIONS_ENVIRONMENTS
             value: {{ .Values.interactive_sessions.environments | toJson | quote }}
+          - name: REANA_VETTED_CONTAINER_IMAGES
+            value: {{ .Values.vetted_container_images | toJson | quote }}
```
michaelbuchar, line 125: Serializing the allowlist into one environment variable may become a scaling limit if the list grows substantially. Consider a mounted ConfigMap if large lists are expected.

CameronMcClymont: It should be able to handle thousands of images in the list no problem, so I don't think it needs a ConfigMap.

tiborsimko (approved):

Issue #800, opened 2024-05-15
system: Introduce Kueue as an Optional Job Submission Approach


xaviertintin:
This feature introduces [Kueue](https://kueue.sigs.k8s.io) as an optional alternative to the existing Kubernetes Job API for submitting user jobs in REANA. This enhances flexibility for REANA admins by allowing them to choose the most suitable job submission method for their specific requirements.

## Goals:

- Configurable Job Submission: Empower REANA administrators to configure Kueue as the job submission method during deployment by leveraging Helm values.
- Automated Kueue Deployment: Streamline the process by automatically deploying the Kueue cluster upon admin selection of the Kueue option within their Helm values file.
- Customizable Kueue: Provide basic configuration options for the Kueue deployment through Helm values and/or chart snippets, catering to specific admin preferences.
- Resource Constraint Compatibility: Ensure all user-defined job resource constraints (e.g., memory limits) are fully supported when using Kueue. This includes thorough testing of job limits and requests to verify their proper transmission through Kueue.
- Transparent Scheduling Feedback: Guarantee that scheduling errors, encompassing both successful scheduling and errors due  o unavailable resources, are clearly communicated back to users in the workflow logs for both the standard and Kueue approaches.

# Implementation Overview (Conceptual):

> The GitHub project with the implementation steps and discussions that lead up to this integration can be found [here](https://github.com/users/xaviertintin/projects/5)
> The repository with integration commands can be found [here](https://github.com/xaviertintin/Thesis)
> Official documentation expected to be released June 2024 at latest

### 1. [reana](https://github.com/reanahub/reana):
- Helm Value Configuration - (`reana/helm/reana/values.yaml`):
  - Introduce a new Helm value (e.g., kueueEnabled) to enable/disable Kueue usage.
- Environment Variable Configuration - (`reana/helm/reana/templates/reana-workflow-controller.yaml`):
  - Define an environment variable (e.g., KUEUE_ENABLED) to reflect the Helm value setting.
  
### 2. [reana-workflow-controller](https://github.com/reanahub/reana-workflow-controller):
- Retrieve the environment variable (KUEUE_ENABLED) in (`reana-workflow-controller/reana_workflow_controller/config.py`) 
-  Select the appropriate job submission method based on the deployment type in (`reana-workflow-controller/reana_workflow_controller/workflow_run_manager.py`):
    - Standard: Utilize the existing Kubernetes Job API.
    - Kueue: Implement Kueue job submission logic.
[...]

PR #903, opened 2025-08-13
feat(helm): add Kueue scheduling option for user workloads

CameronMcClymont:
Introduces [Kueue](https://kueue.sigs.k8s.io/docs/overview/) as an alternative way to submit user jobs.

Kueue is disabled by default.

### To test

- `git checkout` this PR and the related [`reana-job-controller`](https://github.com/reanahub/reana-job-controller/pull/492) and [`reana-workflow-controller`](https://github.com/reanahub/reana-workflow-controller/pull/658) PRs
- Set `kueue.enabled` to `true` in the `values.yaml` (also make sure that this isn't being overridden in `values-dev.yaml`)
- Create, build, and deploy the REANA cluster
- From this repo, run:
  - `./scripts/setup-kueue.sh --install`
  - `./scripts/setup-kueue.sh --configure`
- Run a workflow, e.g. `reana-client run -w root6-roofit`
- Observe Kueue being used: `./scripts/setup-kueue.sh --watch`
- [Optional] Remove Kueue from the cluster with `./scripts/setup-kueue.sh --remove`

Closes https://github.com/reanahub/reana/issues/800


helm/configurations/values-dev.yaml
```diff
@@ -93,3 +93,17 @@ dask-kubernetes-operator:
     - name: reana-code
       hostPath:
         path: /code/reana-dask-kubernetes-operator
+
+debug:
+  enabled: true
+
+vetted_container_images:
```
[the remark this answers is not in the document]
CameronMcClymont: Ah yes I meant to exclude `values-dev.yaml` from the commit, removed now

scripts/configure-kueue.sh
```diff
@@ -0,0 +1,113 @@
...
+  namespaceSelector: {} # match all.
+  resourceGroups:
+  - coveredResources: ["cpu", "memory"]
+    flavors:
+    - name: "default-flavor"
+      resources:
+      - name: "cpu"
+        nominalQuota: 3
+      - name: "memory"
+        nominalQuota: 4Gi
```
tiborsimko, line 69: For run-batch, we can have less CPU and less RAM value configured. This will provide a better example why we need two queues. (Small pods reana-run-batch running just workflow orchestration, large pods reana-run-job running lengthy user payload.) You can try to decease reana-run-batch queue resources and increase reana-run-job queue resources.

scripts/configure-kueue.sh
```diff
@@ -0,0 +1,113 @@
...
+sleep 5
+kubectl get clusterqueue -A --no-headers || true
+kubectl get localqueue -A --no-headers || true
+kubectl get resourceflavor -A --no-headers || true
+print_status "Kueue resources verified"
+
+echo ""
+echo -e "${GREEN}🎉 Kueue configuration completed!${NC}"
+echo ""
+echo "You can now submit jobs to REANA that run via Kueue."
```
tiborsimko, line 110: The output of this script is meant for admins who are installing REANA, not to the end users. So the final message should rather say something like: "You can now enable Kueue scheduler in your REANA instance by configuring `kueue.enabled` Helm chart value to `true`."

helm/reana/README.md
```diff
@@ -124,6 +124,7 @@ collisions.
...
+| `kueue.enabled`                                                                                           | Enable Kueue integration. If true, the Helm chart will create ClusterQueue and LocalQueue resources.                                                                                                                                                                                                                                                                                                                                 | false                                                                                                  |
```
tiborsimko, line 127: This help text is now outdated. Please say rather than it presupposes having a Kueue installed before, with two local queues named batch-queue and job-queue.  (Which is sort of hard-coded right now; admins cannot customise it yet.)

tiborsimko (approved):
Works nicely locally 👍

PR #908, opened 2025-09-25
feat(helm): add quota management secret configuration

### LINKED ISSUE #747: rest: add optional quota management endpoints
### Current behaviour

The CPU and Disk quota for users can be set through command-line by REANA administrators:

```
kubectl exec -i -t deployment/reana-server -c rest-api -- flask reana-admin quota-set -e john.doe@example.org -r disk -l 429496729600 --admin-access-token $REANA_ADMIN_ACCESS_TOKEN
```

This is perfectly fine, however for certain deployment scenarios, one would like to achieve this over REST API without having direct Kubernetes cluster access, so that first support line could manage user quotas without having any cluster privileges.

### Desired behaviour

It would be therefore good to introduce new REST API endpoints that would allow to consult and set the quota for users remotely.

This new REST API should be accessible only after providing certain secret which is not necessarily the same as REANA_ADMIN_ACCESS_TOKEN. This will allow to support scenarios where the frontend support line could be granted rights to manage user quotas without having access to much else.

This secret could be set in a Helm value. And, if the secret would be empty (default value), this functionality would be disabled (and so quota management would be only available from the command line, as is the case now).

The calling of the new REST API endpoints could be simply via `curl` commands. (We don't need to enrich reana-client.) The admin documentation under `docs.reana.io` should then be updated accordingly.

CameronMcClymont:
Closes reanahub/reana-server#747


michaelbuchar (dismissed):

helm/reana/values.yaml
```diff
@@ -107,6 +107,7 @@ components:
       REANA_RATELIMIT_GUEST_USER: "40 per second"
       REANA_RATELIMIT_AUTHENTICATED_USER: "40 per second"
       REANA_RATELIMIT_SLOW: "1/5 second"
+      REANA_ADMIN_QUOTA_MANAGER: ""
```
tiborsimko: Nomenclature: The name `REANA_ADMIN_QUOTA_MANAGER` and the subsequent help phrases such as "REANA admin quota management secret", may be misread in a way that people can read it as a managment of admin quotas (as opposed to the management of user quotas). 

Can we invent some unambiguous name for the feature, say `REANA_QUOTA_MANAGEMENT_SECRET` for consistency with `X-Quota-Management-Secret`, and call the feature in the help strings "user quota management endpoint" and such? (Here and in `r-commons` and `r-server`)

tiborsimko:
### AI review

**Commit headline suggestion:**

```
feat(helm): add quota management secret configuration (#908)
```

(Align with the new `REANA_QUOTA_MANAGEMENT_SECRET` naming — the current message still says "admin quota manager" which was the old naming.)

tiborsimko (approved):

PR #923, opened 2026-01-20
feat(helm): roll all affected deployments when secrets change

tiborsimko:
Split secrets into individual secret files and add checksum annotations to deployments so that pods automatically restart when their respective secrets change. This follows the same pattern established for database secrets previously.

Secret to deployment mappings are:
- App secrets -> r-server, r-w-controller
- Cache secrets -> r-server, r-cache
- GitLab secrets -> r-server, r-w-controller
- Login provider secrets -> r-server
- Mail notification secrets -> r-server
- Message broker secrets -> r-server, r-w-controller, r-m-broker
- OpenSearch secrets -> r-w-controller
- SSO secrets (CERN, EOSC) -> r-server

Note that secret values now use `hasKey` instead of `default` to check whether a value is set. This means that explicitly setting a secret to an empty string (e.g. `REANA_GITLAB_HOST: ""`) will now use the empty string rather than falling back to the default value. If a key is not defined at all, the default is still used.


michaelbuchar (approved):
LGTM!

PR #931, opened 2026-03-20
feat(helm): add quota period Helm support

michaelbuchar:
Closes https://github.com/reanahub/reana-server/issues/774


helm/reana/templates/reana-server.yaml
```diff
@@ -97,6 +97,10 @@ spec:
             value: {{ .Values.quota.default_cpu_limit | default 0 | int64 | quote }}
           - name: REANA_DEFAULT_QUOTA_DISK_LIMIT
             value: {{ .Values.quota.default_disk_limit | default 0 | int64 | quote }}
+          - name: REANA_DEFAULT_QUOTA_CPU_PERIODIC_RESET_ENABLED
+            value: {{ .Values.quota.cpu_periodic_reset.enabled | default false | quote }}
+          - name: REANA_DEFAULT_QUOTA_CPU_PERIODIC_RESET_MONTHS
+            value: {{ .Values.quota.cpu_periodic_reset.period_months | default 3 | quote }}
```
tiborsimko: 🤖 **`period_months` value rendered without the `int64` filter (HIGH)**

The numeric env vars just above (lines 97 and 99 — `default_cpu_limit`, `default_disk_limit`) use `| default N | int64 | quote` to force the value through an int64 coercion before quoting. The new `period_months` line skips `int64`. If a user accidentally sets `period_months: "3"` (string) in their custom values.yaml, the surrounding rendering will pass the literal string through, and the env var will be `"3"` either way today — but the *contract* the rest of the file establishes for numeric env vars is "always run through int64 first." Drifting from it now also means a future Helm version that tightens type handling could cause an inconsistent failure mode for this one var only.

Suggested fix — match the convention used for `default_cpu_limit` / `default_disk_limit`:

```yaml
          - name: REANA_DEFAULT_QUOTA_CPU_PERIODIC_RESET_MONTHS
            value: {{ .Values.quota.cpu_periodic_reset.period_months | default 3 | int64 | quote }}
```

helm/reana/README.md
```diff
@@ -292,6 +292,8 @@ collisions.
...
+| `quota.cpu_periodic_reset.period_months`                                                                  | Length of the CPU accounting window in months.                                                                                                                                                                                                                                                                                                                                                                               | 3                                                                                                      |
```
tiborsimko: 🤖 **Document that these settings apply to newly created users only (NIT)**

The `enabled` description says "for newly created users" — good — but `period_months` says "Length of the CPU accounting window in months" without qualification. An operator skimming the table might assume changing this affects *all* users. Worth aligning the wording and pointing to the backfill path for existing users.

Suggested fix:

```markdown
| `quota.cpu_periodic_reset.enabled`         | Enable deployment default periodic CPU quota for newly created users. Existing users can be migrated via `reana-db quota init-periodic-defaults`. | false |
| `quota.cpu_periodic_reset.period_months`   | Length of the CPU accounting window in months for newly created users. Has no effect on users with an existing periodic quota policy.            | 3     |
```

tiborsimko (commented):
Inline review of the Helm chart changes. Two items, grouped by severity (HIGH / NIT).

helm/reana/templates/reana-server.yaml
michaelbuchar: Fixed

helm/reana/README.md
michaelbuchar: Fixed

helm/reana/values.yaml
```diff
@@ -251,11 +251,16 @@ pgbouncer:
 # Quota
 quota:
   enabled: true
+  default_disk_limit: 0
+  default_cpu_limit: 0
   periodic_update_policy: "{{ .Values.quota.disk_update }}"
   workflow_termination_update_policy:
     "{{ .Values.quota.termination_update_policy }}"
+  cpu_periodic_reset:
+    enabled: false
```
tiborsimko: I'm not sure we would need one more enable? The admins would have to write something like:

```yaml
quota:
  enabled: true
  default_cpu_limit: 1000000
  default_disk_limit: 2000000
  cpu_periodic_reset:
    enabled: true
    period_months: 3
```

We could perhaps simplify to:

```yaml
quota:
  enabled: true
  default_cpu_limit: 1000000
  default_disk_limit: 2000000
  cpu_limit_periodic_reset_period_in_months: 3
```

and the value of 0 would indicate that we don't want to have this enabled.

(The used variable name is just for illustration; we could invent nicer/shorter one... Some more examples for illustration: `reset_cpu_limit_in_months`, `cpu_limit_reset_period`, etc.)

IOW, the structure of the quota part could be flat; we don't need nested part about perioric resets...

michaelbuchar: Fixed!

tiborsimko (approved):

PR #939, opened 2026-04-14
fix(helm): quote Helm template values consistently

tiborsimko:
Ensure all environment variable values in Helm templates use the `| quote` filter to prevent YAML parsing issues on older Helm versions (e.g. 3.18) where large numeric values like 1209600 could be misinterpreted as scientific notation (1.209600e+06).

Replace `!!str` with `| quote` for timeout and Dask worker count defaults. Add `| quote` to memory, Dask, namespace, hostname, email, image, node label, and quota policy values that were previously unquoted. Add `| int` to the PriorityClass value to ensure integer rendering. Quote the large numeric `file_preview_size_limit` default in values.yaml.

### DIFF
--- a/helm/reana/templates/cronjobs.yaml
+++ b/helm/reana/templates/cronjobs.yaml
@@ -39,7 +39,7 @@ spec:
             env:
             {{- if .Values.reana_hostname }}
             - name: REANA_HOSTNAME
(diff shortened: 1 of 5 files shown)


michaelbuchar (approved):
Works for me! 🚀

PR #940, opened 2026-04-14
feat(scripts): upgrade Kueue to v0.17.0

tiborsimko:
Upgrade the Kueue Helm chart from v0.13.2 to v0.17.0 and update all
resource manifests to use v1beta2 API versions for ClusterQueue and
LocalQueue resources.

Use fully-qualified CRD names (e.g. workloads.kueue.x-k8s.io)
consistently across all commands to avoid ambiguity.

Fix install bootstrapping deadlock where the controller pod would
crash-loop before the apiserver cached the CRD conversion webhook.
Detect CrashLoopBackOff and restart the pod, then wait for the
cert-rotator to patch the webhook caBundle before proceeding.

Fix uninstall to wait for Kueue CRDs to terminate and force-remove
finalizers if they get stuck after the controller is gone.

Fix namespace flag quoting in wait_for_cleanup, delete_resources
and force_delete_resources to avoid passing empty string arguments
to kubectl. Fix delete_resources to be namespace-aware and remove
leftover debug output. Fix LocalQueue removal to use -A flag.

### DIFF
--- a/scripts/manage-kueue.sh
+++ b/scripts/manage-kueue.sh
@@ -67,7 +67,7 @@ wait_for_cleanup() {
 
     print_info "Waiting for $resource_type cleanup..."
     while [ $count -lt $timeout ]; do
-        if [ "$(kubectl get "$resource_type" "$namespace_flag" --no-headers 2>/dev/null | wc -l)" -eq 0 ]; then
+        if [ "$(kubectl get "$resource_type" ${namespace_flag:+$namespace_flag} --no-headers 2>/dev/null | wc -l)" -eq 0 ]; then
             return 0
         fi
         sleep 1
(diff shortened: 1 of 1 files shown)


scripts/manage-kueue.sh
```diff
@@ -200,26 +250,27 @@ configure_kueue() {
...
 
-    print_section "📄 Applying Kueue resource manifests..."
+    print_section "⚙️  Applying Kueue resource manifests..."
     print_info "Batch : ${BATCH_CPU} CPU | ${BATCH_MEMORY} memory"
     print_info "Job   : ${JOB_CPU} CPU | ${JOB_MEMORY} memory"
 
+    echo -e "${YELLOW}$ kubectl apply -f -${NC}"
     cat <<EOF | kubectl apply -f -
 ---
 apiVersion: kueue.x-k8s.io/v1beta1
```
CameronMcClymont: Should this also be upgraded to v1beta2 like the other resources?

CameronMcClymont (commented):
Tested with my Kueue setup and did a quick code inspection and LGTM, just one comment regarding the ResourceFlavor manifest version.

scripts/manage-kueue.sh
tiborsimko: Yes, updated.

CameronMcClymont (approved):

PR #945, opened 2026-04-24
feat(helm): add configurable UI security headers

CameronMcClymont:
Add a Helm-managed Nginx ConfigMap for REANA-UI and mount it over the
image's default server configuration.

Expose HTTP security headers through chart values, including HSTS,
frame, content-type, cross-origin, referrer, permissions, and content
security policies. Restart UI pods automatically when the configuration
changes.

Document why the UI content security policy intentionally permits
additional resources compared with reana-server, and test correct
escaping of quoted header values.


helm/reana/values.yaml
```diff
@@ -93,6 +93,24 @@ components:
...
+        permissions_policy:
+          "accelerometer=(), ambient-light-sensor=(), camera=(),
+          display-capture=(), geolocation=(), gyroscope=(), magnetometer=(),
+          microphone=(), payment=(), usb=()"
+        # NOTE: keep in sync with reana-server config.py content security policy headers
+        csp:
+          "default-src 'self'; script-src 'self'; style-src 'self'
+          'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data:;
+          font-src 'self' data: https://fonts.gstatic.com; connect-src 'self';
+          frame-ancestors 'none'; object-src 'none'; base-uri 'self';"
```
michaelbuchar: This CSP breaks existing UI images. This blocks the existing [Launch-on-REANA badge](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/config.js#L26-L27) hosted at https://www.reana.io, and administrator-defined launcher thumbnails. The chart accepts arbitrary [image_url](https://github.com/reanahub/reana/blob/8913d5f85ec1fa5b8bb9874e53947db85a9a805b/helm/reana/templates/reana-config.yaml#L42-L47) values and the UI renders them as images.
I would widen img-src, for example with approved origins or https:, and document the tradeoff.

michaelbuchar:
Some new Helm values are not documented. The PR adds eight configurable headers, but the [REANA-UI values section](https://github.com/reanahub/reana/blob/8913d5f85ec1fa5b8bb9874e53947db85a9a805b/helm/reana/README.md#L92-L106) does not mention components.reana_ui.nginx.security_headers.*. Should we add these?

helm/reana/templates/nginx-config.yaml
```diff
@@ -0,0 +1,55 @@
+{{- if .Values.components.reana_ui.enabled }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nginx-config
+  namespace: {{ .Release.Namespace }}
```
michaelbuchar, line 7: Use a less generic ConfigMap name. This can collide with unrelated workloads in the namespace. Prefer {{ include "reana.prefix" . }}-ui-nginx-config and update the [volume reference](https://github.com/reanahub/reana/blob/8913d5f85ec1fa5b8bb9874e53947db85a9a805b/helm/reana/templates/reana-ui.yaml#L45-L49). This can be ignored if namespaces are guaranteed to be dedicated to one REANA deployment.

helm/reana/values.yaml
CameronMcClymont: Oh yeah nice catch, I've widened img-src to include `https:` and verified e.g. the badge works on the `/launcher-badge` page.

helm/reana/templates/nginx-config.yaml
CameronMcClymont: Good point thanks, updated the ConfigMap name to reana-ui-nginx-config, but the volume reference should be fine as-is since it's just a local binding (it doesn't leave the pod spec) so no collision risk?

michaelbuchar: Yes, agreed. The ConfigMap resource and configMap.name reference are now release-prefixed, which resolves the namespace collision risk. The local volume binding name nginx-config is scoped to this pod spec and does not need to be prefixed.

helm/reana/values.yaml
```diff
@@ -93,6 +93,24 @@ components:
...
+        permissions_policy:
+          "accelerometer=(), ambient-light-sensor=(), camera=(),
+          display-capture=(), geolocation=(), gyroscope=(), magnetometer=(),
+          microphone=(), payment=(), usb=()"
+        # NOTE: keep in sync with reana-server config.py content security policy headers
+        csp:
+          "default-src 'self'; script-src 'self'; style-src 'self'
+          'unsafe-inline' https://fonts.googleapis.com; img-src 'self' data:
+          https:; font-src 'self' data: https://fonts.gstatic.com; connect-src
+          'self'; frame-ancestors 'none'; object-src 'none'; base-uri 'self';"
```
michaelbuchar: Default CSP breaks PDF previews. object-src 'none' is enforced on UI pages, but REANA-UI renders PDF previews with an [`<object data={url}>`](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js#L89-L96). The [CSP specification](https://www.w3.org/TR/CSP/#directive-object-src) confirms that object-src controls embedded objects. Remove that restriction or replace inline PDF previews with links.

helm/reana/templates/nginx-config.yaml
```diff
@@ -0,0 +1,55 @@
...
+{{- with .Values.components.reana_ui.nginx.security_headers.referrer_policy }}
+        add_header Referrer-Policy "{{ . }}" always;
+{{- end }}
+{{- with .Values.components.reana_ui.nginx.security_headers.permissions_policy }}
+        add_header Permissions-Policy "{{ . }}" always;
+{{- end }}
+{{- with .Values.components.reana_ui.nginx.security_headers.csp }}
+        # NOTE: keep in sync with reana-server config.py content security policy headers
+        add_header Content-Security-Policy "{{ . }}" always;
+{{- end }}
```
michaelbuchar, line 38: Quoted header overrides can prevent Nginx startup. The template manually wraps values as "{{ . }}". A legitimate override such as:
```
permissions_policy: 'geolocation=(self "https://example.org")'
```
renders invalid Nginx syntax. I reproduced:
```
nginx: [emerg] unexpected "h" in /etc/nginx/conf.d/default.conf:12
```
Use Helm’s quote pipeline for each header value and add an nginx -t test with quoted origins.

helm/reana/values.yaml
CameronMcClymont: Thanks, relaxed from object-src 'none' to 'self'.

helm/reana/templates/nginx-config.yaml
CameronMcClymont: Thanks, using quote pipelines now and generated a test for this.

michaelbuchar (dismissed):
Nice, looks good to me, just needs rebasing!

PR #946, opened 2026-04-26
feat(helm): replace FLASK_ENV with FLASK_DEBUG for debug mode

### LINKED ISSUE #770: build(python): upgrade to latest Invenio packages, Flask 3.x, SQLAlchemy 2.x
Upgrade the Invenio dependency stack to latest versions:

- Invenio packages: invenio-db 2.5.0, invenio-accounts 7.x, invenio-app 3.x, invenio-oauthclient 7.x (upstream, replacing custom fork), and all other invenio-* packages to their latest major versions
- Flask 2.x → 3.x
- SQLAlchemy 1.4 → 2.x (via invenio-db 2.5.0 and reana-db)
- marshmallow 2.x → 3.x (required by new invenio-rest)
- webargs 5.x/6.x → 8.x
- Remove custom invenio-oauthclient fork (upstream now includes EOSC EU Node AAI provider)

This is a cross-component change affecting reana-db, pytest-reana, reana-commons, reana-server, reana-workflow-controller, reana-job-controller, reana-ui, and reana (Helm templates).

tiborsimko:
-   feat(helm): replace FLASK_ENV with FLASK_DEBUG for debug mode

    The FLASK_ENV environment variable was deprecated in Flask 2.3
    and removed in Flask 3.x. With the upgrade of REANA components
    to Flask 3.x, replace all FLASK_ENV=development settings in the
    Helm templates with the equivalent FLASK_DEBUG=1 across the
    reana-server, reana-workflow-controller, and cronjob deployments.

-   fix(reana-dev): install all client components in one pip call

    Previously, `reana-dev client-install` installed components
    one-by-one in sequence. This caused dependency resolution
    issues when local branches had different version pins than
    the published PyPI packages: earlier installs would lock in
    versions from PyPI that later installs could not override.

    Install all components in a single `pip install` command so
    that pip resolves constraints from all local source directories
    together, picking compatible versions across the board.

Closes reanahub/reana-server#770.

### DIFF
--- a/reana/reana_dev/client.py
+++ b/reana/reana_dev/client.py
@@ -11,13 +11,15 @@
 import base64
 import json
 import logging
+import os
 import subprocess
+import sys
 import traceback
 
 import click
 
 from reana.config import REPO_LIST_CLIENT
-from reana.reana_dev.utils import run_command
+from reana.reana_dev.utils import get_srcdir, run_command
 
 
 @click.group()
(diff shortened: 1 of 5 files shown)


reana/reana_dev/client.py
```diff
@@ -27,12 +28,23 @@ def client_commands():
 
 @client_commands.command(name="client-install")
 def client_install():  # noqa: D301
-    """Install latest REANA client and its dependencies."""
+    """Install latest REANA client and its dependencies.
```
michaelbuchar, line 32: ## AI review
### Missing repos are silently skipped by client-install

The new batching logic only appends a component when setup.py or pyproject.toml exists under its expected source directory. If a required checkout such as reana-db is absent locally, the command now skips it and pip can satisfy that dependency from PyPI, recreating the mixed local/PyPI environment this change is trying to avoid. The previous implementation failed immediately when a checkout was missing, which was safer. Consider erroring out when an expected client repo directory is absent and add a regression test for that path.

tiborsimko: Thanks. `client-install` now exits with an error if a `REPO_LIST_CLIENT` checkout directory is missing, and only silently skips when the directory exists but has neither `setup.py` nor `pyproject.toml` (the `reana-client-go` case). Folded into the original commit. Regression test will follow separately.

michaelbuchar (approved):

PR #947, opened 2026-05-11
feat(reana-dev): reuse existing local branch in git-checkout-pr

tiborsimko:
Add a new ``--reuse-existing-local-branch`` flag to ``reana-dev git-checkout-pr``. When set and a local ``pr-123`` branch already exists for pull request 123 for a component, the command checks it out instead of failing with ``a branch named 'pr-123' already exists``. No upstream code sync is performed in that case.


michaelbuchar:
The commitlint needs fixing, everything else looks good!

tiborsimko:
> The commitlint needs fixing, everything else looks good!

Fixed. And renamed the CLI option to "--reuse-..." which better captures the meaning of the flag than "--use-..." did.

michaelbuchar (approved):
LGTM!

PR #948, opened 2026-05-12
feat(reana-dev): add --issue option to git-checkout-pr

CameronMcClymont:
Adds `--pull` flag and introduces `--reset` to override local changes when pulling.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -58,6 +58,7 @@ from reana.reana_dev.utils import (
     upgrade_requirements,
     validate_directory,
     get_commit_pr_suffix,
+    find_standard_component_name,
 )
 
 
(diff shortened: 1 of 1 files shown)


reana/reana_dev/git.py
```diff
@@ -794,43 +794,162 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
             )
 
 
+def _get_prs_from_issue(repo, issue_number):
+    """Return (component, pr_number) pairs for all PRs linked to a GitHub issue.
+
+    Uses GitHub's issue timeline API to find cross-referenced events, which are
+    created when a PR references an issue (e.g. "Closes #123").
+    """
+    import requests
```
tiborsimko: 🤖 **Use `gh api` instead of `requests`**

`gh` is already a hard dependency of `reana_dev` and the conventional way this codebase talks to GitHub — see `git.py:408, 1541-1542` (`gh repo fork`, `gh pr create`, `gh pr list`) and `utils.py:1055, 1076` (`gh pr view --json …`, `gh {pr,issue} list --json …`), all using the `run_command(..., return_output=True)` + `json.loads(...)` pattern.

Switching the issue-timeline lookup to `gh api` resolves three concerns at once:

- **Auth / rate limit.** The current `requests.get(url)` call is unauthenticated, so it hits GitHub's 60 req/hour anonymous cap. `gh api` automatically uses the user's `gh auth login` token (5000 req/hour) — no `GITHUB_TOKEN` env plumbing needed.
- **Missing dependency.** `requests` is not declared in `reana/setup.py` `install_requires` (the existing `requests.get` in `utils.py:478` is a latent issue, but this PR widens the surface). Going through `gh api` sidesteps the question.
- **Manual pagination.** `gh api … --paginate` replaces the hand-rolled `response.links.get("next", {}).get("url")` loop.

It's also a style win — readers don't have to context-switch between "this reana_dev command shells out to gh" and "this one uses Python `requests`".

### Suggested fix

```python
def _get_prs_from_issue(repo, issue_number):
    """Return (component, pr_number) pairs for all PRs linked to a GitHub issue.

    Uses GitHub's issue timeline API to find cross-referenced events, which are
    created when a PR references an issue (e.g. "Closes #123").
    """
    display_message(f"Searching for PRs linked to reanahub/{repo}#{issue_number}...")

    output = run_command(
        f"gh api repos/reanahub/{repo}/issues/{issue_number}/timeline --paginate",
        display=False,
        return_output=True,
    )
    events = json.loads(output)

    prs = set()
    for event in events:
        if event.get("event") != "cross-referenced":
            # We're only looking for cross-referencing events (not e.g. comments)
            continue
        source = event.get("source", {}).get("issue", {})
        if not source.get("pull_request"):
            # We're only looking for events mentioning PRs
            continue
        prs.add((source["repository"]["name"], str(source["number"])))

    if not prs:
        click.secho(
            f"No PRs found linked to reanahub/{repo}#{issue_number}.",
            fg="yellow",
        )

    return list(prs)
```

The function-scoped `import requests` can then be dropped.

reana/reana_dev/git.py
```diff
@@ -794,43 +794,162 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
...
+
+
 @click.option(
     "--branch", "-b", nargs=2, multiple=True, help="Which PR? [component PR#]"
 )
-@click.option("--fetch", is_flag=True, default=False)
+@click.option(
+    "--issue", "-i", nargs=2, default=None, help="Derive PRs from a GitHub issue. [repo issue#]"
+)
+@click.option("--pull", is_flag=True, default=False, help="Fetch and fast-forward existing local PR branches?")
```
tiborsimko: 🤖 **Retain `--fetch` flag**

Removing `--fetch` is a silent breaking change — any existing script or shell alias that passes `--fetch` will start failing with Click's "no such option" error. The semantic it provided (refresh upstream remote-tracking refs, independent of local-branch manipulation) is also genuinely useful on its own and isn't fully covered by the new `--pull`/`--reset` flags:

- `--pull` and `--reset` couple the fetch to integration behaviour (ff-merge or hard-reset) that a reviewer may not want.
- For inspecting what changed upstream without touching the local `pr-123` branch — e.g. `git log pr-123..upstream/pr/123`, `git diff pr-123 upstream/pr/123` — the old `--fetch` was the natural primitive. The new design has no clean way to express "just refresh upstream refs for me, leave my local branches alone."

The fix is small: keep `--fetch` as an orthogonal flag that controls only whether `git fetch upstream` runs, independent of `--pull`/`--reset`.

### Suggested fix

```python
@click.option(
    "--fetch", is_flag=True, default=False,
    help="Fetch upstream remote-tracking refs first?"
)
@click.option(
    "--pull", is_flag=True, default=False,
    help="Fetch and fast-forward existing local PR branches?"
)
@click.option(
    "--reset", is_flag=True, default=False,
    help="Fetch and hard-reset existing local PR branches?"
)
def git_checkout_pr(branch, issue, fetch, pull, reset):
    ...
    for component, pull_request in prs_to_checkout:
        if fetch or pull or reset:
            run_command("git fetch upstream", component)
        ...
```

This keeps the breaking-change surface to zero (existing `--fetch` invocations keep working), and lets a reviewer say "just give me the latest upstream/pr/123, I'll handle the rest with `git diff pr-123 upstream/pr/123`" without opting into pull/reset semantics on existing local branches.

tiborsimko:
🤖 **PR description / code mismatch**

The PR description mentions:

> Also replace --fetch flag with --pull and introduce --force flag to override local changes when pulling.

But the actual code uses `--reset`, not `--force`. Looks like the flag was renamed during review but the description wasn't updated.

Worth tidying since release notes / changelogs often start from PR bodies, and this will confuse readers comparing the description against the code.

reana/reana_dev/git.py
```diff
@@ -794,43 +794,162 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
...
+    :param issue: Derive PRs to check out from the cross-references of a GitHub
+                  issue. The value consists of two strings: the reanahub
+                  repository name and the issue number. For example,
+                  ``-i pytest-reana 156`` will check out all PRs that
+                  reference the pytest-reana#156 issue.
+    :param pull: Fetch and fast-forward existing local PR branches? [default=False]
+    :param reset: Fetch and hard-reset existing local PR branches? [default=False]
     :type branch: list
-    :type fetch: bool
+    :type issue: tuple
```
tiborsimko: 🤖 **Docstring type for `issue` (NIT)**

The runtime value of `issue` is `Optional[Tuple[str, str]]` — it's `None` when the flag isn't passed, and a 2-tuple of strings when it is. The current `:type issue: tuple` only covers half of that.

The rest of this CLI is careful with sphinx-style types, so worth tightening:

```
:type issue: tuple or None
```

or, if you prefer the `Optional` spelling:

```
:type issue: Optional[tuple]
```

reana/reana_dev/git.py
```diff
@@ -794,43 +794,162 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
...
+            run_command("git fetch upstream", component)
+
+        if branch_exists(component, f"pr-{pull_request}"):
+            run_command(f"git checkout pr-{pull_request}", component)
+
+            if reset:
+                run_command(f"git reset --hard upstream/pr/{pull_request}", component)
+            elif pull:
+                dirty = run_command(
+                    "git status --porcelain", component, display=False, return_output=True
```
tiborsimko: 🤖 **Untracked files count as "dirty" (NIT)**

`git status --porcelain` outputs entries for both modified-tracked files *and* untracked files (lines prefixed with `??`). So the `if dirty:` skip in `--pull` will refuse to fast-forward whenever the working tree has any untracked files — even ones completely unrelated to the PR (e.g. local scratch files, `.pyc` leftovers if `.gitignore` is incomplete, a stray editor backup).

This is probably the *right* conservative behaviour — better to refuse and let the user reconcile than to fast-forward over potentially-meaningful unstaged state. But it's not what a casual reader expects from a variable called `dirty`, and a future contributor might "fix" it to filter out `??` lines and break the safety.

Worth a one-liner so the conservative choice is explicit:

```python
# `git status --porcelain` also lists untracked files. We intentionally
# treat them as "dirty" — refusing to ff-merge is safer than ff'ing
# over scratch state the user may want to preserve.
dirty = run_command(
    "git status --porcelain", component, display=False, return_output=True
)
```

reana/reana_dev/git.py
```diff
@@ -794,43 +794,162 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
...
+        for component, pull_request in _get_prs_from_issue(repo, issue_number):
+            if component not in REPO_LIST_ALL:
+                click.secho(
+                    f"Skipping PR #{pull_request} in {component} (not a known REANA component).",
+                    fg="yellow",
+                )
+                continue
+            prs_to_checkout.append((component, pull_request))
+
+    for component, pull_request in prs_to_checkout:
```
tiborsimko: 🤖 **Duplicate-component handling**

`prs_to_checkout` can end up with multiple entries for the same component from two sources (or even from the same source repeated), and the current loop silently lets the last entry win — without telling the user.

Two flavours of the same issue:

```
# explicit -b conflict
reana-dev git-checkout-pr -b reana-client 100 -b reana-client 101

# -b vs -i conflict (the new failure mode this PR introduces)
reana-dev git-checkout-pr -b reana-client 100 -i pytest-reana 156
# ↑ where issue 156 cross-references reana-client#101
```

In both cases the loop walks both entries: iteration 1 creates `pr-100`, iteration 2 creates `pr-101`, and the working tree ends up on `pr-101`. The user asked for `#100`; they get `#101`.

This isn't strictly new — the `-b -b` form was always able to collide. But two things change with this PR:

- **More likely.** With `-i`, the user doesn't directly see which PRs got pulled in from the issue's cross-references, so collisions with an explicit `-b` are easy to overlook.
- **Less detectable.** The previous code used `git checkout -b pr-N upstream/pr/N`, which errored out when `pr-N` already existed locally — i.e. the second iteration of a `-b -b` collision often failed loudly on a second invocation. The new `branch_exists()` path correctly handles existing branches with `git checkout pr-N`, but in doing so also swallows the safety net that previously surfaced these collisions.

Since the conflict-handling code now lives in one place (the consolidated `prs_to_checkout` loop), this is a good time to add a small guard that catches both flavours.

### Suggested fix

```python
# Detect per-component conflicts (whether from -b repetition or -b vs -i).
# Same (component, PR#) pair appearing twice is harmless and silently deduped.
by_component = {}
for component, pr in prs_to_checkout:
    if component in by_component and by_component[component] != pr:
        click.secho(
            f"Conflict for {component}: asked to check out both "
            f"pr-{by_component[component]} and pr-{pr}. "
            f"Remove the duplicate from -b/-i.",
            fg="red",
        )
        sys.exit(1)
    by_component[component] = pr
prs_to_checkout = list(by_component.items())
```

This errors loudly on real conflicts (different PRs for the same component) and silently dedups exact matches (same `(component, PR#)` from both sources — e.g. when `-i` happens to surface a PR you also passed explicitly via `-b`).

reana/reana_dev/git.py
CameronMcClymont: Switched to using gh api instead of requests.

reana/reana_dev/git.py
CameronMcClymont: Added `--fetch` back in.

reana/reana_dev/git.py
CameronMcClymont: I can update this to `:type issue: Optional[tuple]`, but then `branch`, `fetch`, `pull`, and `reset` should all be updated too since they're technically optional?

reana/reana_dev/git.py
CameronMcClymont: Added this check now.

reana/reana_dev/git.py
tiborsimko: 🤖 Good question, sorry the earlier NIT was unclear — two senses of "optional" got conflated:

- **CLI-optional**: the user doesn't have to pass the flag (Click has a default).
- **Type-optional** (`Optional[T]` = `T | None`): the *runtime value* can be `None`.

`:type` is Sphinx/Python-type, so only the second sense counts. A param that's CLI-optional but always non-`None` at runtime is just its base type, not `Optional`.

Walking the five:

| Param | Click decorator | Runtime value | `:type` |
|---|---|---|---|
| `branch` | `nargs=2, multiple=True` | always `tuple` (possibly `()`) | `tuple` |
| `issue` | `nargs=2, default=None` | `None` or `tuple` | `Optional[tuple]` ✅ |
| `fetch`/`pull`/`reset` | `is_flag=True, default=False` | always `bool` | `bool` |

So only `issue` is genuinely `Optional`. For `branch`, plain `tuple` is right (and also `tuple` not `list` — `multiple=True` returns a tuple; the pre-PR `:type branch: list` was already slightly off, my prior NIT missed it). The other three correctly stay `bool` as you have them.

reana/reana_dev/git.py
```diff
@@ -794,61 +794,200 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
...
+
+    for component, pull_request in branch:
+        component = select_components([component])[0]
+        if component not in REPO_LIST_ALL:
+            display_message("Ignoring unknown component.", component)
+            continue
+        prs_to_checkout.append((component, pull_request))
+
+    if issue is not None:
+        repo, issue_number = issue
```
tiborsimko: 🤖 **`-i` doesn't expand shortened component names**

Repro:

```
$ reana-dev git-checkout-pr -i r-server 774
[2026-05-13T11:18:17] : Searching for PRs linked to reanahub/r-server#774...
gh: Not Found (HTTP 404)
```

The `-b` path normalizes shortened names via `select_components()` at line 914 (so `-b r-server 774` correctly becomes `reana-server`), but the `-i` path destructures `repo, issue_number = issue` at line 921 and passes the raw string straight to `gh api repos/reanahub/{repo}/…`. Result: any short form that works for `-b` produces a 404 with `-i`.

Worth fixing — short forms are documented and used throughout the rest of `reana-dev`, so users will reasonably expect `-i r-server 774` to work just like `-b r-server 774`.

### Suggested fix

`find_standard_component_name()` (in `utils.py:69`) is the right primitive — it's idempotent on standard names and expands short ones. Also gracefully passes through non-REANA reanahub repos (e.g. `awesome-workflows`) unchanged, since the heuristic is "any multi-char prefix part means standard form."

```python
if issue is not None:
    repo, issue_number = issue
    try:
        repo = find_standard_component_name(repo)
    except Exception:
        click.secho(
            f"Component name {repo} cannot be uniquely mapped — "
            "pass the full repository name.",
            fg="red",
        )
        sys.exit(1)
    for component, pull_request in _get_prs_from_issue(repo, issue_number):
        ...
```

(Add `find_standard_component_name` to the existing `from reana.reana_dev.utils import (...)` block.)

This mirrors what `-b` already does and keeps the two flags consistent.

reana/reana_dev/git.py
```diff
@@ -794,61 +794,200 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
             )
 
 
+def _get_prs_from_issue(repo, issue_number):
+    """Return (component, pr_number) pairs for all PRs linked to a GitHub issue.
+
+    Uses GitHub's issue timeline API to find cross-referenced events, which are
+    created when a PR references an issue (e.g. "Closes #123").
+    """
+    display_message(f"Searching for PRs linked to reanahub/{repo}#{issue_number}...")
```
tiborsimko: 🤖 **Drop the "Searching for PRs..." status line (NIT)**

This line renders inconsistently with the rest of `reana-dev`:

```
[2026-05-13T11:18:30] : Searching for PRs linked to reanahub/reana-server#774...
                       ↑ missing component — display_message was called without one
```

…and the `gh api` call that follows is `display=False`, so unlike every other line in `reana-dev` output, the user doesn't see the underlying command being run. Fixing both (pass `repo` as component + flip `display=True`) would just produce two lines of plumbing for one informational ping.

Cleaner: drop the `display_message` entirely. The empty-result case is already covered by the yellow `"No PRs found linked to …"` warning, and the populated case is immediately followed by per-component `[timestamp] component: git checkout pr-…` lines, which is the information the user actually needs.

```python
def _get_prs_from_issue(repo, issue_number):
    """Return (component, pr_number) pairs for all PRs linked to a GitHub issue.

    Uses GitHub's issue timeline API to find cross-referenced events, which are
    created when a PR references an issue (e.g. "Closes #123").
    """
    # ← drop the display_message call

    output = run_command(
        f"gh api repos/reanahub/{repo}/issues/{issue_number}/timeline --paginate",
        display=False,
        return_output=True,
    )
    ...
```

tiborsimko (approved):

PR #953, opened 2026-05-19
feat(helm): add controller uWSGI ConfigMaps

### LINKED ISSUE #952: feat: finish migration to uwsgi and config maps

CameronMcClymont:
Add Helm-managed uWSGI ConfigMaps for reana-workflow-controller and
reana-job-controller.

Mount the workflow-controller configuration in its deployment and use it
for production startup. Create the job-controller configuration in both
infrastructure and runtime namespaces so that run-batch pods can mount
it.

Expose controller uWSGI worker, logging, recycling, and memory settings
through chart values. Rename the existing reana-server configuration
template to distinguish the three component-specific ConfigMaps.

Supports reanahub/reana-workflow-controller#683 and
reanahub/reana-job-controller#517.

Closes reanahub/reana#952

### DIFF
--- a/helm/reana/templates/reana-workflow-controller.yaml
+++ b/helm/reana/templates/reana-workflow-controller.yaml
@@ -57,6 +57,7 @@ spec:
         checksum/gitlab-secrets: {{ include (print $.Template.BasePath "/gitlab-secrets.yaml") . | sha256sum }}
         checksum/app-secrets: {{ include (print $.Template.BasePath "/app-secrets.yaml") . | sha256sum }}
         checksum/message-broker-secrets: {{ include (print $.Template.BasePath "/message-broker-secrets.yaml") . | sha256sum }}
+        checksum/uwsgi-config: {{ include (print $.Template.BasePath "/uwsgi-config-reana-workflow-controller.yaml") . | sha256sum }}
         {{- if $opensearchEnabled }}
         checksum/opensearch-secrets: {{ include (print $.Template.BasePath "/opensearch-secrets.yaml") . | sha256sum }}
         {{- end }}
@@ -77,6 +78,9 @@ spec:
         args: ["flask run --host=0.0.0.0"]
         tty: true
         stdin: true
+        {{- else }}
+        # do not use "/bin/sh -c" here as that breaks signal propagation
+        command: ["uwsgi", "--ini", "/var/reana/uwsgi/uwsgi.ini"]
         {{- end }}
         volumeMounts:
           {{- if .Values.debug.enabled }}
@@ -85,6 +89,8 @@ spec:
           {{- end }}
           - mountPath: {{ .Values.shared_storage.shared_volume_mount_path }}
             name: reana-shared-volume
+          - name: uwsgi-config-reana-workflow-controller
+            mountPath: '/var/reana/uwsgi'
           {{- range $workspace_paths := .Values.workspaces.paths }}
           {{- if ne (typeOf $workspace_paths) "string" -}}
           {{- fail "The workspaces.paths should be a list of strings e.g node_host_path:pod_mount_path" -}}
@@ -410,6 +416,10 @@ spec:
             - key: ca.crt
               path: ca.crt
       {{- end }}
+      - name: uwsgi-config-reana-workflow-controller
+        configMap:
+          defaultMode: 420
+          name: uwsgi-config-reana-workflow-controller
       {{- if .Values.debug.enabled }}
       - name: reana-code
         hostPath:
--- a/helm/reana/templates/uwsgi-config-reana-job-controller.yaml
+++ b/helm/reana/templates/uwsgi-config-reana-job-controller.yaml
@@ -1,0 +1,56 @@
+{{- $namespaces := list .Release.Namespace }}
+{{- if and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace) }}
+{{- $namespaces = append $namespaces .Values.namespace_runtime }}
+{{- end }}
+{{- range $namespaces }}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: uwsgi-config-reana-job-controller
+  namespace: {{ . }}
+data:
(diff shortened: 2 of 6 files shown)


helm/reana/templates/reana-workflow-controller.yaml
```diff
@@ -57,6 +57,7 @@ spec:
         checksum/gitlab-secrets: {{ include (print $.Template.BasePath "/gitlab-secrets.yaml") . | sha256sum }}
         checksum/app-secrets: {{ include (print $.Template.BasePath "/app-secrets.yaml") . | sha256sum }}
         checksum/message-broker-secrets: {{ include (print $.Template.BasePath "/message-broker-secrets.yaml") . | sha256sum }}
+        checksum/uwsgi-config: {{ include (print $.Template.BasePath "/uwsgi-config-reana-workflow-controller.yaml") . | sha256sum }}
```
michaelbuchar, line 60: ### AI-aided review

reana-server mounts uwsgi-config-reana-server, but the server Deployment does not include a checksum annotation for that ConfigMap. Future helm upgrade changes to components.reana_server.uwsgi.* will update the ConfigMap but will not restart the uWSGI process, so the new settings may not take effect until a manual rollout. Add a checksum next to the existing server annotations, mirroring R-W-C’s new checksum/uwsgi-config [reana-server.yaml#L31-L44](https://github.com/reanahub/reana/blob/7e81d4646eca885c88ded3423fb1dba034f36105/helm/reana/templates/reana-server.yaml#L31-L44)

helm/reana/templates/uwsgi-config-reana-server.yaml
```diff
@@ -34,10 +34,11 @@ data:
...
 
     # Workers management
-    max-requests = 1999
+    idle = {{ .Values.components.reana_server.uwsgi.idle }}
+    max-requests = {{ .Values.components.reana_server.uwsgi.max_requests }}
     max-requests-delta = 149
-    # max-worker-lifetime = 3600 # https://github.com/unbit/uwsgi/issues/1894
-    reload-on-rss = 250
+    max-worker-lifetime = {{ .Values.components.reana_server.uwsgi.max_worker_lifetime }}
+    reload-on-rss = {{ .Values.components.reana_server.uwsgi.reload_on_rss }}
```
michaelbuchar, line 41: max_worker_lifetime: 3600 is re-enabled by default for server/RWC/RJC. The old server config explicitly had this disabled with a comment pointing to an upstream uWSGI issue, but this PR removes that guard and applies the setting to all three configs. With multi-worker components, workers can become eligible for reload together every hour unless lifetime reloads are staggered. I’d either keep it disabled by default or add an explicit lifetime delta/jitter setting.
Refs: [values.yaml#L122](https://github.com/reanahub/reana/blob/7e81d4646eca885c88ded3423fb1dba034f36105/helm/reana/values.yaml#L122), [uwsgi-config-reana-server.yaml#L40](https://github.com/reanahub/reana/blob/7e81d4646eca885c88ded3423fb1dba034f36105/helm/reana/templates/uwsgi-config-reana-server.yaml#L40), old disabled line [uwsgi-config.yaml#L39](https://github.com/reanahub/reana/blob/fe71ab361986f3e215e1bf0509886e4e0ce8f8c9/helm/reana/templates/uwsgi-config.yaml#L39). uWSGI documents max-worker-lifetime as reloading workers after the configured seconds: [uWSGI options](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#max-worker-lifetime).

helm/reana/values.yaml
```diff
@@ -117,6 +117,10 @@ components:
       log_all: true
       log_4xx: true
       log_5xx: true
+      idle: 60
```
michaelbuchar: idle: 60 is also added by default. In uWSGI this means entering cheap mode after inactivity, not just a harmless worker-management metadata value. That can cold-start server/RWC/RJC after 60 seconds idle, which is a behavior change from the existing server config and the old RWC Dockerfile command. I’d remove the default unless the team explicitly wants cheap-mode behavior.
Ref: [values.yaml#L120](https://github.com/reanahub/reana/blob/7e81d4646eca885c88ded3423fb1dba034f36105/helm/reana/values.yaml#L120), uWSGI docs for idle: [uWSGI options](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#idle).

helm/reana/templates/uwsgi-config-reana-job-controller.yaml
```diff
@@ -0,0 +1,56 @@
+{{- $namespaces := list .Release.Namespace }}
+{{- if .Values.namespace_runtime }}
+{{- $namespaces = append $namespaces .Values.namespace_runtime }}
+{{- end }}
```
michaelbuchar, line 4: Minor edge case: uwsgi-config-reana-job-controller.yaml appends .Values.namespace_runtime without checking whether it equals .Release.Namespace. If someone explicitly sets namespace_runtime to the release namespace, Helm renders the same ConfigMap twice in the same namespace. Add a ne .Values.namespace_runtime .Release.Namespace guard or dedupe the list.

CameronMcClymont: Oh yeah nice catch. Added the guard as suggested.

helm/reana/templates/uwsgi-config-reana-server.yaml
CameronMcClymont: Updated so `max-worker-lifetime` is not included by default, as we had before.

helm/reana/templates/reana-workflow-controller.yaml
CameronMcClymont: Nice catch, added for rwc now.

helm/reana/values.yaml
CameronMcClymont: As above, updated to be 0 by default.

michaelbuchar (dismissed):
Helm logic looks good!

PR #955, opened 2026-05-19
feat(helm): support separate runtime namespace for workflow pods

### LINKED ISSUE #954: feat(helm): add separate runtime namespace for workflow pods

CameronMcClymont:
Add the namespace_runtime option to run workflow pods separately from
REANA infrastructure services. Create and manage the runtime namespace,
service account, RBAC bindings, and Kerberos configuration, and use
namespace-qualified database service names for cross-namespace access.

Support hostpath and chart-managed static Manila CephFS storage. For
CephFS, create namespace-local PVCs backed by distinct PV objects that
point to the same share. Reject incompatible NFS and custom StorageClass
configurations when a separate runtime namespace is enabled.

Remove unreachable pre-Kubernetes-1.21 Manila provisioning branches and
their orphaned configuration values. Document namespace ownership,
lifecycle, network-isolation, and secret-management constraints, and add
Helm rendering coverage for supported and rejected configurations.

Closes #954

Co-authored-by: Tibor Šimko <tibor.simko@cern.ch>


helm/reana/templates/reana-shared-persistent-volume.yaml
```diff
@@ -43,24 +89,26 @@ spec:
...
       shareAccessID: {{ .Values.shared_storage.cephfs.cephfs_os_share_access_id }}
-{{- else }}
+{{- end }}
 ---
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
+# Identically-named PVC in the runtime namespace so that batch pods can mount
+# the same shared storage without cross-namespace PVC references.
+apiVersion: v1
+kind: PersistentVolumeClaim
```
michaelbuchar, line 86: NFS runtime pods may mount a different workspace. For NFS, [reana-shared-persistent-volume.yaml](https://github.com/reanahub/reana/blob/6acd744a1dd00a6f599a471fb04150d15d8e55eb/helm/reana/templates/reana-shared-persistent-volume.yaml#L93-L112) creates a second PVC using the same StorageClass. A dynamically provisioned claim normally receives its own PV; PVC-to-PV bindings are one-to-one. The infrastructure and workflow pods can therefore see different exports.

Modern CephFS handles this correctly by creating two PVs that target the same share. NFS needs equivalent backend-specific handling or an explicit unsupported-backend failure. See Kubernetes [persistent volume binding](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#binding) and [dynamic provisioning](https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/).

helm/reana/values.yaml
```diff
@@ -167,6 +190,11 @@ ingress:
     hosts: []
   extra: []
 
+# Optionally, use a separate namespace for user workflow pods (run-batch, run-job, run-session).
+# The namespace, runtime service account, and RBAC bindings are created automatically.
+# For network storage backends (cephfs/nfs), a shared PVC is also created in this namespace.
+namespace_runtime: reana-runtime
+
```
michaelbuchar: The fixed chart-owned reana-runtime namespace is an unsafe default. This enables reana-runtime for every installation, while [serviceaccount.yaml](https://github.com/reanahub/reana/blob/6acd744a1dd00a6f599a471fb04150d15d8e55eb/helm/reana/templates/serviceaccount.yaml#L6-L17) makes Helm own that Namespace. This can conflict with an existing namespace or a second REANA release and risks deleting shared namespace contents on uninstall. Prefer an empty opt-in default, a release-scoped default, or a separate createRuntimeNamespace switch.

helm/reana/templates/kerberos-config.yaml
```diff
@@ -1,9 +1,14 @@
+{{- $namespaces := list .Release.Namespace }}
+{{- if .Values.namespace_runtime }}
+{{- $namespaces = append $namespaces .Values.namespace_runtime }}
+{{- end }}
+{{- range $namespaces }}
```
michaelbuchar, line 5: Explicit same-namespace configuration renders duplicate resources. Rendering with --namespace infra --set namespace_runtime=infra produces duplicate Kerberos ConfigMaps from [kerberos-config.yaml](https://github.com/reanahub/reana/blob/6acd744a1dd00a6f599a471fb04150d15d8e55eb/helm/reana/templates/kerberos-config.yaml#L1-L5), duplicate network PVCs, and a chart-managed Namespace/infra. Guard all runtime copies with: and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace)

michaelbuchar:
One more thing: the documentation is stale. [README.md](https://github.com/reanahub/reana/blob/6acd744a1dd00a6f599a471fb04150d15d8e55eb/helm/reana/README.md#L237) still documents `.Release.Namespace` as the default, while the chart now defaults to `reana-runtime`.

helm/reana/templates/kerberos-config.yaml
CameronMcClymont: Easy enough to add the guards so I've pushed this change, but probably unlikely that an admin would configure the infra namespace and the runtime namespace to have the exact same name?

helm/reana/values.yaml
CameronMcClymont: Good point, updated to be opt-in (empty string rather than 'reana-runtime' by default).

helm/reana/templates/reana-shared-persistent-volume.yaml
CameronMcClymont: Added an explicit failure and error message now.

michaelbuchar:
I think one same-namespace edge case is still broken.

Rendering with `--namespace infra --set namespace_runtime=infra` avoids the duplicate runtime resources now, but `roles.yaml` still adds a ClusterRoleBinding subject for `<prefix>-runtime`, and `reana-workflow-controller.yaml` still sets `REANA_RUNTIME_KUBERNETES_SERVICEACCOUNT_NAME=<prefix>-runtime`. `serviceaccount.yaml` only creates that runtime ServiceAccount when `namespace_runtime != .Release.Namespace`, so this configuration points workflow pods at a missing ServiceAccount.

Can we use the same guard in roles.yaml and reana-workflow-controller.yaml:
`and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace)`
or intentionally create the runtime ServiceAccount even for same-namespace config?

michaelbuchar:
There are some merge conflicts that need fixing.. :/

helm/reana/values.yaml
```diff
@@ -167,6 +190,12 @@ ingress:
     hosts: []
   extra: []
 
+# Optionally, use a separate namespace for user workflow pods (run-batch, run-job, run-session).
+# When set, the chart creates the namespace, runtime service account, and RBAC bindings
+# automatically; for network storage backends (cephfs/nfs) a shared PVC is also created there.
```
michaelbuchar: The new namespace runtime comment is stale. It says the runtime PVC is created for cephfs/nfs, but the chart now explicitly rejects shared_storage.backend=nfs when namespace_runtime is set. I verified helm template fails with the new unsupported-backend error in that configuration, so the comment should be narrowed to CephFS only or explicitly say NFS is unsupported.

CameronMcClymont: Updated now.

michaelbuchar (dismissed):
LGTM!

helm/reana/values.yaml
```diff
@@ -216,6 +216,12 @@ ingress:
     hosts: []
   extra: []
 
+# Optionally, use a separate namespace for user workflow pods (run-batch, run-job, run-session, run-dask).
```
tiborsimko: **🤖 `IMAGE_PULL_SECRETS` not replicated to the runtime namespace (MEDIUM)**

Kubernetes Secrets are namespace-scoped. When an operator configures private-registry credentials via:

- `components.reana_job_controller.environment.IMAGE_PULL_SECRETS`
- `components.reana_workflow_controller.environment.IMAGE_PULL_SECRETS`

the job-controller injects them by name into every job pod spec at `reana-job-controller/reana_job_controller/kubernetes_job_manager.py:466-470`:

```python
for secret_name in current_app.config["IMAGE_PULL_SECRETS"]:
    image_pull_secrets.append({"name": secret_name})
self.job["spec"]["template"]["spec"]["imagePullSecrets"] = image_pull_secrets
```

With `namespace_runtime` set, job pods (and Dask worker/scheduler pods) run in the runtime namespace, but the referenced Secrets must also exist in that namespace — otherwise workflows using private images fail with `ImagePullBackOff` on the runtime-namespace pods.

User secrets (`UserSecretsStore` in `reana-commons/reana_commons/k8s/secrets.py`) are already namespace-aware — they read/write in `REANA_RUNTIME_KUBERNETES_NAMESPACE`. The gap is specifically the admin-provided pull credentials, which the chart doesn't manage at all.

Suggested fix — document the requirement next to `namespace_runtime` in `helm/reana/values.yaml` and the corresponding README row:

```yaml
# Optionally, use a separate namespace for user workflow pods (run-batch,
# run-job, run-session, run-dask). When set, the chart creates the namespace,
# runtime service account, and RBAC bindings automatically; for cephfs on
# Kubernetes >= 1.21 a shared PV+PVC is also created there.
# NOTE: any Secret listed in IMAGE_PULL_SECRETS must be created manually
# in this namespace as well; the chart does not replicate it.
# Leave empty (default) to run workflow pods in the same namespace as the
# REANA services.
```

A more ambitious fix would teach the chart to replicate (or `kubectl create secret docker-registry`-style provision) the listed pull secrets into both namespaces, but that's a larger change beyond this PR's scope.

helm/reana/templates/roles.yaml
```diff
@@ -48,7 +48,7 @@ subjects:
 - kind: ServiceAccount
   name: {{ include "reana.prefixed_infrastructure_svaccount_name" . }}
   namespace: {{ .Release.Namespace }}
-{{- if .Values.namespace_runtime }}
+{{- if and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace) }}
```
tiborsimko, line 51: **🤖 Runtime ServiceAccount could be scoped to a Role (LOW)**

`helm/reana/templates/roles.yaml` binds the runtime ServiceAccount to the same `<prefix>-deployment-manager` **ClusterRole** as the infrastructure ServiceAccount:

```yaml
- kind: ServiceAccount
  name: {{ include "reana.prefixed_runtime_svaccount_name" . }}
  namespace: {{ .Values.namespace_runtime }}
```

That ClusterRole carries cluster-wide `get/list/create/update/watch` on nodes, secrets, deployments, services, ingresses and `*` on `storageclasses` — broader than what any pod in the runtime namespace actually needs.

In practice the blast radius is narrow: user-step job pods (`reana-job-controller/kubernetes_job_manager.py:204`) and interactive sessions (`reana-workflow-controller/k8s.py:92`) set `automountServiceAccountToken: False`, so they never carry the SA token. Only the trusted `run-batch` orchestrator pod uses the runtime SA, and its workload is REANA-controlled code, not arbitrary user code. So this is not a "compromised user workload escapes the namespace" issue, just a least-privilege one.

Suggested fix — out of scope for this PR; worth a follow-up issue to split the ClusterRole into:
- a true ClusterRole for genuinely cluster-scoped operations (nodes/metrics, cluster-wide CR watches, storageclasses);
- a Role bound per namespace for pod/service/configmap CRUD.

Bind the runtime SA only to the per-namespace Role plus the minimal ClusterRole subset it actually needs. The wide ClusterRole is also what currently makes Dask + Traefik plumbing "just work" cross-namespace, so any narrowing must keep those flows green.

helm/reana/templates/serviceaccount.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ include "reana.prefixed_infrastructure_svaccount_name" . }}
-{{- if .Values.namespace_runtime }}
+{{- if and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace) }}
```
tiborsimko, line 6: **🤖 `helm uninstall` cascade on the runtime namespace not documented (NIT)**

`helm/reana/templates/serviceaccount.yaml:6-12` makes Helm the owner of the runtime Namespace:

```yaml
{{- if and .Values.namespace_runtime (ne .Values.namespace_runtime .Release.Namespace) }}
---
apiVersion: v1
kind: Namespace
metadata:
  name: {{ .Values.namespace_runtime }}
```

`helm uninstall` of the REANA release therefore deletes the runtime namespace, which cascades to every resource inside — not only the chart-managed PVC and ServiceAccount, but anything users or other charts placed there after install. Helm refuses to adopt pre-existing unmanaged resources at install time, so the "shared namespace at install" risk is moot; the surviving concern is the post-install footprint that grows organically (debug pods, ad-hoc Secrets, sidecar workloads) and gets swept on uninstall.

Suggested fix — add a one-line warning to the values.yaml comment block:

```yaml
# Optionally, use a separate namespace for user workflow pods (run-batch,
# run-job, run-session, run-dask). When set, the chart creates and owns
# the namespace; `helm uninstall` will delete it and every resource
# inside, so do not place unrelated workloads there.
```

The corresponding README row should mirror the warning.

helm/reana/values.yaml
CameronMcClymont: Updated comment and README.

helm/reana/templates/serviceaccount.yaml
CameronMcClymont: Updated comment (tried to fuse it with the other suggestion) and updated README.

helm/reana/templates/roles.yaml
CameronMcClymont: Created an [issue for this](https://github.com/reanahub/reana/issues/975) since the security benefits would be small and it would be nice to get this PR merged.

helm/reana/values.yaml
CameronMcClymont: Or would you rather go for the more 'ambitious fix'?

tiborsimko: The documentation-only fix is sufficient here; no need to implement automatic replication. IMAGE_PULL_SECRETS refers to operator-managed, namespace-scoped Secrets whose contents and lifecycle are outside the chart’s ownership. Operators can provision them in both namespaces directly or through tools such as External Secrets.

I’ll take care of rebasing this branch onto current master. The shared-storage conflict needs a little care because #984 introduced shared_storage.storage_class_name: a mechanical resolution could mix a dynamically provisioned infrastructure PVC with the static Manila- CephFS runtime PV from this PR. On a Rook-CephFS cluster without the Manila driver, the runtime PVC would remain unbound and workflows would stay Pending. I’ll preserve the current master behaviour and reject namespace_runtime together with a custom storage_class_name, while retaining the supported static CephFS and hostpath cases.

I’ll also add chart-render tests covering the relevant storage and namespace combinations, standard ownership labels on the generated Namespace, and a small wording pass describing this as administrative/topological namespace separation rather than network isolation. NetworkPolicies, projected run-batch tokens and narrower runtime RBAC will remain separate follow-ups.

Since this requires rebasing, I’ll force-push the updated history to reana-namespaces on your fork. I’ll keep my reconciliation as a separate fixup commit on top of your rebased original so the delta remains easy to review; it can be squashed before merge while retaining your authorship and crediting the fixup with Co-authored-by.

Let’s keep automatic secret provisioning outside this PR and track it separately if operational experience shows it is needed. Thanks!

tiborsimko: Done as described. I rebased the branch onto current `master` and force-pushed the updated history. The reconciliation is kept in separate fixup commit `2142d17`, which preserves current storage behaviour, rejects unsupported runtime-namespace storage combinations, retains static Manila CephFS and hostpath support, adds runtime Namespace labels and chart-render coverage, and clarifies that this provides administrative separation rather than network isolation.

`IMAGE_PULL_SECRETS` remain operator-provisioned, while NetworkPolicies, projected tokens and narrower runtime RBAC remain separate follow-ups. I also updated the PR description accordingly. The fixup can be squashed into the original commit before merge, preserving the original authorship and crediting the fixup with `Co-authored-by`.

helm/reana/templates/reana-shared-persistent-volume.yaml
```diff
@@ -1,4 +1,7 @@
 {{- if not (eq .Values.shared_storage.backend "hostpath") }}
+{{- $isCephFs := eq .Values.shared_storage.backend "cephfs" }}
+{{- $hasCustomStorageClass := not (empty .Values.shared_storage.storage_class_name) }}
+{{- $isStaticCephFs := and $isCephFs (not $hasCustomStorageClass) }}
```
tiborsimko, line 4: **🤖 PR955-04: Unreachable pre-1.21 branch removed from only one of the two parallel PV templates (LOW)**

The removal itself checks out. `helm/reana/Chart.yaml` declares:

```yaml
kubeVersion: ">= 1.21.0-0"
```

When REANA is installed or rendered as the top-level chart, Helm enforces that constraint before rendering, so `helm install`, `helm upgrade` and `helm template --kube-version 1.20` all fail with `chart requires kubeVersion: ...` rather than reaching the guard. The branch cannot be reached in any supported deployment, and the rewritten `{{- if $isStaticCephFs }}` is logically equivalent to the old `and $isCephFs (not $hasCustomStorageClass)` plus the nested `$isModernK8s`.

`helm/reana/templates/reana-infrastructure-persistent-volume.yaml` is a parallel implementation of the same CephFS selection logic, and it still carries the guard in two places:

```
helm/reana/templates/reana-infrastructure-persistent-volume.yaml:17
  {{- else if and (eq .Values.infrastructure_storage.backend "cephfs") (semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion) }}

helm/reana/templates/reana-infrastructure-persistent-volume.yaml:25
  {{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion }}
```

Line 25 still gates the legacy `manila-provisioner` StorageClass block — the same construct this commit removes from the shared template. The `kubeVersion` argument applies equally to both files, so the two templates now disagree about whether the version check carries any meaning, and the next person to touch either one has to re-derive the answer.

Suggested fix — apply the same treatment to `reana-infrastructure-persistent-volume.yaml`: hoist a `$isStaticCephFs`-style variable alongside the existing conditions, drop both `semverCompare` calls, and delete the `{{- else }}` StorageClass arm. That keeps the equivalent CephFS selection logic consistent across the two templates. If you do fold it in, it would be worth adding an infrastructure-CephFS rendering case to `tests/test_helm.py` and asserting that no legacy REANA StorageClass is emitted, so the supported behaviour is pinned while the cleanup happens. If you would rather keep this PR focused, a separate cleanup issue serves equally well — the file is outside this PR's diff, so this is a follow-up rather than something to block on.

helm/reana/templates/reana-shared-persistent-volume.yaml
```diff
@@ -45,24 +47,53 @@ spec:
...
-apiVersion: storage.k8s.io/v1
-kind: StorageClass
+# Kubernetes requires a 1:1 PV-to-PVC binding. This second PV uses a distinct
+# volumeHandle but points at the same underlying CephFS share as the one above.
+apiVersion: v1
+kind: PersistentVolume
 metadata:
-  name: {{ include "reana.prefix" . }}-shared-volume-storage-class
-  namespace: {{ .Release.Namespace }}
-provisioner: {{ .Values.shared_storage.cephfs.provisioner }}
```
tiborsimko, line 55: **🤖 PR955-05: Three documented `shared_storage.cephfs` values are left with no consumer (LOW)**

The legacy StorageClass block removed here was the only place in the chart that read `shared_storage.cephfs.provisioner`, `shared_storage.cephfs.type` and `shared_storage.cephfs.availability_zone`:

```yaml
provisioner: {{ .Values.shared_storage.cephfs.provisioner }}
parameters:
  type: {{ .Values.shared_storage.cephfs.type }}
  zones: {{ .Values.shared_storage.cephfs.availability_zone }}
```

After the removal, no template references any of the three, but all three remain in `values.yaml`:

```yaml
  cephfs:
    provisioner: manila-provisioner
    type: "Geneva CephFS Testing"
    availability_zone: nova
```

and all three keep their rows in `helm/reana/README.md`, still advertised as `**[CERN only]**` configuration:

| key | README description |
| --- | --- |
| `shared_storage.cephfs.availability_zone` | **[CERN only]** OpenStack Availability zone |
| `shared_storage.cephfs.provisioner` | **[CERN only]** CephFS provisioner |
| `shared_storage.cephfs.type` | **[CERN only]** CephFS availability zone |

An administrator who sets `shared_storage.cephfs.provisioner` now gets silence: the value is accepted, documented, and ignored. Configuration that still looks live is more costly than configuration that was never offered, because it invites operators to tune something with no effect and gives them no signal when it does nothing. The three values were already unreachable on every supported Kubernetes version, so dropping them changes no behaviour — it just makes the value surface honest about it.

Suggested fix — in the same commit that removes the branch, drop the three keys from the `shared_storage.cephfs` block in `values.yaml` and delete their rows from the README configuration table, so the value surface and the templates stay in step. Note that applying PR955-04 would also remove the infrastructure template's remaining references to the analogous user-supplied fields; there are no separate `infrastructure_storage.cephfs` defaults or README rows to delete, since `infrastructure_storage` defaults to an empty mapping.

helm/reana/templates/reana-shared-persistent-volume.yaml
```diff
@@ -11,17 +14,16 @@ spec:
...
+  {{- else if $isStaticCephFs }}
   storageClassName: ""
   volumeName: {{ include "reana.prefix" . }}-shared-persistent-volume-storage
   {{- else }}
   storageClassName: {{ include "reana.prefix" . }}-shared-volume-storage-class
   {{- end }}
 # Which storage class to create depending on the backend
-{{- if and (eq .Values.shared_storage.backend "cephfs") (not .Values.shared_storage.storage_class_name) }}
-{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion }}
+{{- if $isStaticCephFs }}
```
tiborsimko, line 25: **🤖 PR955-06: Comment still describes a StorageClass choice that no longer exists (NIT)**

```yaml
# Which storage class to create depending on the backend
{{- if $isStaticCephFs }}
---
apiVersion: v1
kind: PersistentVolume
```

The comment was accurate while the block had two arms, one of which created a `StorageClass`. It now guards a single `PersistentVolume` and no storage class is created here at all, so the comment describes a shape the code no longer has.

Suggested fix — reword to match what is rendered, or drop the comment entirely since `$isStaticCephFs` already names the condition:

```yaml
# Static Manila CephFS volume backing the shared PVC
{{- if $isStaticCephFs }}
```

CameronMcClymont: Nice catch, took the suggested comment.

helm/reana/templates/reana-shared-persistent-volume.yaml
CameronMcClymont: Removed the dead `values.yaml` values as suggested.

helm/reana/templates/reana-shared-persistent-volume.yaml
CameronMcClymont: A yes nice catch, we can clean up the same dead guard here too.

tiborsimko (approved):

PR #956, opened 2026-05-19
feat(helm): allow customising minimum Kubernetes job user ID

### LINKED ISSUE #951: helm: allow cluster admins to configure the minimum accepted `kubernetes_uid` value
## Current behaviour

User-supplied `kubernetes_uid` workflow hint values are accepted as long as they are `>= 100`. This is hard-coded in REANA ecosystem, e.g. in the `reana-job-controller` component. A REANA deployment may want to strengthen the policy to allow only higher user IDs, similar to OpenShift.

  ## Expected behaviour

The minimum accepted `kubernetes_uid` value should be configurable per REANA deployment via the Helm value files, for example by means of an environment variable such as  `REANA_KUBERNETES_JOBS_MIN_USER_UID`.

tiborsimko:
REANA already refuses runtime container user IDs below 100 for `kubernetes_uid` workflow hints, blocking root and the statically reserved system-user range. This commit introduces a `kubernetes_jobs_min_user_uid` Helm value (defaulting to the existing minimum of 100) that propagates as the
`REANA_KUBERNETES_JOBS_MIN_USER_UID` environment variable into reana-server and reana-workflow-controller pods, so that cluster administrators can customise their deployments and, if desired, set even higher minimum runtime container user ID limits.

Closes reanahub/reana#951


michaelbuchar (commented):
LGTM, just some CI checks are still failing

michaelbuchar (approved):
LGTM!

PR #958, opened 2026-05-20
feat(reana-dev): add create-branch option to git-upgrade command

CameronMcClymont:
This commit adds a --create-branch option to the git-upgrade command. When used, it creates the branch locally from the upstream branch and pushes it to the origin, if the branch exists in upstream. This option is especially useful when upgrading maintenance branches that don't exist locally or in your fork.

Replaces https://github.com/reanahub/reana/pull/884
Co-authored-by: Alputer <alp.tuna.453@gmail.com>

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -1163,20 +1163,26 @@ def git_upgrade(component, exclude_components, create_branch, base):  # noqa: D3
         if not branch_exists(component, base):
             if create_branch:
                 # Check if upstream branch exists before creating local branch
-                cmd = f"git ls-remote --heads upstream refs/heads/{base}"
-                if not run_command(cmd, component, display=False, return_output=True):
+                if not run_command(
+                    ["git", "ls-remote", "--heads", "upstream", f"refs/heads/{base}"],
+                    component,
+                    display=False,
+                    return_output=True,
+                ):
                     display_message(
                         f"Branch {base} does not exist in upstream, skipping.",
                         component=component,
                     )
                     continue
 
-                run_command("git fetch upstream", component)
-                run_command(f"git checkout -b {base} upstream/{base}", component)
+                run_command(["git", "fetch", "upstream"], component)
+                run_command(
+                    ["git", "checkout", "-b", base, f"upstream/{base}"], component
+                )
                 try:
-                    run_command(f"git push -u origin {base}", component)
+                    run_command(["git", "push", "-u", "origin", base], component)
                 finally:
-                    run_command("git checkout -", component)
+                    run_command(["git", "checkout", "-"], component)
                 display_message(
                     f"Branch {base} created from upstream and pushed to origin.",
                     component=component,
@@ -1189,11 +1195,11 @@ def git_upgrade(component, exclude_components, create_branch, base):  # noqa: D3
             # Branch was just created fresh or is missing so skip the merge/push upgrade flow
             continue
 
-        run_command("git fetch upstream", component)
-        run_command(f"git checkout {base}", component)
-        run_command(f"git merge --ff-only upstream/{base}", component)
-        run_command(f"git push origin {base}", component)
-        run_command("git checkout -", component)
+        run_command(["git", "fetch", "upstream"], component)
+        run_command(["git", "checkout", base], component)
+        run_command(["git", "merge", "--ff-only", f"upstream/{base}"], component)
+        run_command(["git", "push", "origin", base], component)
+        run_command(["git", "checkout", "-"], component)
 
 
 @click.argument("range", default="")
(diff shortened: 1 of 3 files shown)


reana/reana_dev/git.py
```diff
@@ -1142,28 +1148,50 @@ def git_upgrade(component, exclude_components, base):  # noqa: D301
...
+
     components = select_components(component, exclude_components)
     for component in components:
         if not branch_exists(component, base):
-            display_message(
-                "Missing branch {}, skipping.".format(base), component=component
-            )
+            if create_branch:
+                # Check if upstream branch exists before creating local branch
+                cmd = f"git ls-remote --heads upstream {base}"
```
michaelbuchar: The upstream branch existence check accepts suffix matches.
Example: this runs:
```
git ls-remote --heads upstream maint-1.2
```
Git also matches refs/heads/release/maint-1.2. The code then attempts to create upstream/maint-1.2 at [git.py#L1175](https://github.com/reanahub/reana/blob/760261d2f5bb47d1334f35d2f814555078bbbc1f/reana/reana_dev/git.py#L1175), which does not exist, and aborts.

Use an exact ref:
`
git ls-remote --exit-code --heads upstream refs/heads/<base>`

reana/reana_dev/git.py
```diff
@@ -1142,28 +1148,50 @@ def git_upgrade(component, exclude_components, base):  # noqa: D301
...
+                    display_message(
+                        f"Branch {base} does not exist in upstream, skipping.",
+                        component=component,
+                    )
+                    continue
+
+                run_command("git fetch upstream", component)
+                run_command(f"git checkout -b {base} upstream/{base}", component)
+                run_command(f"git push -u origin {base}", component)
+                run_command("git checkout -", component)
```
michaelbuchar: A rejected push leaves the repository on the newly created branch. The code creates the branch, pushes it, and only then restores the previous checkout at [git.py#L1175-L1177](https://github.com/reanahub/reana/blob/760261d2f5bb47d1334f35d2f814555078bbbc1f/reana/reana_dev/git.py#L1175-L1177). run_command() exits immediately on failure in [utils.py#L380-L388](https://github.com/reanahub/reana/blob/760261d2f5bb47d1334f35d2f814555078bbbc1f/reana/reana_dev/utils.py#L380-L388). I reproduced this with a divergent origin/maint-1.2: the push failed and the working checkout remained on maint-1.2. Restore the original branch in a finally block.

michaelbuchar: Branch names are interpolated into shell=True commands without quoting. New commands here break valid Git refs containing shell metacharacters. I verified that Git accepts `maint;echo`, while the PR executes it as multiple shell commands. This also creates a command-injection path if --base is passed through automation. We should prefer argument arrays with shell=False. The existing upgrade flow at lines 1191-1193 should be corrected at the same time.

michaelbuchar:
There is also a testing gap we can address or put in a separate PR. Codecov reports only 11.11% patch coverage. We can add regression tests for: 
1. Exact upstream branch creation.
2. Suffix-only upstream branch rejection.
3. Missing upstream branch.
4. Divergent origin push with checkout restoration.
5. Branch names requiring shell-safe handling.

reana/reana_dev/git.py
CameronMcClymont: Used an exact ref as suggested so it doesn't accidentally match a tag named the same as your branch. The `--exit-code` addition is unnecessary and introduces problems.

reana/reana_dev/git.py
CameronMcClymont: Thanks, added try/finally so `git checkout` runs whether or not the push succeeds.

CameronMcClymont: Kind of an edge case for handling e.g. refs with spaces, but it's a nice option to have in `run_command` so I've added it.

michaelbuchar (dismissed):
LGTM!

michaelbuchar:
Just a rebase is needed before we merge this

tiborsimko:
Thanks, I have improved some small nits whilst merging:

  - Added a regression test for the existing-branch merge/push flow, so the refactor from shell strings to list form has explicit coverage.
  - Switched the `run_command` display to `shlex.join` so list-form commands log faithfully when arguments contain spaces or metacharacters.
  - Tidied the `git-upgrade` loop so all five commands use the list form consistently — no behaviour change, just less mixed style.
  - Widened the type hint to `Union[str, Sequence[str]]` so tuples and other sequences are accepted too.
  - Dropped `test_suffix_only_upstream_branch_rejected` as it overlapped with `test_missing_upstream_branch_skipped`; the refspec format is already asserted in `test_create_branch_exact_upstream`.

PR #960, opened 2026-06-02
feat(reana-dev): add git-iteration-rollover command

CameronMcClymont:
Adds a reana-dev command to migrate all unfinished issues and PRs from
the previous iteration of a GitHub Projects V2 board to the current one.
Items in the Done column are skipped; everything else keeps its Status
(Backlog, Ready for work, In work, etc.).

Source and target iterations are auto-detected as the most recently
completed and earliest active iterations, respectively. Use the override
options `--from` and `--to` when needed. Move a single item with the
`--issue` option, or use `--dry-run` to report changes without writing.

Name-based project lookup errors when more than one project shares the
title. When `--issue` is used, missing items, items not assigned to any
iteration, and items in a different iteration are reported as errors.
Project field names are resolved using their exact casing so Done items
are reliably skipped.

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -2459,4 +2459,641 @@ def get_aggregate_changelog(  # noqa: C901
         )
 
 
+def _run_graphql(query, variables=None):
+    """Execute a GitHub GraphQL query via gh CLI and return the response data.
+
+    :param query: GraphQL query or mutation string.
+    :param variables: Optional dict of GraphQL variables.
+    :raises SystemExit: on API error or non-zero exit code.
+    :return: Parsed ``data`` field from the GraphQL response.
+    :rtype: dict
+    """
+    payload = {"query": query}
+    if variables:
+        payload["variables"] = variables
+    try:
+        result = subprocess.run(
+            ["gh", "api", "graphql", "--input", "-"],
+            input=json.dumps(payload),
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError as err:
+        click.secho(f"GitHub GraphQL API error: {err.stderr}", fg="red")
+        sys.exit(1)
+    data = json.loads(result.stdout)
+    if "errors" in data:
+        for error in data["errors"]:
+            click.secho(f"GraphQL error: {error.get('message', error)}", fg="red")
+        sys.exit(1)
+    return data["data"]
+
+
+def _find_project_number_by_name(org, name):
+    """Return the project number whose title matches *name* exactly.
+
+    Scans all Projects V2 in *org* to detect ambiguous titles. GitHub allows
+    multiple projects to share the same title; in that case the user must
+    pass the numeric project number to disambiguate, since picking the first
+    match could silently target the wrong project for this destructive
+    command.
+
+    :param org: GitHub organisation login.
+    :param name: Exact project title to look for.
+    :raises SystemExit: when no matching project is found, or when more than
+        one project shares the title.
+    :return: Project number.
+    :rtype: int
+    """
+    query = """
+    query($org: String!, $cursor: String) {
+      organization(login: $org) {
+        projectsV2(first: 50, after: $cursor) {
+          pageInfo { hasNextPage endCursor }
+          nodes { number title }
+        }
+      }
+    }
+    """
+    matches = []
+    cursor = None
+    while True:
+        variables = {"org": org}
+        if cursor:
+            variables["cursor"] = cursor
+        data = _run_graphql(query, variables)
+        projects = data["organization"]["projectsV2"]
+        for project in projects["nodes"]:
+            if project["title"] == name:
+                matches.append(project["number"])
+        if not projects["pageInfo"]["hasNextPage"]:
+            break
+        cursor = projects["pageInfo"]["endCursor"]
+    if not matches:
+        click.secho(f"Project '{name}' not found in organisation {org}.", fg="red")
+        sys.exit(1)
+    if len(matches) > 1:
+        numbers = ", ".join(str(n) for n in matches)
+        click.secho(
+            f"Multiple projects in {org} are titled '{name}' (numbers: {numbers}). "
+            f"Re-run with --project NUMBER to disambiguate.",
+            fg="red",
+        )
+        sys.exit(1)
+    return matches[0]
+
+
+def _get_project_info(org, project_number):
+    """Return project id and fields for a GitHub Projects V2 project.
+
+    :param org: GitHub organisation login.
+    :param project_number: Project number within the organisation.
+    :return: Project node with ``id`` and ``fields``.
+    :rtype: dict
+    """
+    query = """
+    query($org: String!, $number: Int!) {
+      organization(login: $org) {
+        projectV2(number: $number) {
+          id
+          fields(first: 100) {
+            nodes {
+              ... on ProjectV2Field { id name }
+              ... on ProjectV2SingleSelectField {
+                id
+                name
+                options { id name }
+              }
+              ... on ProjectV2IterationField {
+                id
+                name
+                configuration {
+                  iterations { id title startDate duration }
+                  completedIterations { id title startDate duration }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+    """
+    data = _run_graphql(query, {"org": org, "number": project_number})
+    project = data["organization"]["projectV2"]
+    if not project:
+        click.secho(
+            f"Project #{project_number} not found in organisation {org}.", fg="red"
+        )
+        sys.exit(1)
+    return project
+
+
+def _get_project_items_page(project_id, iter_field_name, status_field_name, after=None):
+    """Fetch one page of items from a GitHub Projects V2 project.
+
+    Each item node contains ``statusValue`` (the single-select value for
+    *status_field_name*) and ``iterValue`` (the iteration value for
+    *iter_field_name*), queried by name to avoid fieldValues connection
+    truncation. Both field names must match the project's actual casing —
+    ``fieldValueByName`` is case-sensitive.
+
+    :param project_id: Node ID of the project.
+    :param iter_field_name: Name of the iteration field.
+    :param status_field_name: Name of the Status single-select field.
+    :param after: Pagination cursor (``endCursor`` from a previous page).
+    :return: ``items`` connection dict with ``pageInfo`` and ``nodes``.
+    :rtype: dict
+    """
+    query = """
+    query($id: ID!, $iterFieldName: String!, $statusFieldName: String!, $after: String) {
+      node(id: $id) {
+        ... on ProjectV2 {
+          items(first: 100, after: $after) {
+            pageInfo { hasNextPage endCursor }
+            nodes {
+              id
+              content {
+                ... on Issue {
+                  number
+                  title
+                  repository { name }
+                }
+                ... on PullRequest {
+                  number
+                  title
+                  repository { name }
+                }
+              }
+              statusValue: fieldValueByName(name: $statusFieldName) {
+                ... on ProjectV2ItemFieldSingleSelectValue { optionId }
+              }
+              iterValue: fieldValueByName(name: $iterFieldName) {
+                ... on ProjectV2ItemFieldIterationValue { iterationId }
+              }
+            }
+          }
+        }
+      }
+    }
+    """
+    variables = {
+        "id": project_id,
+        "iterFieldName": iter_field_name,
+        "statusFieldName": status_field_name,
+    }
+    if after:
+        variables["after"] = after
+    data = _run_graphql(query, variables)
+    return data["node"]["items"]
+
+
+def _update_item_iteration(project_id, item_id, field_id, iteration_id):
+    """Update the iteration field value of a GitHub Projects V2 item.
+
+    :param project_id: Node ID of the project.
+    :param item_id: Node ID of the project item.
+    :param field_id: Node ID of the iteration field.
+    :param iteration_id: ID of the target iteration.
+    """
+    mutation = """
+    mutation($projectId: ID!, $itemId: ID!, $fieldId: ID!, $iterationId: String!) {
+      updateProjectV2ItemFieldValue(input: {
+        projectId: $projectId
+        itemId: $itemId
+        fieldId: $fieldId
+        value: { iterationId: $iterationId }
+      }) {
+        projectV2Item { id }
+      }
+    }
+    """
+    _run_graphql(
+        mutation,
+        {
+            "projectId": project_id,
+            "itemId": item_id,
+            "fieldId": field_id,
+            "iterationId": iteration_id,
+        },
+    )
+
+
+def _parse_issue_filter(issue):
+    """Parse the ``--issue`` option into ``(repo, number)`` components.
+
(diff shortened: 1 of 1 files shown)


reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,529 @@ def get_aggregate_changelog(  # noqa: C901
...
+    :return: The Status field ID and the set of option IDs whose name is "done".
+    :rtype: tuple[str | None, set]
+    """
+    for field in fields:
+        if field.get("name", "").lower() == "status" and "options" in field:
+            done_ids = {
+                opt["id"] for opt in field["options"] if opt["name"].lower() == "done"
+            }
+            return field["id"], done_ids
+    return None, set()
```
michaelbuchar: Status discovery fails open and can migrate completed work. This returns an empty set if the Status field or Done option is missing. [_migrate_items()](https://github.com/reanahub/reana/blob/e01207c53a3d3f2ddde0403315c9b1beda867013/reana/reana_dev/git.py#L2824-L2836) then migrates every source-iteration item. Abort with a clear error unless Status and Done are resolved.

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,529 @@ def get_aggregate_changelog(  # noqa: C901
...
+    help="Title of the target iteration to move items to. "
+    "Auto-detected as the first active iteration if not provided.",
+)
+@click.option(
+    "--issue",
+    "-i",
+    default=None,
+    help="Move only this specific issue or PR. "
+    "Format: REPO#NUMBER (e.g. reana-server#123) or plain NUMBER to match any repo.",
+)
```
michaelbuchar, line 3005: --issue 123 can modify multiple repositories. The option explicitly matches any repository. Issue numbers commonly collide across repositories. Maybe require REPO#NUMBER, or abort when a plain number is ambiguous?

michaelbuchar:
Nested GraphQL connections are truncated. Project fields use [fields(first: 30)](https://github.com/reanahub/reana/blob/e01207c53a3d3f2ddde0403315c9b1beda867013/reana/reana_dev/git.py#L2503-L2528), while item values use [fieldValues(first: 20)](https://github.com/reanahub/reana/blob/e01207c53a3d3f2ddde0403315c9b1beda867013/reana/reana_dev/git.py#L2568-L2581), without pagination. Larger projects can silently skip items or migrate Done items when iteration or Status values fall outside the first page. Should we paginate both connections or query required fields deterministically? Reference: GitHub’s [pagination guide](https://docs.github.com/en/graphql/guides/using-pagination-in-the-graphql-api).

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,529 @@ def get_aggregate_changelog(  # noqa: C901
...
+                  number
+                  title
+                  repository { name }
+                }
+                ... on PullRequest {
+                  number
+                  title
+                  repository { name }
+                }
+              }
```
michaelbuchar, line 2627: Closed issues and merged or closed PRs can be migrated. This content query does not fetch lifecycle state, and selection only considers Project Status. If unfinished excludes closed content, query state and skip anything other than OPEN.

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,529 @@ def get_aggregate_changelog(  # noqa: C901
...
+        if "configuration" not in field or "iterations" not in field["configuration"]:
+            continue
+        if iteration_field is None:
+            iteration_field = field
+        else:
+            click.secho(
+                f"Warning: multiple iteration fields found, using '{iteration_field['name']}'.",
+                fg="yellow",
+            )
+            break
```
michaelbuchar, line 2743: Multiple iteration fields are handled by choosing the first one. The command warns and continues. Maybe for a mutating command, let's require an explicit field or abort?

reana/reana_dev/git.py
CameronMcClymont: Yes agreed we should enforce REPO#NUMBER, added now.

reana/reana_dev/git.py
CameronMcClymont: Unlikely the fields on the project board will ever change, but handled now anyway.

reana/reana_dev/git.py
CameronMcClymont: If I understand this correctly, you're saying that closed/merged PRs/issues in the board are included in the migration? This is fine because we always put closed/merged PRs/issues in the Done column ourselves.

reana/reana_dev/git.py
CameronMcClymont: We only have one iteration field, which is the two-week-long `Iteration` field so this is not a case we need to handle.

CameronMcClymont:
Renamed command to `reana-dev git-iteration-rollover`

reana/reana_dev/git.py
michaelbuchar: Thanks!

reana/reana_dev/git.py
michaelbuchar: Yeah, I believe it does not remove the code-level risk: the command will still migrate a closed issue or merged PR if someone forgot to move it to Done. But we do not have to address it now if you think it's not important or we can discuss with Tibor for his opinion

michaelbuchar (dismissed):
LGTM pending rebase

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,541 @@ def get_aggregate_changelog(  # noqa: C901
...
+    cursor = None
+    while True:
+        variables = {"org": org}
+        if cursor:
+            variables["cursor"] = cursor
+        data = _run_graphql(query, variables)
+        projects = data["organization"]["projectsV2"]
+        for project in projects["nodes"]:
+            if project["title"] == name:
+                return project["number"]
```
tiborsimko: **🤖 `_find_project_number_by_name` returns first match on a non-unique key (MEDIUM)**

GitHub allows two Projects V2 with identical titles in the same org, and `reanahub` currently has two `"@tiborsimko's untitled project"` entries (confirmed via `gh api graphql` against the live org). The loop here:

```python
for project in projects["nodes"]:
    if project["title"] == name:
        return project["number"]
```

returns whichever the API happens to enumerate first. Pagination order isn't a stable contract, so the same `--project NAME` invocation could target different projects on different days. For a command that bulk-rewrites the iteration field on every non-Done item in the project, that's a real footgun.

MEDIUM rather than HIGH because: (a) the two current duplicates are stub "untitled" projects nobody is likely to type by name, and (b) `--dry-run` exists and would surface the wrong target before any mutation. But the structural bug is real and easy to close.

Suggested fix — collect all matches across the full pagination, then disambiguate:

```python
matches = []
# inside the pagination loop:
for project in projects["nodes"]:
    if project["title"] == name:
        matches.append(project["number"])
# after the loop (do not early-return inside it):
if not matches:
    click.secho(f"Project '{name}' not found in organisation {org}.", fg="red")
    sys.exit(1)
if len(matches) > 1:
    click.secho(
        f"Multiple projects in {org} are titled '{name}' (numbers: "
        f"{', '.join(str(n) for n in matches)}). "
        f"Re-run with --project NUMBER to disambiguate.",
        fg="red",
    )
    sys.exit(1)
return matches[0]
```

This requires scanning all project pages instead of short-circuiting at the first match, but the org enumeration is bounded by the number of projects (50 per page) and only runs when `--project` is given by name, so the extra cost is acceptable.

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,541 @@ def get_aggregate_changelog(  # noqa: C901
...
+    for item in all_items:
+        content = item.get("content")
+        if not content:
+            continue
+
+        item_number = content.get("number")
+        item_repo = content.get("repository", {}).get("name", "")
+        item_title = content.get("title", "")
+
+        if issue_number is not None:
```
tiborsimko, line 2949: **🤖 `--issue` typo silently reports success (MEDIUM)**

In `_migrate_items` the per-issue filter and the source-iteration filter are stacked:

```python
if issue_number is not None:
    if item_number != issue_number or item_repo != issue_repo:
        continue
...
if item_iter_id != prev_iter["id"]:
    continue
```

If the user runs `git-iteration-rollover --project ... --issue reana-server#123` and the item exists in the project but is currently assigned to a different iteration (or to no iteration), the loop falls through both `continue`s and `moved_count == skipped_count == 0`. The final message is then:

```
No items found in iteration 'Sprint 7'.
```

That message implies the *iteration* is empty, which it usually isn't — it hides the real situation (the item exists but wasn't where the user thought it was, or the repo/number is misspelled). A typo therefore silently performs no rollover and exits 0. Note that `--issue` accepts both issues and pull requests (`_get_project_items_page` matches `... on Issue` and `... on PullRequest`), so error wording should say "item" rather than "issue".

Suggested fix — when `issue_number is not None`, track whether the `(repo, number)` pair was matched anywhere in `all_items` and elevate "no match" to an error. Three failure modes deserve distinct messages:

- typo / not in project: `Item {repo}#{number} is not in this project.`
- in project, but unassigned: `Item {repo}#{number} is not assigned to any iteration.`
- in project, assigned elsewhere: `Item {repo}#{number} is in iteration '{actual_title}', not '{prev_iter['title']}'. Use --from '{actual_title}' to roll it over.`

A few implementation notes for the suggested-fix author:

- `iterValue` only carries `iterationId`, not the iteration title. Build an `id → title` map at the call site from `iteration_field["configuration"]["iterations"] + completedIterations` (both already fetched in `_get_project_info`), then look up the title before formatting.
- `_migrate_items` does not currently receive the project name (or the iteration ID→title map), so either thread them through or do the per-issue resolution outside the function — easier to scan the items once for the filter case and report the diagnosis before entering the migration loop.
- All three should exit non-zero so scripted callers detect the mistake.

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,541 @@ def get_aggregate_changelog(  # noqa: C901
...
+                  title
+                  repository { name }
+                }
+                ... on PullRequest {
+                  number
+                  title
+                  repository { name }
+                }
+              }
+              statusValue: fieldValueByName(name: "Status") {
```
tiborsimko: **🤖 `_get_project_items_page` hardcodes the Status field name (LOW)**

The per-item GraphQL query selects:

```python
statusValue: fieldValueByName(name: "Status") {
  ... on ProjectV2ItemFieldSingleSelectValue { optionId }
}
iterValue: fieldValueByName(name: $iterFieldName) {
  ... on ProjectV2ItemFieldIterationValue { iterationId }
}
```

The iteration field name is parameterised (good — it's discovered dynamically via `_find_iteration_field`), but `"Status"` is hardcoded. `fieldValueByName` is case-sensitive (confirmed by live testing: `"Status"` returns a value, `"status"` returns `null`). Meanwhile `_get_done_option_ids` does a case-*insensitive* match — `field.get("name", "").lower() == "status"` — so the two paths disagree on casing.

The concrete failure: a project whose status field is named literally `"status"` (or any non-`"Status"` casing) passes Done-detection — `done_option_ids` is populated — but every item's `statusValue` comes back `null` from the per-item query. `if item_status_option_id in done_option_ids` therefore never fires, and Done items get migrated alongside the rest. (A project with a totally different name like `"State"` is *not* affected: `_get_done_option_ids` would fail to find a status field at all and raise `ClickException("Project has no 'Status' field with options...")`, so the command aborts cleanly. The bug is only in the casing-mismatch corner.)

Suggested fix — discover the status field's actual `name` once (in `_get_done_option_ids`, or split out a sibling helper that returns both the option IDs and the field name), then thread that name through `_get_project_items_page` the same way `iter_field_name` is threaded:

```python
def _get_project_items_page(project_id, iter_field_name, status_field_name, after=None):
    query = """
    query($id: ID!, $iterFieldName: String!, $statusFieldName: String!, $after: String) {
      ...
      statusValue: fieldValueByName(name: $statusFieldName) { ... }
      iterValue: fieldValueByName(name: $iterFieldName) { ... }
      ...
    }
    """
```

reana/reana_dev/git.py
```diff
@@ -2423,4 +2423,541 @@ def get_aggregate_changelog(  # noqa: C901
...
+    while True:
+        page = _get_project_items_page(project_id, iter_field_name, cursor)
+        items.extend(page["nodes"])
+        if not page["pageInfo"]["hasNextPage"]:
+            break
+        cursor = page["pageInfo"]["endCursor"]
+    return items
+
+
+def _migrate_items(
```
tiborsimko, line 2917: **🤖 `_migrate_items` takes nine positional parameters (NIT)**

```python
def _migrate_items(
    all_items,
    project_id,
    prev_iter,
    curr_iter,
    iteration_field_id,
    done_option_ids,
    issue_number,
    issue_repo,
    dry_run,
):
```

Called positionally from `git_iteration_rollover`:

```python
moved_count, skipped_count = _migrate_items(
    all_items,
    project_info["id"],
    prev_iter,
    curr_iter,
    iteration_field["id"],
    done_option_ids,
    issue_number,
    issue_repo,
    dry_run,
)
```

Hard to scan, and a swapped `issue_number`/`issue_repo` at the call site would be a silent bug since both come from `_parse_issue_filter` (one `int`, one `str`, but the function does not type-check).

Suggested fix — call by keyword at the only call site:

```python
moved_count, skipped_count = _migrate_items(
    all_items=all_items,
    project_id=project_info["id"],
    prev_iter=prev_iter,
    curr_iter=curr_iter,
    iteration_field_id=iteration_field["id"],
    done_option_ids=done_option_ids,
    issue_number=issue_number,
    issue_repo=issue_repo,
    dry_run=dry_run,
)
```

A genuine reduction in the parameter count would require introducing a small context object (e.g. a `@dataclass` carrying project ID, iteration field ID, source/target iterations, Done option IDs, and the issue filter) and passing it as a single argument. Swapping the extracted scalars for their parent dicts (`project_info` for `project_id`, `iteration_field` for `iteration_field_id`) does not reduce the count — it just renames what each parameter carries.

reana/reana_dev/git.py
tiborsimko: Addressed in commit `a15949d`, kept as a separate fixup on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana/reana_dev/git.py
tiborsimko: Addressed in commit `a15949d`, kept as a separate fixup on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana/reana_dev/git.py
tiborsimko: Addressed in commit `a15949d`, kept as a separate fixup on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana/reana_dev/git.py
tiborsimko: Addressed in commit `a15949d`, kept as a separate fixup on top of the feat commit for review clarity. This fixup commit can be squashed into the feat commit before merge, preserving the first author with `Co-authored-by` for the fixup.

michaelbuchar (approved):
LGTM!

PR #961, opened 2026-06-02
fix(reana-dev): avoid multi-manifest indices in kind load

tiborsimko:
The `kind load docker-image` command fails on daemons that use the containerd image store (default on Docker Desktop, progressively rolling out on Docker CE). It shells out to `docker save` without a platform filter and feeds the result to `ctr import --all-platforms --digests` inside the node; the saved archive references every sibling manifest in the multi-arch index — both other platforms and SLSA/SBOM attestations — but only carries the blobs for the locally fetched platform, so the import aborts on the first missing referenced digest.

For `kind-load-docker-image`, switch to a platform-pinned `docker image save -o <tar>` plus `kind load image-archive <tar>`. The filtered archive contains a single manifest, so the import is clean regardless of which image store the daemon uses. The platform is read from the local image record via `docker image inspect --format '{{.Os}}/{{.Architecture}}'`, so single-platform upstream images pulled under emulation (e.g. amd64-only images on an arm64 daemon) still match what Docker actually stored, rather than tripping on a non-existent daemon-native variant. Always pass an explicit image tag, since on the containerd image store a bare repo name can match multiple tagged images and `docker image save` may then resolve to the wrong one. Expose `-t/--tag` (default `latest`) to make the tag selectable, matching the sibling `docker-pull` and `docker-build` commands.

For `docker-build`, switch to `docker buildx build --provenance=false --sbom=false --load` so locally built images are a single self-contained manifest, and check buildx availability only when at least one dockerised component is queued for build.

### DIFF
--- a/reana/reana_dev/kind.py
+++ b/reana/reana_dev/kind.py
@@ -1,12 +1,15 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
-# Copyright (C) 2020, 2021, 2023, 2025 CERN.
+# Copyright (C) 2020, 2021, 2023, 2025, 2026 CERN.
 #
 # REANA is free software; you can redistribute it and/or modify it
 # under the terms of the MIT License; see LICENSE file for more details.
 
 """`reana-dev`'s kind commands."""
+
+import os
+import tempfile
 
 import click
 
@@ -24,7 +27,56 @@ def kind_commands():
     """Kind commands group."""
 
 
+def _kind_load_image(image, nodes, component):
+    """Load a Docker image into the local kind cluster.
+
+    Goes via a platform-pinned ``docker image save`` plus
+    ``kind load image-archive``, rather than ``kind load docker-image``.
+    The latter shells out to ``docker save`` without a platform filter,
+    which on daemons that use the containerd image store (default on
+    Docker Desktop and progressively rolling out on Docker CE) exports
+    the full multi-arch index even when only one platform was pulled.
+    The subsequent ``ctr import --all-platforms --digests`` inside the
+    kind node then aborts on the first sibling-manifest digest whose
+    content is not in the archive. Saving with ``--platform`` produces
+    a single-manifest tarball that imports cleanly.
+
+    The platform is read from the local image record (not from the
+    daemon platform), because single-platform upstream images may have
+    been pulled under emulation — e.g. an amd64-only image on an arm64
+    daemon. ``docker image save --platform <daemon>`` would then fail
+    with ``no suitable export target found``.
+
+    This is a workaround for a Kind issue tracked upstream at
+    https://github.com/kubernetes-sigs/kind/issues/3795.
+    """
+    platform = run_command(
+        "docker image inspect --format '{{.Os}}/{{.Architecture}}' " + image,
+        display=False,
+        return_output=True,
+    )
+    fd, archive = tempfile.mkstemp(suffix=".tar", prefix="reana-kind-load-")
+    os.close(fd)
+    try:
+        run_command(
+            "docker image save --platform {0} -o {1} {2}".format(
+                platform, archive, image
+            ),
+            component,
+        )
+        cmd = "kind load image-archive {0}".format(archive)
+        if nodes:
+            cmd = "{0} --nodes {1}".format(cmd, ",".join(nodes))
+        run_command(cmd, component)
+    finally:
+        try:
+            os.unlink(archive)
+        except FileNotFoundError:
+            pass
+
+
 @click.option("--user", "-u", default="reanahub", help="DockerHub user name [reanahub]")
+@click.option("--tag", "-t", default="latest", help="Image tag [latest]")
 @click.option(
     "--component",
     "-c",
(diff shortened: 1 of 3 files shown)


Sohn123 (changes requested):
I tested it locally, but it seems like the changes don't fix the problem. In fact they made it worse on my machine. 
I removed all existing images by running: `reana-dev cluster-undeploy && docker system prune -a`
I pulled the docker images, with the correct flags using `reana-dev cluster-create -m /var/reana:/var/reana`:
```
[2026-06-02T16:24:48] reana: docker pull --platform linux/amd64 docker.io/library/postgres:14.10
14.10: Pulling from library/postgres
c57ee5000d61: Pull complete 
b6688a8d81af: Pull complete 
488633c8ea56: Pull complete 
ec1126f7a9f6: Pull complete 
bd4bb07f416b: Pull complete 
bd2705304f5a: Pull complete 
510930354b87: Pull complete 
123a55cee189: Pull complete 
073099a4d6f5: Pull complete 
45211a234e51: Pull complete 
7bcd1ca95189: Pull complete 
2f8c2684013a: Pull complete 
e4fc4fc541b7: Pull complete 
4b61c3e3b567: Pull complete 
70a746511e9d: Download complete 
745bc8085cb9: Download complete 
Digest: sha256:34f4ee87fed96f254c4970e6a348c24b24b4f05ebb0880e29c8d15a5cde4e763
Status: Downloaded newer image for postgres:14.10
docker.io/library/postgres:14.10
```
But when trying to load them it fails:
```
[2026-06-02T16:25:34] reana: kind load docker-image docker.io/library/postgres:14.10
Image: "docker.io/library/postgres:14.10" with ID "sha256:34f4ee87fed96f254c4970e6a348c24b24b4f05ebb0880e29c8d15a5cde4e763" not yet present on node "kind-control-plane", loading...
ERROR: failed to load image: command "docker exec --privileged -i kind-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
Command Output: ctr: content digest sha256:ac6b57b65c04a4076feab30d2f80e33525b978381d5f555463dd5cbc1ee3197d: not found
[2026-06-02T16:25:34] reana: Command 'kind load docker-image docker.io/library/postgres:14.10' returned non-zero exit status 1.
[2026-06-02T16:25:34] reana: Command 'reana-dev kind-load-docker-image -c reana' returned non-zero exit status 1.
```
But local builds + loading the images with kind works fine using `reana-dev cluster-build --exclude-components=r-a-krb5,r-a-vomsproxy,r-a-rucio --parallel 8`. Also deploying to the cluster works fine afterwards. So it seems like loading the remote images onto the cluster isn't even necessary. I would propose we remove the logic to only pull the current platform as it just doesn't help and rather catch the error and issue a warning to the user, that the image will be downloaded when deploying the cluster for the first time.

tiborsimko:
> I tested it locally, but it seems like the changes don't fix the problem.

I see, I could reproduce the problem locally and a new version should fix it fine. Please retest.

reana/reana_dev/kind.py
```diff
@@ -24,7 +27,53 @@ def kind_commands():
...
+    The subsequent ``ctr import --all-platforms --digests`` inside the
+    kind node then aborts on the first sibling-manifest digest whose
+    content is not in the archive. Saving with ``--platform`` produces
+    a single-manifest tarball that imports cleanly.
+
+    The platform is read from the local image record (not from the
+    daemon platform), because single-platform upstream images may have
+    been pulled under emulation — e.g. an amd64-only image on an arm64
+    daemon. ``docker image save --platform <daemon>`` would then fail
+    with ``no suitable export target found``.
```
Sohn123, line 48: I don't know if that fits your documentation style, but I would add a link to the github discussion where this issue is tracked, because this is only a workaround and not the intended behavior. So when people want to understand why we go that way it might provide even more context: https://github.com/kubernetes-sigs/kind/issues/3795

reana/reana_dev/kind.py
```diff
@@ -24,7 +27,53 @@ def kind_commands():
...
+    """
+    platform = run_command(
+        "docker image inspect --format '{{.Os}}/{{.Architecture}}' " + image,
+        display=False,
+        return_output=True,
+    )
+    fd, archive = tempfile.mkstemp(suffix=".tar", prefix="reana-kind-load-")
```
Sohn123, line 58: Also don't know about your coding conventions, but I would propose to use the Tempfile context manager, so we avoid to leak resources in any failure case. 
```python
with tempfile.TemporaryFile((suffix=".tar", prefix="reana-kind-load-") as archive:
    ...
```

Sohn123 (commented):
Thanks for putting in the effort. I tested again and it now works flawlessly. I left some minor nit picks, but those are also a matter of taste, so I'm happy to approve it if we decide to keep it like this. Do we want to keep an issue open to track this workaround, that links to the open kind issue, so in the future we might switch back to the normal `kind load docker-image` instead of our workaround?

reana/reana_dev/kind.py
tiborsimko: Thanks, added.

reana/reana_dev/kind.py
tiborsimko: Hmm, I'd lean towards leaving the `mkstemp` and `try/finally` as-is. The current shape already cleans up on every exit path (success, exception from run_command, interrupt), so there's no actual resource leak to fix? It's just a slightly older idiom than with `NamedTemporaryFile(...)`. So swapping it would be a stylistic change with no behavioural difference... BTW `TemporaryFile` specifically wouldn't work here, because `docker image save -o` needs a named path on disk, so one would need something like:

```python
  with tempfile.TemporaryDirectory(prefix="reana-kind-load-") as tmpdir:
      archive = os.path.join(tmpdir, "image.tar")
      run_command(
          "docker image save --platform {0} -o {1} {2}".format(
              platform, archive, image
          ),
          component,
      )
      cmd = "kind load image-archive {0}".format(archive)
      if nodes:
          cmd = "{0} --nodes {1}".format(cmd, ",".join(nodes))
      run_command(cmd, component)
```

But I'm not sure it's worth it...

Sohn123: Ok then let's stick with the current implementation 👍🏼

Sohn123 (approved):
I tested it and it works on Ubuntu now and we clarified the stylistic nits so for me it is ready to merge :shipit:

PR #965, opened 2026-06-02
fix(reana-dev): activate environment using dot-source command

### LINKED ISSUE #962: Bug: `reana-dev python-unit-tests` not working on ubuntu
Running the unit tests with the `reana-dev` wrapper fails because it tries to use the `source` command in the sh shell. This works on MacOS because there `/bin/sh` is an alias for `/bin/bash`. On ubuntu it is by default `/bin/dash` which does not support the source command. 

```
[2026-06-02T16:57:31] reana: virtualenv ~/.virtualenvs/_reana -p python3.12
created virtual environment CPython3.12.13.final.0-64 in 205ms
  creator CPython3Posix(dest=/home/alexander-sohn/.virtualenvs/_reana, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, via=copy, app_data_dir=/home/alexander-sohn/.local/share/virtualenv)
    added seed packages: pip==25.1.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
[2026-06-02T16:57:32] reana: source  ~/.virtualenvs/_reana/bin/activate && which python
/bin/sh: 1: source: not found
[2026-06-02T16:57:32] reana: Command 'source  ~/.virtualenvs/_reana/bin/activate && which python' returned non-zero exit status 127.
```

Sohn123:
Running the python unit tests with the `reana-dev` wrapper failed on ubuntu, because the default shell does not support the `source` feature. This pr replaces source with the more general `. /path/to/source` syntax, making it compatible with ubuntu. 

Fixes #962

### DIFF
--- a/reana/reana_dev/python.py
+++ b/reana/reana_dev/python.py
@@ -134,7 +134,7 @@ def python_unit_tests(
             )
             display_message(msg, component)
         elif is_component_python_package(component):
-            cmd_activate_venv = f"source  ~/.virtualenvs/_{component}/bin/activate"
+            cmd_activate_venv = f".  ~/.virtualenvs/_{component}/bin/activate"
             if does_component_need_db(component):
                 run_command(
                     f"docker stop postgres__{component}\n"
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -8,6 +8,7 @@ organisation on GitHub, in alphabetical order:
 - [Adelina Lintuluoto](https://orcid.org/0000-0002-0726-1452)
 - [Agisilaos Kounelis](https://orcid.org/0000-0001-9312-3189)
 - [Alastair Lyall](https://orcid.org/0009-0000-4955-8935)
+- [Alexander Sohn](https://orcid.org/0009-0000-3673-4637)
 - [Alizee Pace](https://www.linkedin.com/in/aliz%C3%A9e-pace-516b4314b/)
 - [Alp Tuna](https://orcid.org/0009-0001-1915-3993)
 - [Ana Trisovic](https://orcid.org/0000-0003-1991-0533)


reana/reana_dev/python.py
```diff
@@ -134,7 +134,7 @@ def python_unit_tests(
             )
             display_message(msg, component)
         elif is_component_python_package(component):
-            cmd_activate_venv = f"source  ~/.virtualenvs/_{component}/bin/activate"
+            cmd_activate_venv = f".  ~/.virtualenvs/_{component}/bin/activate"
```
tiborsimko, line 137: Thanks. I think it would be good to mention the "why" in the commit message, not only in the PR description. (So that we record motivation for "offline" work too.) Also, the "Activate" should be lowercase, and we could explicitly mention "dash" shell as the reason, and terminate with "Closes #123" to link commit to the issue.

Here is a suggested commit message headline and body:

```
  fix(reana-dev): activate environment using dot-source command (#965)

  The python-unit-tests helper activated the throw-away virtualenv
  with `source`, which is a bash/zsh extension. On Ubuntu the default
  `/bin/sh` is `dash`, which does not implement `source` and fails
  with `source: not found`, preventing the unit tests from starting.

  Replace `source` with the POSIX dot-source command (`.`), which
  behaves identically and is supported by `dash`, `bash`, and `zsh`.

  Closes #962
```

tiborsimko: Also, please add yourself to the `AUTHORS.md` file.

Sohn123: I updated the commit message with your proposal and added myself to the Author.md file. So should be good to merge :) :shipit:

tiborsimko (approved):

PR #969, opened 2026-06-08
perf(reana-dev): speed up git-status on macOS

tiborsimko:
Each per-component step in `reana-dev git-status` previously shelled out
via `subprocess.check_output(..., shell=True)` with pipelines like `git
branch | grep "^*" | colrm 1 2` and `git branch -a | grep -c
remotes/<branch>`. Each `shell=True` call forks `/bin/sh`, which in turn
forks `git` and any pipeline helpers. On macOS where `fork`/`exec`
carries dyld closure resolution, library validation and AMFI checks,
this overhead dominated the runtime.

Rework the hot path to call `git` directly with list-form arguments and
`cwd=srcdir`:

  - New `remote_ref_exists(srcdir, branch)` helper using `git show-ref
    --verify --quiet refs/remotes/<branch>`, used in both
    `compare_branches` and `git_status` instead of the `git branch -a |
    grep` pipeline and the `"remotes/..." in get_all_branches(...)`
    membership test.
  - `get_current_branch` (in `git.py` and `utils.py`),
    `get_current_commit` and `get_all_branches` rewritten to use list
    arguments, `cwd=srcdir` and no shell pipeline.
  - `compare_branches` now takes `srcdir` explicitly, making the
    no-`chdir` contract obvious and avoiding a hidden
    `get_srcdir(component)` lookup inside.
  - `print_branch_difference_report` takes an optional `srcdir`, accepts
    a pre-computed `current_branch`/`commit` from the caller
    (`git_status` already had them) and tolerates
    `branch_to_compare=None` for detached-HEAD checkouts, in which case
    the ahead/behind comparison is skipped but `--short` still prints
    `git status --short`.
  - `is_component_behind_branch` computes `srcdir` once and threads it
    to `compare_branches`.
  - `git_status` now computes `srcdir` once per component, detects
    detached HEAD (`git rev-parse --abbrev-ref HEAD` returns `HEAD`) and
    prints `(detached HEAD)` instead of erroring on a non-existent
    `origin/HEAD` lookup.

For the frequently used `reana-dev git-status -c CLUSTER -c CLIENT -c
reana` command, which walks 19 REANA source repositories, wall-clock
time on macOS drops from ~2.4s to ~1.3s (~1.9x), with user CPU time
falling from ~1.0s to ~0.4s and system CPU time from ~1.0s to ~0.3s.

Output is byte-identical for branches that exist on
`master`/`origin`/`upstream`. Detached-HEAD output is intentionally not
byte-identical: today's pipeline emits a literal `(HEAD detached at
<sha>)` label and then trips the downstream `origin/...` lookup with an
`ERROR: Branch ... does not exist` line; the rewrite detects the
detached state up front, prints `(detached HEAD)` and skips the
ahead/behind comparison.

### DIFF
--- a/reana/reana_dev/client.py
+++ b/reana/reana_dev/client.py
@@ -89,10 +89,17 @@ def client_setup_environment(
                 env_var_value=server_hostname or "https://localhost:30443",
             )
         )
-        get_access_token_cmd = f"kubectl get secret -n {namespace} -o json {instance_name}-admin-access-token"
-        secret_json = json.loads(
-            subprocess.check_output(get_access_token_cmd, shell=True).decode()
-        )
+        get_access_token_cmd = [
+            "kubectl",
+            "get",
+            "secret",
+            "-n",
+            namespace,
+            "-o",
+            "json",
+            f"{instance_name}-admin-access-token",
+        ]
+        secret_json = json.loads(subprocess.check_output(get_access_token_cmd).decode())
         admin_access_token_b64 = secret_json["data"]["ADMIN_ACCESS_TOKEN"]
         admin_access_token = base64.b64decode(admin_access_token_b64).decode()
         export_lines.append(
--- a/reana/reana_dev/docker.py
+++ b/reana/reana_dev/docker.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of REANA.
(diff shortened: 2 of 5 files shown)


michaelbuchar (approved):
Looks good to me, I think the only gap is missing regression tests for the new paths but we can add those later!

PR #972, opened 2026-06-17
feat(helm): add values to set backend-specific concurrency limits

CameronMcClymont:
Closes reanahub/reana-workflow-controller#663


helm/reana/templates/reana-server.yaml
```diff
@@ -90,6 +90,14 @@ spec:
             value: {{ include "reana.prefix" . }}
           - name: REANA_COMPUTE_BACKENDS
             value: {{ .Values.compute_backends | toJson | quote }}
+          - name: REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS
```
tiborsimko, line 104: **🤖 RWC663-05: Chart always emits the new K8S cap variable, defeating the advertised legacy fallback (MEDIUM)**

reana-commons preserves backward compatibility by reading the legacy variable only when the new one is absent:

```python
REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS = int(
    os.getenv(
        "REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS",
        os.getenv("REANA_MAX_CONCURRENT_BATCH_WORKFLOWS", "30"),
    )
)
```

But the chart now emits `REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS` unconditionally, from a value that defaults to 30 (`helm/reana/templates/reana-server.yaml:93`, and again for the scheduler container at line 345). Under Helm the fallback can therefore never fire. An operator upgrading with an existing values file:

```yaml
components:
  reana_server:
    environment:
      REANA_MAX_CONCURRENT_BATCH_WORKFLOWS: 10
```

renders *both* variables, the new one wins, and the cluster silently starts admitting 30 concurrent Kubernetes workflows instead of the operator's intended 10 — three times the configured load, with no warning anywhere. The compatibility fallback ends up helping only non-Helm deployments, which is the opposite of where legacy values files actually live.

Suggested fix — resolve the deprecated value when computing the new authoritative variable, with the precedence: explicit `concurrency_limits.kubernetes` → legacy `environment.REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` → built-in default 30. For example, make `concurrency_limits.kubernetes` nullable in `values.yaml` (`kubernetes: ~`) and resolve in the template (in both containers):

```yaml
{{- $limits := .Values.components.reana_server.concurrency_limits }}
{{- $env := .Values.components.reana_server.environment }}
{{- $k8sCap := 30 }}
{{- if hasKey $env "REANA_MAX_CONCURRENT_BATCH_WORKFLOWS" }}
{{- $k8sCap = index $env "REANA_MAX_CONCURRENT_BATCH_WORKFLOWS" }}
{{- end }}
{{- if not (kindIs "invalid" $limits.kubernetes) }}
{{- $k8sCap = $limits.kubernetes }}
{{- end }}
- name: REANA_MAX_CONCURRENT_K8S_BATCH_WORKFLOWS
  value: {{ $k8sCap | quote }}
```

Note that sprig's `default` function cannot express this precedence: it treats a numeric `0` as empty, so a `default`-chain would silently replace an intentional zero cap ("backend closed") with the legacy value or 30 — explicit `hasKey`/nil checks are needed. The generic `environment` loop will still emit the legacy variable itself; that is harmless once the new variable always carries the resolved value, but for tidiness the loop could skip the `REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` key. Alternatively, `fail` template rendering when the legacy key is present, with a message pointing at `concurrency_limits.kubernetes`. At the very minimum, the Helm README needs an upgrade note stating that a legacy `environment.REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` entry is ignored and must be moved to `concurrency_limits.kubernetes`.

helm/reana/values.yaml
```diff
@@ -139,6 +138,17 @@ components:
...
       REANA_ACCESS_TOKEN_ISSUANCE_POLICY: "manual"
       REANA_TOKEN_MANAGEMENT_SECRET: ""
+    # Upper limits on concurrently pending/running batch workflows.
+    # A running workflow consumes one slot in every compute backend
+    # its steps use, plus a Dask slot if it requests a Dask cluster.
+    # It is admitted only while below every applicable limit.
+    concurrency_limits:
+      kubernetes: 30
+      external: 200 # default for any external backend (e.g. htcondorcern/slurmcern)
+      dask: 10
```
tiborsimko: **🤖 RWC663-09: Dask cap default is 10 in Helm values but 5 everywhere else (LOW)**

Three sources of truth for the default Dask concurrency cap disagree:

- `helm/reana/values.yaml:148` — `concurrency_limits.dask: 10`
- `helm/reana/README.md` — documents the default as `5`
- `reana-commons/reana_commons/config.py` — `REANA_MAX_CONCURRENT_DASK_WORKFLOWS` defaults to `5`

So a Helm deployment gets 10 while its own README promises 5, and a bare-metal or development deployment gets 5. Whichever number is intended, the three should agree — otherwise the first operator who reads the README and then checks the dashboard files a bug report.

Suggested fix — pick one default and align all three places. `5` matches both the code default and the documented value, so setting `values.yaml` to 5 is the smallest change; if `10` is the deliberate choice, update the README table and the reana-commons default in the same series instead.

tiborsimko:
Note: this branch currently conflicts with `master` in `helm/reana/values.yaml`, due to #889, #984 and #982 having landed there since the branch point. The conflict looks purely mechanical — I found no semantic interaction between those changes and the concurrency limits — so a straightforward rebase should do.

helm/reana/templates/reana-server.yaml
CameronMcClymont: Fixed, using the precedence you suggested: explicit `concurrency_limits.kubernetes` → deprecated `environment.REANA_MAX_CONCURRENT_BATCH_WORKFLOWS` → built-in default of 30.

The resolution lives in a `reana.k8sConcurrencyLimit` helper in `_helpers.tpl` so both the REST API and the scheduler container get the same value, and `concurrency_limits.kubernetes` now ships as `~`.

One correction to the sketch: `kindIs "invalid"` does not work here, because Helm drops nil-valued keys from `.Values` altogether — with `kubernetes: ~` the key is simply absent, so `get` returns `""` (a string) and the check never fires. `hasKey` is exact for this, and still distinguishes an intentional `0` from "not configured", which was the reason not to use `default`.

Rather than an upgrade note saying the deprecated value is ignored, the chart now honours it, so an existing values file keeps its intended cap on upgrade. The README gained a "Workflow concurrency limits" section documenting the per-backend/Dask model, the meaning of a `0` limit, and how to migrate off the deprecated variable.

I left the generic `environment` loop emitting the deprecated key as-is — harmless now that the new variable always carries the resolved value, and it keeps the operator's setting visible in the pod env.

Covered by `test_server_kubernetes_concurrency_limit`, which renders all four cases (default / deprecated-only / explicit-wins / explicit-zero) and asserts both containers agree.

helm/reana/values.yaml
CameronMcClymont: Fixed so `values.yaml` sets `concurrency_limits.dask: 5` (matching the README).

Added `test_server_concurrency_limit_defaults_match_reana_commons` so they cannot drift again.

helm/reana/templates/reana-server.yaml
CameronMcClymont: Correction to my note above on `kindIs "invalid"` — it was right for Helm 4 but wrong in general, and CI caught it.

The two Helm majors represent `kubernetes: ~` differently: **Helm 4 drops** the nil-valued key from `.Values` entirely, while **Helm 3 keeps** the key with a nil value. So neither check alone is enough — `hasKey` misfires on Helm 3 (key present, value nil, cap rendered empty) and `kindIs "invalid"` misfires on Helm 4 (key absent, `get` returns `""`, which is a string and not "invalid").

The helper now treats a value as configured only when the key is present **and** non-nil, which is correct on both:

```
{{- if hasKey $limits "kubernetes" -}}
{{- $kubernetes := index $limits "kubernetes" -}}
{{- if not (kindIs "invalid" $kubernetes) -}}
{{- $cap = $kubernetes -}}
{{- end -}}
{{- end -}}
```

An explicit `0` is present and non-nil under both majors, so it still reads as "configured to zero" rather than falling through. I verified all four precedence cases render identically under Helm 3.16.4 and Helm 4.2.0; `tests/test_helm.py` passes against both.

CameronMcClymont:
Rebased — the conflict is gone and the PR shows as mergeable again.

The branch now sits directly on top of current `master` (0 commits behind), so #889, #984 and #982 are all in. Your read was right that it was purely mechanical: `values.yaml` picked up the infra resource requests/limits, the `fsGroup` shared-storage block and the quota-management secret default, none of which touch `concurrency_limits`, and the only fixup needed was context drift around the neighbouring keys.

The other PRs in the set (reanahub/reana-server#780, reanahub/reana-commons#541, reanahub/reana-db#267) were rebased onto their current masters at the same time and are all mergeable too.

PR #979, opened 2026-06-26
feat(reana-dev): register reana-datastore-s3fs component

tiborsimko:
Add `reana-datastore-s3fs` component to the ALL and CLUSTER component selectors, and list it in the generated build status wiki page under Datastores.


michaelbuchar (approved):
LGTM!

PR #980, opened 2026-06-29
feat(helm): support sandboxed workflow specification validation

### LINKED ISSUE #8: Move workflow specification building and validation to the server
## Background

Today, validating a workflow specification requires the client to call the corresponding workflow engine (CWL, Serial, Snakemake, Yadage) to build the specification, then validate it locally. The server receives this specification and executes it as-is.

## Problem

This client-side approach has two consequences:

- Client language lock-in. Supporting a new client requires reimplementing every supported workflow engine in that language. This is why reana-client-go currently cannot create or submit workflows — porting all engines to Go is impractical.
- No security boundary. The server trusts a specification generated on the client without re-checking it. Client-side validation is a quality-of-life feature that catches mistakes early, not a guarantee the server can rely on.

## Proposal

Move specification building and validation to the server, so any client only needs to send a workflow snapshot (source files) and let the server do the rest.

Because building a specification means evaluating the workflow file, which can execute arbitrary code (e.g. Snakemake run: blocks, Jinja templating), this must happen inside a sandbox with least-privilege access — otherwise the server itself becomes remotely exploitable by a malicious specification.

## Acceptance Criteria

- [ ] Specification building and validation happen server-side, so a client only needs to submit workflow files, not run the corresponding engine
- [ ] Specification building and validation run in a sandboxed, least-privilege environment isolated from the rest of the server
- [ ] The sandbox has network access to fetch external dependencies referenced by the workflow (e.g. git+https CWL tools)

Sohn123:
Configure the Helm chart and local development tooling to run workflow
specification validation in short-lived sandboxed Jobs. Pass the
validator image and non-secret environment to the workflow controller,
reject sandbox-reserved environment names, grant the controller
permission to manage validation NetworkPolicies, and prefetch the
validator image.

Expose aligned limits for validation-snapshot size and file count,
remote-source fetching, and validator ephemeral storage. Configure
whether validation Jobs may use public egress.

Document that NetworkPolicy enforcement requires a compatible CNI,
combines additively with other policies, and cannot isolate a pod from
its resident node. Provide a live probe for verifying the deployed
cluster's isolation behaviour.

Add Helm and live-cluster coverage for validator configuration, RBAC,
bounded snapshots, workspace seeding, unsafe ZIP rejection, and
byte-exact workspace uploads.

Closes reanahub/reana-workflow-validator#8


helm/reana/templates/roles.yaml
```diff
@@ -18,6 +18,10 @@ rules:
 - apiGroups: ["", "apps", "networking.k8s.io"]
   resources: ["deployments", "services", "ingresses"]
   verbs: ["get", "create", "delete"]
+# NetworkPolicy for the sandboxed workflow specification validator
+- apiGroups: ["networking.k8s.io"]
+  resources: ["networkpolicies"]
+  verbs: ["get", "list", "create", "update", "patch", "delete"]
```
tiborsimko: **🤖 SNDBX-17: RBAC grants `networkpolicies` verbs the controller never uses (NIT)**

*Component:* reana (umbrella) — `helm/reana/templates/roles.yaml`;
reana-workflow-controller — `reana_workflow_controller/spec_validation.py:392-441`

The new role rule grants the workflow-controller broad NetworkPolicy access:

```yaml
- apiGroups: ["networking.k8s.io"]
  resources: ["networkpolicies"]
  verbs: ["get", "list", "create", "update", "patch", "delete"]
```

But `_ensure_egress_network_policy` only ever calls
`create_namespaced_network_policy` and, on 409,
`replace_namespaced_network_policy` (`spec_validation.py:421,427`) — i.e. `create`
and `update`. `list`, `patch`, and `delete` are not exercised by the code. Granting
unused verbs is harmless but widens the controller's privileges beyond least-
privilege for no functional reason.

This is optional least-privilege cleanup, not a feature risk.

Suggested fix — trim the verb list to `["create", "update"]` (add `get` only if a
read is introduced). If the broader set is intentional future-proofing, a one-line
comment saying so would prevent a future reviewer flagging it again.

michaelbuchar:
Small local-dev follow-up: this PR adds reana-workflow-validator as a runtime dependency of the new validation path, but the usual local image prefetch/load flow still does not include that image. In a fresh local/kind setup that means the feature can appear broken until the validator image is loaded manually. I think the deployment changes here are fine, but should we also update scripts/prefetch-images.sh in a follow-up so local development stays aligned with the new runtime dependency?

michaelbuchar:
I pulled the full 7-PR stack locally and validated the end-to-end server-side sandboxed validation flow in a kind cluster, including the server/controller/validator wiring, the RBAC/network-policy requirement, and the non-serial validate/create/start paths from both the Python and Go clients. The core design looks good and the main happy path worked in practice, so overall this is solid work and a good step forward on the issue! I left a couple of review comments for the remaining follow-ups before merge.

helm/reana/templates/roles.yaml
Sohn123: I trimmed the list as proposed.

helm/reana/values.yaml
```diff
@@ -214,6 +214,11 @@ components:
       REANA_OPENSEARCH_ENABLED: false
       REANA_OPENSEARCH_USE_SSL: true
       REANA_OPENSEARCH_CA_CERTS: "/code/certs/ca.crt"
+      # Whether the sandboxed workflow specification validator may reach the
+      # public network (e.g. to resolve a remote yadage `toplevel` or CWL
+      # `$import`). Cluster-internal traffic (to other pods/services) is always
+      # blocked regardless of this setting. Set to "false" for maximum lockdown.
+      REANA_SPEC_VALIDATION_ALLOW_EGRESS: true
```
tiborsimko, line 228: **🤖 PR980-01: Qualify and document the sandbox network-isolation assumptions (LOW)**

*Component:* reana (umbrella) — `helm/reana/values.yaml`
(`REANA_SPEC_VALIDATION_ALLOW_EGRESS`), `helm/reana/README.md`

The new value is accompanied by an absolute security claim:

```yaml
# `$import`). Cluster-internal traffic (to other pods/services) is always
# blocked regardless of this setting. Set to "false" for maximum lockdown.
REANA_SPEC_VALIDATION_ALLOW_EGRESS: true
```

That is stronger than Kubernetes `NetworkPolicy` can guarantee. A policy
resource has no effect unless the cluster's network plugin implements
enforcement; traffic to and from the pod's resident node is always allowed;
and applicable policies combine additively, so another policy selecting the
validator pods can widen their egress. Creating the policy object therefore
does not prove that all cluster-internal destinations are blocked.

The operator-facing documentation also does not say what the isolation
requires or provides:

- `REANA_SPEC_VALIDATION_ALLOW_EGRESS` is described in `values.yaml`, but has
  no row in `helm/reana/README.md`, which documents 16
  other `components.reana_workflow_controller.environment.REANA_*` values.
- The NetworkPolicy-capable-CNI prerequisite is stated nowhere.
- The resident-node limitation is not documented. Operators hardening
  multi-tenant clusters should know that same-node host listeners may remain
  reachable. The controller's corresponding code-level claims are covered by
  [PR693-01 on reana-workflow-controller#693](https://github.com/reanahub/reana-workflow-controller/pull/693).

Since the sandbox deliberately runs user-controlled workflow definitions,
the chart docs are where an operator decides whether their cluster meets
the isolation assumptions.

Suggested fix — qualify the `values.yaml` comment so it describes conditional
NetworkPolicy enforcement rather than saying internal traffic is always
blocked; add a README row for `REANA_SPEC_VALIDATION_ALLOW_EGRESS` covering its
default, effect, and security implication; and document the NetworkPolicy-capable
CNI prerequisite and resident-node limitation in a short "Sandbox isolation
prerequisites" note. Add an integration test on the supported cluster
configuration that attempts forbidden egress from a validator pod and verifies
that it is actually denied, rather than checking only that the policy object
exists.

Sohn123: I added the documentation to clearly highlight limitations/requirements of what we need so this is properly enforced.

Sohn123:
> Small local-dev follow-up: this PR adds reana-workflow-validator as a runtime dependency of the new validation path, but the usual local image prefetch/load flow still does not include that image. In a fresh local/kind setup that means the feature can appear broken until the validator image is loaded manually. I think the deployment changes here are fine, but should we also update scripts/prefetch-images.sh in a follow-up so local development stays aligned with the new runtime dependency?

I added this to the script now :)

scripts/test-spec-validator-network-policy.sh
```diff
@@ -0,0 +1,307 @@
...
+    --ignore-not-found=true --wait=true >/dev/null
+
+cat <<EOF | kubectl apply -f - >/dev/null
+apiVersion: v1
+kind: Pod
+metadata:
+  name: ${probe_pod_name}
+  namespace: ${kubernetes_namespace}
+  labels:
+    app: reana-spec-validator
```
tiborsimko: **🤖 PR980-02: Live NetworkPolicy test probe no longer matches the release-scoped selector (LOW)**

*Component:* reana (umbrella) —
`scripts/test-spec-validator-network-policy.sh` (probe pod at ~lines 178-179,
selector assertion at ~lines 149-150)

The PR693-05 fix correctly made the runtime selector match three labels
(`reana_workflow_mode`, `app`, `app.kubernetes.io/instance`). The live
enforcement probe pod, however, still carries only one:

```yaml
metadata:
  name: ${probe_pod_name}
  namespace: ${kubernetes_namespace}
  labels:
    app: reana-spec-validator
```

Because `matchLabels` is a conjunction, the release-scoped policy no longer
selects this probe, so the probe runs with no egress restriction. On an
enforcing CNI the "cluster-internal destination must be blocked" checks then
fail — not because enforcement is broken, but because the probe is not covered
by the policy — which makes the single live proof of egress enforcement
misleading. The selector assertion only reads `.spec.podSelector.matchLabels.app`
and so would not catch the mismatch either:

```sh
policy_selector=$(kubectl -n "${kubernetes_namespace}" get networkpolicy \
    "${policy_name}" -o jsonpath='{.spec.podSelector.matchLabels.app}')
if [ "${policy_selector}" != "reana-spec-validator" ]; then
```

Production label selection is correct and the script fails visibly rather than
passing silently, so the impact is confined to this manual QA script; it is
nonetheless high priority because it is the only live test of the egress
security boundary. Note the mismatched probe fails on *both* an enforcing CNI
(the policy does not select it) and a non-enforcing CNI (policies are ignored
altogether), so until the labels are fixed the script cannot distinguish those
two cases — which defeats its entire purpose.

Suggested fix — give the probe pod all three validator labels
(`reana_workflow_mode: spec-validation`, `app: reana-spec-validator`,
`app.kubernetes.io/instance: ${instance_name}`) so the policy selects it, and
strengthen the assertion to compare the full `matchLabels` map (all three keys)
rather than only `app`.

Sohn123: Fixed. The live probe now carries all three release-scoped labels, the script asserts the complete selector, and it verifies that the policy selects the probe before testing egress.

tiborsimko (approved):

PR #983, opened 2026-06-30
feat(helm): enrich disabled local development configuration

tiborsimko:
Pre-stage commonly toggled features in the dev values file as flip switches and inert defaults, so they can be easily enabled, disabled, or adjusted for local testing without copying configuration snippets from elsewhere.

Add local-development switches and hints for (i) compute backends (commented CERN HTCondor/Slurm/Compute4PUNCH); (ii) Dask (full sizing subsection) and the Dask Kubernetes operator; (iii) Kueue scheduler; (iv) mail notifications; (v) vetted container images; (vi) custom interactive session images and cleanup periods; (vii) workspace retention rules; (viii) quota enforcement limits and CPU quota replenishment; (ix) maintenance mode; (x) SSO-only login testing; and (xi) access-token issuance and token/quota management API testing.

Keep defaults inactive or equivalent to chart defaults where possible, so the dev values file remains suitable for ordinary local deployments.


helm/configurations/values-dev.yaml
```diff
@@ -113,3 +118,97 @@ secrets:
...
+
+# Quota (enabled by default; change zero values to test enforcement).
+quota:
+  enabled: true
+  default_disk_limit: 0 # bytes (0 = unlimited); e.g. 10737418240 = 10 GiB
+  default_cpu_limit: 0 # milliseconds (0 = unlimited); e.g. 3600000 = 1 hour
+  cpu_period_reset_months: 0 # 0 = no default reset; e.g. 1 = monthly for new users
+
+# Maintenance: set `enabled: true` to scale all REANA workloads down to zero
+# and suspend the periodic cronjobs.
```
michaelbuchar: Should this wording be narrowed a bit? When I render with dask.enabled=true and maintenance.enabled=true, the core REANA deployments scale down, but the dask-kubernetes-operator still renders running because it is gated by dask.enabled, not maintenance.enabled. Do we want to say something like ‘core REANA workloads’ here, or should maintenance also cover the Dask operator?

helm/configurations/values-dev.yaml
```diff
@@ -113,3 +118,97 @@ secrets:
   opensearch:
     user: reana-opensearch-user-for-development
     password: reana-opensearch-password-for-development
+
+# Compute backends; only Kubernetes is active by default.
+compute_backends:
+  - kubernetes
+  # - htcondorcern
+  # - slurmcern
+  # - compute4punch
```
michaelbuchar, line 127: Should the Helm README be updated as part of this too? This change now advertises compute4punch in values-dev.yaml, but the documented compute_backends list in helm/reana/README.md still mentions only kubernetes, htcondorcern, and slurmcern.

helm/configurations/values-dev.yaml
tiborsimko: Good catch! I have reworded the sentence. Covering Dask for the maintenance mode would be quite some work... An issue for later?

helm/configurations/values-dev.yaml
tiborsimko: Added to `helm/reana/README.md`.

michaelbuchar:
Looks good to me, only the readthedocs might need a re-run, approving!

michaelbuchar (approved):

PR #984, opened 2026-07-02
feat(helm): support fsGroup-based shared storage access

michaelbuchar:
Add an optional shared_storage.fs_group chart value and apply it to infrastructure pods that mount the shared workspace volume.

This supports storage backends that rely on group-based write access instead of root-owned volume initialisation, especially for local CephFS-style development setups.

The runtime-side non-root rollout expects shared storage to stay writable through a common group contract, so this change wires the Helm layer needed for that model.


tiborsimko:
Thanks, I have validated the behaviour on the default hostPath deployment: run-batch and user-job security contexts, workflow-engine/job-controller UID/GID, workspace writes, regular workflow execution, and Jupyter sessions. 

I think that CephFS/CSI/NFS fsGroup behaviour can be tested post-merge.

tiborsimko (approved):

PR #988, opened 2026-07-14
fix(reana-dev): use current PR references in git-checkout-pr

### LINKED ISSUE #987: reana-dev: fix stale and nondeterministic git-checkout-pr conflicts
### Current behaviour

Running the same `git-checkout-pr` command twice:

```console
$  reana-dev git-checkout-pr -i reana-workflow-controller 663 --pull --reset
```

produces two different failures:

```
 Conflict for reana-commons: asked to check out both pr-541 and pr-535.
```
and:

```
 Conflict for reana-db: asked to check out both pr-265 and pr-267.
```

There seem to be two problems here:

1. Historical cross-references include superseded closed PRs such as reana-commons#535 and reana-db#265.
2. Set iteration order makes the reported conflict nondeterministic between runs.

### Expected behaviour

The command should use currently linked open PRs and process them deterministically.

tiborsimko:
Query GitHub's current open pull-request references instead of historical issue timeline events, and sort results before conflict detection so repeated invocations behave consistently.

Add regression coverage for issue-derived checkout selection and deterministic conflicts.

Closes #987

### DIFF
--- a/reana/reana_dev/git.py
+++ b/reana/reana_dev/git.py
@@ -854,41 +854,41 @@ def git_checkout(branch, component, exclude_components, fetch):  # noqa: D301
 
 
 def _get_prs_from_issue(repo, issue_number):
-    """Return (component, pr_number) pairs for all PRs linked to a GitHub issue."""
-    # Use the issue timeline API to get all events for this issue
+    """Return sorted (component, pr_number) pairs for open PRs linked to an issue."""
     output = run_command(
-        f"gh api repos/reanahub/{repo}/issues/{issue_number}/timeline --paginate",
+        [
+            "gh",
+            "issue",
+            "view",
+            str(issue_number),
+            "--repo",
+            f"reanahub/{repo}",
+            "--json",
+            "closedByPullRequestsReferences",
+        ],
         display=False,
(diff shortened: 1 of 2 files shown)


CameronMcClymont (dismissed):
Nice fix for this edge case. Just one small thing, feel free to merge if you think it's not worth addressing:

- The help text, docstrings, yellow warning, etc all refer to "open PRs" which technically isn't true because the `gh` query also returns merged PRs I think. This shouldn't really be a problem for us in practice because we always merge all PRs associated with an issue at the same time, it's just a wording accuracy thing.

tiborsimko:
> * The help text, docstrings, yellow warning, etc all refer to "open PRs" which technically isn't true because the `gh` query also returns merged PRs I think. This shouldn't really be a problem for us in practice because we always merge all PRs associated with an issue at the same time, it's just a wording accuracy thing.

Good catch! `closedByPullRequestsReferences` can indeed also include merged PRs. I’ve corrected the wording to say “currently linked PRs.” In our intended workflow we use `git-checkout-pr -i` only for open feature PRs, which are merged together, so no functional change should be necessary.

PR #989, opened 2026-07-30
feat(scripts): add local CERN CoreDNS helper

tiborsimko:
Add a helper for installing, inspecting, and removing a dedicated
CoreDNS fragment that forwards CERN reverse-DNS zones to the name
servers available through the CERN VPN. This lets HTCondor resolve PTR
records needed to construct Kerberos service principals.

Limit the helper to compatible k3s configurations, preserve unrelated
entries in the shared ConfigMap, and restart CoreDNS after changes.
Reference the helper from the local-development Helm values.

### DIFF
--- a/etc/cern-coredns-local.yaml
+++ b/etc/cern-coredns-local.yaml
@@ -1,0 +1,33 @@
+# This file is part of REANA.
+# Copyright (C) 2026 CERN.
+#
+# REANA is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+# Local development only. Manage this configuration with
+# scripts/cern-coredns-local.sh so that other entries in the shared
+# coredns-custom ConfigMap are preserved.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: coredns-custom
+  namespace: kube-system
+data:
+  cern-reverse.server: |
+    # Forward CERN reverse-DNS zones to the DNS servers provided by the VPN.
+    # HTCondor needs these PTR records to construct Kerberos service principals.
+    138.137.in-addr.arpa:53 {
+        errors
+        cache 30
+        forward . 137.138.16.5 137.138.17.5
+    }
+    184.188.in-addr.arpa:53 {
+        errors
+        cache 30
+        forward . 137.138.16.5 137.138.17.5
+    }
+    185.188.in-addr.arpa:53 {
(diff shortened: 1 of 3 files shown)


CameronMcClymont (commented):
I tested the script with Colima and seems to work well. I don't have CERN VPN set up so couldn't test if the records resolve correctly, but I assume it worked for you on CERN's network?

I'm not an expert in this area, but one thing flagged by AI:

kubectl apply --server-side silently deletes sibling keys in coredns-custom when the ConfigMap carries a last-applied-configuration annotation — i.e. when another rule was added with plain kubectl apply -f, which is how the k3s docs present custom *.server entries. Repro on a cluster with no existing coredns-custom:

```
kubectl apply -f other-rule.yaml          # creates it — no warning
./scripts/cern-coredns-local.sh install   # reports success
kubectl get cm coredns-custom -n kube-system -o jsonpath='{.data}'
# other-rule.server is gone
```

This is the documented CSA→SSA upgrade behaviour. Keys added via kubectl create/edit or by another server-side manager are unaffected — only the client-side-apply case.

kubectl patch --type merge avoids it and is symmetric with how remove already works (verified: both keys survive). Since preserving other entries is the stated reason for the helper, this seemed worth raising rather than assuming it's fine.

tiborsimko:
Thanks, good catch — confirmed. I reproduced the silent sibling-key deletion when `coredns-custom` had originally been created using client-side `kubectl apply`.

The helper now creates the ConfigMap when absent and merge-patches it when present. This preserves unrelated entries while still supporting a fresh k3s cluster.

Testing also exposed a related missing-key issue: when the ConfigMap contained another rule but not ours, `status` reported the configuration as installed and `remove` failed because the Go template emitted `<no value>`. This is fixed as well.

I verified the absent, foreign-key-only, client-side-applied, reinstall, and removal cases. I also verified the full DNS path from inside the Colima cluster on the CERN network: `188.184.103.237` resolves to `batchkeeper1.cern.ch`.

Addressed in commit [`eecc100`](https://github.com/reanahub/reana/pull/989/commits/eecc100a85ed2328bb2181343f9eeea0546e920e), kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

CameronMcClymont (approved):
Nice fix, looks good now!

PR #993, opened 2026-08-18
feat(reana-dev): support concurrent work in independent srcdirs

tiborsimko:
Enable developers to work on several multi-repository features and
review several multi-repository pull-request sets concurrently, without
repeatedly switching branches across the canonical REANA source
collection.

Add `reana-dev srcdir-create`, `srcdir-workon`, `srcdir-rename`,
`srcdir-list`, and `srcdir-delete` to manage complete, independent
source copies. Compose each srcdir with the existing `git-checkout` and
`git-checkout-pr` commands. Discover the canonical source and srcdir
roots through marker files, and start every repository from `master`.
Remove copied local branches while preserving canonical remote-tracking
refs so explicitly requested branches can be restored.

Prefer APFS clones on macOS and reflinks on Linux for fast,
space-efficient creation. Fall back to a full archival copy when
copy-on-write is unavailable. Probe Linux reflink support between the
actual source and destination paths, and report the selected strategy
before copying. Include the approximate disk use for full copies.

Optionally provision a per-srcdir mise virtual environment named after
the srcdir. Install the editable source checkout with release tooling,
including build and Twine, by default. Use a dedicated tmux session when
invoked from tmux, and otherwise open a shell in the current terminal.
Keep `-t`/`--tmux` and `--no-tmux` as explicit overrides. Tag managed
sessions with their owning srcdir so the tool never touches unrelated
sessions.

Summarise changed, dirty, and unretained work. Audit commits and stashes
before moving a srcdir to Trash.

Srcdirs isolate source code, branch state, and optional Python
environments, not shared runtime resources such as Docker images,
databases, ports, or Kubernetes deployments.


reana/reana_dev/srcdir.py
```diff
@@ -0,0 +1,2304 @@
...
+        (
+            candidate
+            for candidate in references
+            if _git_ref_exists(repository, candidate)
+        ),
+        None,
+    )
+    if reference is None:
+        raise SrcdirError(
+            f"Cannot initialise {repository_name}: no {GIT_DEFAULT_BASE_BRANCH} "
```
CameronMcClymont, line 593: Any repository without a master ref aborts the whole creation. Could instead skip with a warning instead of aborting. For example, I get `Error: Cannot initialise hsf-training-reana-webpage: no master reference exists.` because I don't have every single REANA repo cloned.

reana/reana_dev/srcdir.py
```diff
@@ -0,0 +1,2304 @@
...
+        git_path = candidate / ".git"
+        if git_path.is_file():
+            raise SrcdirError(
+                f"Cannot copy {source_root}: {candidate.name}/.git is a file, "
+                "which indicates a linked Git worktree."
+            )
+        registered_worktrees = git_path / "worktrees"
+        if registered_worktrees.is_dir() and any(registered_worktrees.iterdir()):
+            raise SrcdirError(
+                f"Cannot copy {source_root}: {candidate.name} has registered "
```
CameronMcClymont, line 397: Registered git worktrees abort creation, one repo at a time. On the first run for me, I get `Error: Cannot copy .../reana-hub: reana-client has registered linked Git worktrees.` Three repos are affected for me (two prunable, one live), but the error only names the first. Could collect all offending repos into one message and treat prunable registrations as prunable.

reana/reana_dev/srcdir.py
```diff
@@ -0,0 +1,2304 @@
...
+    if kitty_sessions is None:
+        _echo_warning(
+            "Kitty state unavailable; run srcdir-list from a Kitty window with "
+            "remote control enabled."
+        )
+    for name, error in errors:
+        _echo_warning(f"cannot inspect {name}: {error}")
+
+
+def _refresh_audit_remotes(repository: Path, source_repository: Path) -> List[str]:
```
CameronMcClymont, line 2044: `srcdir-delete`'s audit took 2m29s at 8% CPU on 46 repos. `srcdir-list` already parallelises (1.2s), but `_refresh_audit_remotes` doesn't, so this could be optimised.

reana/reana_dev/srcdir.py
```diff
@@ -0,0 +1,2304 @@
...
+    Copied local branches are removed; explicitly requested branches can still
+    be restored using reana-dev git-checkout or git-checkout-pr.
+
+    Tmux resolves session targets by prefix, so srcdir names that differ early
+    are quicker to switch between.
+    """
+    _validate_name(name)
+    canonical_source, managed_root = _resolve_context(source_root, srcdir_root)
+    repositories = _discover_repositories(canonical_source)
+    _initialise_registry(canonical_source, managed_root)
```
CameronMcClymont, line 869: A failed create leaves `.reana-srcdirs` and `.reana-srcdirs.json` behind. `_initialise_registry` runs before the copy and isn't rolled back.

CameronMcClymont (changes requested):
Nice PR, just a few comments worth addressing I think.

PR #996, opened 2026-08-19
feat(helm): schedule workflow log pruning

### LINKED ISSUE #964: introduce system-wide retention policy for workflow engine, service and job logs
### Current behaviour

REANA preserves user runtime workflow logs forever. Every workflow run
accumulates:

- **Engine logs** — the workflow engine pod that orchestrates the run (Serial /
  CWL / Yadage / Snakemake).
- **Service logs** — supporting service pods (e.g. Dask) that run alongside the
  workflow.
- **Job logs** — per-step user job stdout/stderr.

Because there is no automated eviction, the total log volume may grow
unboundedly with the number of workflow runs. On busy instances this is
undesirable: logs are usually only consulted in the days immediately following a
run. They do not need to occupy the database space for the lifetime of the
deployment.

### Expected behaviour

Introduce a system-wide log retention policy. The operator-facing surface
mirrors the existing **workspace retention** feature
(`workspaces.retention_rules.*`) so that administrators encounter a familiar
shape, but the underlying mechanism is deliberately simpler: there is no
per-workflow log retention rule, only a single cluster-wide period.

```yaml
logs:
  retention_rules:
    maximum_period: forever # e.g. "60" days; "forever" preserves current behaviour
    cronjob_schedule: "0 3 * * *" # everyday at 3am (staggered from workspace cleanup)
```

After the configured retention period (e.g. 60 days), a periodic cronjob would
delete:

- engine logs
- service logs
- job logs

for all workflow runs older than the threshold, all the while preserving the
workflow record itself, its UUID, parameters, status (`finished`, `failed`,
`stopped`), start/end timestamps, etc.

GitHub Actions offers a good UX precedent here. Past runs remain listed forever
in the Actions tab with their full status, but opening an old run's logs
displays:

> The logs for this run have expired and are no longer available.

We should do the same in REANA. When a user requests logs for an expired run via
`reana-client logs`, the REANA UI, or the REST API, they should see a clear,
friendly message explaining that the logs have been pruned by the cluster's
retention policy.

#### Suggested scope

- **helm**: add `logs.retention_rules.maximum_period` (default `forever`) and
  `logs.retention_rules.cronjob_schedule`, mirroring the existing
  `workspaces.retention_rules.*` values. Render the period into the reana-server
  pod environment via
  `{{ tpl .Values.logs.retention_rules.maximum_period . | quote }}`, exactly as
  is already done for workspaces.
- **reana-db**: add a single `logs_pruned_at` timestamp column on the workflow
  model, set by the cronjob when it actually deletes a run's logs. No separate
  rule table and no `created → active → pending → applied` state machine: a
  cluster-wide policy has no per-workflow lifecycle to track. Storing the
  exact timestamp gives a clear audit trail and lets the UI tell users
  precisely when their logs were pruned.
- **reana-server**: add a cronjob that walks workflows older than the
  retention threshold whose `logs_pruned_at` is `NULL`, deletes their
  associated engine/service/job logs, and stamps `logs_pruned_at = now()`.
- **reana-ui** and **reana-client**: when fetching logs of a workflow whose
  `logs_pruned_at` is set, surface a clear "the logs for this run were pruned
  by the cluster's retention policy on `<date>` and are no longer available"
  message.

#### Notes

- The default of `forever` ensures this is a strictly opt-in behavioural change
  for existing deployments.
- Composes naturally with the workspace data retention feature: a workflow may
  have its logs pruned but its workspace kept, or vice versa.

michaelbuchar:
Closes reanahub/reana#964


helm/reana/templates/cronjobs.yaml
```diff
@@ -1,5 +1,6 @@
 {{- $sharedStorageFsGroup := .Values.shared_storage.fs_group | toString }}
 {{- $sharedStorageFsGroupChangePolicy := .Values.shared_storage.fs_group_change_policy | toString }}
+{{- $logRetentionPeriod := tpl (.Values.logs.retention_rules.maximum_period | toString) . }}
```
tiborsimko: **🤖 PR996-01: A null retention period enables a broken CronJob and crashes reana-server (LOW)**

The workspace setting protects its default, but the new log setting is passed
directly through `toString` and `tpl` in both the server deployment and CronJob
template:

```yaml
{{- $logRetentionPeriod := tpl (.Values.logs.retention_rules.maximum_period | toString) . }}
...
value: {{ tpl (.Values.logs.retention_rules.maximum_period | toString) . | quote }}
```

When an override explicitly sets `logs.retention_rules.maximum_period: null`,
Helm renders `"<nil>"`. Since that string differs from `"forever"`, the chart
also enables the logs-prune CronJob. Rendering and deployment succeed, but both
the server process and pruning command then fail while evaluating
`int("<nil>")`; the server enters a crash loop instead of preserving the safe
default.

Suggested fix — normalise a null or omitted value to `"forever"` before using
it in either template and apply the same rule to the deployment and CronJob.
Preserve numeric `0`, which the existing test treats as a valid zero-day policy;
a plain Helm `default` applied directly to the raw value would treat `0` as
empty as well. Extend the existing log-retention Helm rendering test with an
explicit null override.

michaelbuchar: Centralized retention-period normalization so explicit null safely becomes forever while zero and templated finite periods remain valid, with rendering coverage for all cases

helm/reana/templates/_helpers.tpl
```diff
@@ -15,6 +15,15 @@ naming spec: `my-reana-batch-yadage-3c640169-d3b7-41ad-9c09-392c903fc1d8`
 {{- end -}}
 {{- end -}}
 
+{{/* Normalise an explicitly null workflow log retention period safely. */}}
+{{- define "reana.logs_retention_period" -}}
+{{- if kindIs "invalid" .Values.logs.retention_rules.maximum_period -}}
+forever
+{{- else -}}
+{{- tpl (.Values.logs.retention_rules.maximum_period | toString) . -}}
```
tiborsimko, line 23: **🤖 PR996-02: An empty templated retention period deploys a crash-looping server (LOW)**

The new `reana.logs_retention_period` helper handles an explicitly null value, but it does not validate the value after `tpl` has rendered it. Both of these supported configurations therefore produce an empty value:

```yaml
maximum_period: ""
maximum_period: "{{ .Values.missing_log_retention_period }}"
```

The resulting manifests set `LOG_RETENTION_PERIOD` to `""` in both the server Deployment and an enabled `logs-prune` CronJob, because the empty value does not equal `forever`. Importing `reana_server.config` then evaluates `int("")`, so the server crash-loops and every pruning run fails. A misspelt or omitted template reference can consequently pass `helm template` and break the deployment only at pod start.

Suggested fix — evaluate and trim the rendered value inside the helper, then accept only `forever` or a non-negative integer and use Helm's `fail` function for an empty or otherwise invalid result. This should preserve `0`. It would also make the intended distinction explicit: a raw null value means "not configured" and maps to `forever`, whereas an empty rendered template indicates a configuration error and fails rendering. Please extend the Helm rendering test with both an explicit empty string and a templated value whose referenced key is absent.

helm/reana/templates/cronjobs.yaml
tiborsimko: **🤖 PR996-01 follow-up (2026-08-30): verified fixed at a494100**

The new `reana.logs_retention_period` helper detects an explicitly null value with `kindIs "invalid"` and normalises it to `forever` for both the server Deployment and the CronJob condition. This prevents the previous `<nil>` value from enabling a broken CronJob or reaching `int()` at server import.

The implementation preserves the supported numeric `0` value, and the rendering test now covers the null case. No residual concern remains for the explicitly null input raised in this finding. A separate finding, PR996-02, covers empty values produced after template rendering.

PR #1002, opened 2026-08-26
feat(reana-dev): add Go client installation and selection

### LINKED ISSUE #1001: reana-dev: support the Go client
Amend `reana-dev` to install `reana-client-go` alongside the Python client and optionally use it for client operations via `--client go`, keeping Python as the default.

This will facilitate local development and parity testing as the Go client approaches full feature parity with the Python client.

tiborsimko:
Install the Python client stack and `reana-client-go` side by side in the active virtual environment, using the running interpreter for `pip` and the Go client's Makefile for executable installation. Warn and skip Go installation when Go or make is unavailable; Python installation still proceeds. Go removal similarly warns and skips if make is unavailable.

Add `--client python|go` selection to `run-example` and `run-ci`, keeping Python as the default and propagating the choice to nested runs. Explicit Go CI runs validate build tools before cluster operations, then rebuild and smoke-test the Go executable. An installed Go client can run examples without its build tools.

Cover installation, removal, missing tools, Go build and smoke-test failures, client selection and early CI validation. All 76 unit tests pass, alongside Black, Flake8, pydocstyle and commitlint checks.

Closes #1001


tiborsimko:
Note: Integration testing of examples with `tests:` depends on [reanahub/reana-client-go#189](https://github.com/reanahub/reana-client-go/pull/189).

reana/reana_dev/client.py
```diff
@@ -46,22 +58,89 @@ def client_install():  # noqa: D301
...
+    Python components are installed in a single pip invocation so that
+    pip can resolve version constraints from all local source
+    directories together, avoiding conflicts when local branches
+    have different dependency pins than published PyPI versions. The Go
+    client is built into the current virtual environment's scripts directory.
+    """
+    scripts_dir = get_scripts_dir()
+    _ensure_client_components_checked_out()
+    _require_executable("go", "install")
+    _require_executable("make", "install")
```
CameronMcClymont: Non-blocking: `client-install` requires both `go` and `make` before doing anything, and `run_ci` always calls `client-install`. A developer with no Go toolchain loses `reana-dev run-ci`, even for Python-only work. The code already has `is_component_go_package()` and gating these two `_require_executable` calls on it would keep the Python-only path working. Same for the new `make` requirement in `client-uninstall`.

A fix would look something like:

```
if any(is_component_go_package(component) for component in REPO_LIST_CLIENT):
    _require_executable("go", "install")
    _require_executable("make", "install")
```

CameronMcClymont (dismissed):
On small non-blocker worth considering, otherwise good to go 🦭

reana/reana_dev/client.py
tiborsimko: Thanks! Addressed in `37a0fee`, squashed into the original commit.

Python installation now proceeds when Go build tools are missing, with a warning that Go installation was skipped. Python removal similarly proceeds without `make`, with a warning that any installed Go binary remains. Checking for a Go package alone would still require the tools in a normal checkout containing `reana-client-go`.

An explicit `run-ci --client go` checks Go and `make` before cluster operations, so an older installed binary cannot bypass rebuilding. `run-example --client go` can use an installed binary without build tools. I also simplified client selection to `--client`, keeping Python as the default and dropping `REANA_DEV_CLIENT`.

CameronMcClymont (approved):
Thanks for addressing, LGTM 🦭

PR #1003, opened 2026-08-27
Group support

### LINKED ISSUE #999: Group persistence, synchronization, and authorization policy
OIDC/JWT edge authentication (reana#976 and its companion PRs) validates
individual user identity only. There is no group concept anywhere in the
stack: no group membership model, no synchronization of group data from
the identity provider, and no authorization policy beyond the single
required role checked at login.

Scope:
- a group/membership model in reana-db
- synchronizing group membership from the identity provider (ID/access
  token claims, or a directory lookup)
- a policy layer that authorizes REANA operations based on group
  membership, not just the current single-role check

Sohn123:
Closes: reanahub/reana#999


helm/reana/templates/reana-server.yaml
```diff
@@ -188,39 +241,195 @@ spec:
...
+          - name: REANA_AUTH_REQUIRED_ROLE
+            value: {{ required "keycloak.required_role must be set when bundled development Keycloak is enabled" .Values.keycloak.required_role | quote }}
+          - name: REANA_AUTH_CLIENT_ID
+            value: {{ required "keycloak.cli_client_id must be set when bundled development Keycloak is enabled" .Values.keycloak.cli_client_id | quote }}
+          - name: REANA_AUTH_WEB_CLIENT_ID
+            value: {{ required "keycloak.web_client_id must be set when bundled development Keycloak is enabled" .Values.keycloak.web_client_id | quote }}
+          - name: REANA_AUTH_BFF_ENABLED
+            value: "true"
+          - name: REANA_GROUP_BACKENDS
+            value: '[{"type":"keycloak","provider":"keycloak","groups_claim":"groups", "server_url":"http://{{ include "reana.prefix" . }}-keycloak:8080{{ .Values.keycloak.relative_path }}","realm":"{{ .Values.keycloak.realm }}","client_id":"{{ .Values.keycloak.server_client_id | default "reana-server" }}", "client_secret_env":"REANA_AUTH_WEB_CLIENT_SECRET"}]'
```
tiborsimko, line 279: **🤖 PR1003-02: Bundled Keycloak backend uses the wrong configurable client ID (MEDIUM)**

The rendered group backend pairs `REANA_AUTH_WEB_CLIENT_SECRET` with:

```yaml
client_id: {{ .Values.keycloak.server_client_id | default "reana-server" }}
```

`keycloak.server_client_id` is not defined in `values.yaml`. The realm client, service account, assigned roles, and mounted secret are all based on `keycloak.web_client_id`. The defaults happen to share the value `reana-server`, hiding the mismatch until an operator customizes the web client ID.

With `keycloak.web_client_id=custom-web`, the chart still requests a client-credentials token as `reana-server` using `custom-web`'s secret, so group synchronization and search fail.

Suggested fix — render `.Values.keycloak.web_client_id` as the group backend client ID, or introduce a real separate service-account client with its own declared value and secret. Extend the Helm test to assert that a customized client ID is reflected in `REANA_GROUP_BACKENDS`.

helm/configurations/values-escape.yaml
```diff
@@ -0,0 +1,57 @@
...
+  openidConfigUrl: ""
+  clientId: ""
+  webClientId: ""
+  bffEnabled: true
+  scopes: "openid profile email offline_access"
+  rolesClaim: reana_roles
+  requiredRole: "reana:user"
+  groupBackends: []
+  indigoIamGroupBackend:
+    enabled: true
```
tiborsimko, line 48: **🤖 PR1003-03: The ESCAPE authentication profile no longer renders as documented (MEDIUM)**

`values-escape.yaml` now enables the INDIGO IAM group backend while leaving its `clientId` and group-backend secret empty. The invocation documented at the top of the file supplies only the existing web client values and secret, so Helm fails first on the missing group secret and then on the missing group backend client ID.

The existing Helm test avoids the failure by explicitly disabling `auth.indigoIamGroupBackend`, which means it no longer exercises the shipped profile's defaults. A developer following the documented local IAM recipe cannot deploy it successfully.

Suggested fix — either disable the group backend by default and document the opt-in values, or update the profile to reuse the seeded client consistently and document every required override. Test the exact command advertised in the profile header without compensating overrides.

tiborsimko (commented):
**🤖 PR1003-01: No group-membership refresh CronJob is deployed (HIGH)**

The new administrative command says it is run periodically by a `group-membership-refresh` CronJob, and group authorization rejects snapshots older than `REANA_GROUP_MEMBERSHIP_MAX_AGE`, which defaults to 24 hours. The Helm chart adds no such CronJob; `templates/cronjobs.yaml` still contains only the system-status and resource-quota jobs.

Users who authenticate mainly through bearer tokens do not receive recurring login synchronization. Their group-derived access therefore disappears by default after one day, despite successful authentication, and no scheduled component attempts to refresh it. The chart also provides no alertable signal that the required freshness mechanism is absent.

This is an integration omission: the server command and database freshness gate are introduced by the group PRs, while this Helm PR does not deploy the job they require.

Suggested fix — add a configurable CronJob using the reana-server image, database configuration, and group backend secrets. Schedule it well inside the maximum age, expose an enable switch and deadline, and ensure claim-only backends have a separately defined refresh strategy rather than being reported as successfully refreshed.

helm/reana/templates/reana-keycloak.yaml
```diff
@@ -0,0 +1,463 @@
...
+              /opt/keycloak/bin/kcadm.sh update "clients/${web_client_uuid}" \
+                --config "${config}" \
+                -r "${KEYCLOAK_REALM}" \
+                -s serviceAccountsEnabled=true
+              /opt/keycloak/bin/kcadm.sh add-roles \
+                --config "${config}" \
+                -r "${KEYCLOAK_REALM}" \
+                --uusername "service-account-${KEYCLOAK_WEB_CLIENT_ID}" \
+                --cclientid realm-management \
+                --rolename query-groups \
```
tiborsimko, line 413: **🤖 PR1003-04: Browser and directory service-account credentials are coupled (MEDIUM)**

The bundled Keycloak configuration enables a service account on the same confidential client used for the browser BFF authorization-code flow, then grants its service account `query-groups` and `view-users`. The same `REANA_AUTH_WEB_CLIENT_SECRET` consequently authorizes both browser token exchange and realm-directory reads.

`view-users` may be required by the Keycloak endpoint used to enumerate a user's groups, so the issue is not simply an unnecessary role. The architectural problem is credential coupling: compromise or accidental disclosure of the browser client secret now has a materially larger directory-reading blast radius. The CERN profile similarly defaults the group API client to the web client and secret.

Suggested fix — create a separate non-browser service-account client, disable its standard flow, grant only the roles required by the group backend, and store its secret independently. Require external-provider profiles to opt into reusing a web client rather than making that the default.

helm/reana/templates/reana-server.yaml
```diff
@@ -188,39 +241,195 @@ spec:
...
+          {{- $_ := set $cernGroupBackend "max_groups" . }}
+          {{- end }}
+          {{- with .Values.auth.cernGroupBackend.httpTimeout }}
+          {{- $_ := set $cernGroupBackend "http_timeout" . }}
+          {{- end }}
+          {{- $groupBackends = append $groupBackends $cernGroupBackend }}
+          {{- end }}
+          {{- if .Values.auth.indigoIamGroupBackend.enabled }}
+          {{- $indigoIamGroupBackend := dict "type" "indigo_iam" "provider" (.Values.auth.indigoIamGroupBackend.provider | default "indigo_iam") "client_id" (required "auth.indigoIamGroupBackend.clientId must be set when auth.indigoIamGroupBackend.enabled=true." .Values.auth.indigoIamGroupBackend.clientId) "client_secret_env" (.Values.auth.indigoIamGroupBackend.clientSecretEnv | default "REANA_GROUP_BACKEND_INDIGO_IAM_CLIENT_SECRET") }}
+          {{- if or $.Values.auth.caBundle $.Values.auth.caBundlePath }}
```
tiborsimko, line 316: **🤖 PR1003-05: IAM CA and TLS settings are rendered from undefined chart values (MEDIUM)**

The INDIGO backend template reads `auth.caBundlePath`, `auth.caBundleMountPath`, and `auth.tlsVerify`. None is declared in `values.yaml`; the chart exposes only `auth.caBundle` for issuer requests.

When `auth.caBundle` is set, the template guard runs but assigns `ca_bundle` from the two missing path values, producing an empty value. `tls_verify` is likewise rendered from an undefined setting. Even after runtime certificate verification is enabled, the chart cannot express a valid custom CA path or an explicit verified-TLS default for the IAM backend.

Suggested fix — define a clear group-backend TLS contract in `values.yaml`, reuse the mounted `auth.caBundle` path where appropriate, and default `tlsVerify` to true. Add a rendering test that sets a custom CA and asserts the exact `ca_bundle` and `tls_verify` values passed through `REANA_GROUP_BACKENDS`.

helm/reana/templates/reana-keycloak.yaml
```diff
@@ -0,0 +1,463 @@
...
+                  --noquotes
+              )
+              if [ -z "${web_client_uuid}" ]; then
+                echo "Reconciled Keycloak web client was not found" >&2
+                exit 1
+              fi
+              /opt/keycloak/bin/kcadm.sh update "clients/${web_client_uuid}" \
+                --config "${config}" \
+                -r "${KEYCLOAK_REALM}" \
+                -s serviceAccountsEnabled=false
```
tiborsimko, line 403: **🤖 PR1003-06: Keycloak reconciliation recreates the service-account user every time (LOW)**

The reconciliation hook unconditionally sets `serviceAccountsEnabled=false` and then `true` on the web client to recreate its service-account user before assigning roles. This runs on every reconciliation, not only when the user is missing.

The steady-state operation therefore deletes and recreates the principal and temporarily removes its role assignments during every deploy or hook rerun. Group synchronization and search can fail in that window, and any additional state attached to the service-account user is discarded.

Suggested fix — query for the expected service-account user and toggle the setting only when recovery is necessary. Make the normal reconciliation path idempotent and test a second run against an already configured realm.

helm/reana/templates/auth-secrets.yaml
```diff
@@ -0,0 +1,20 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "reana.prefix" . }}-auth-secrets
+  namespace: {{ .Release.Namespace }}
+  annotations:
+    "helm.sh/resource-policy": keep
+type: Opaque
+data:
+  {{- if or .Values.keycloak.enabled .Values.auth.bffEnabled (and .Values.auth.cernGroupBackend.enabled (eq (.Values.auth.cernGroupBackend.clientSecretEnv | default "REANA_GROUP_BACKEND_CERN_CLIENT_SECRET") "REANA_AUTH_WEB_CLIENT_SECRET")) }}
```
tiborsimko, line 11: **🤖 PR1003-07: Group backend secrets can be required without being consumed (LOW)**

Enabling a group backend widens the `auth-secrets.yaml` outer guard. With `bffEnabled=true` but no issuer, the Secret template now requires `REANA_AUTH_WEB_CLIENT_SECRET` even though the reana-server authentication environment block is skipped entirely; the resulting diagnostic asks for an unused BFF secret instead of identifying the absent issuer.

A second mismatch occurs when the CERN backend names `REANA_AUTH_WEB_CLIENT_SECRET` but BFF authentication is disabled. The Secret key is rendered, yet the container mounts that key only when Keycloak or BFF is enabled; the CERN-specific mount is deliberately skipped for the web-secret name. The backend then sees no credential and silently remains unavailable.

Suggested fix — derive Secret creation and environment mounts from one normalized list of enabled backend credential references. Validate that an external group backend has a configured issuer, mount every referenced key regardless of BFF state, and make error messages name the component that actually requires each secret.

PR #1013, opened 2026-09-17
feat(reana-dev): adopt saved client connection settings

### LINKED ISSUE #792: client: save connection settings and harmonise error messages
### Current behaviour

The clients remember the server after login, but TLS settings still depend on environment variables. Error messages differ between Python and Go. For example, connecting to a local server with a self-signed certificate produces:

```console
$ env | grep REANA
REANA_SERVER_URL=https://localhost:30443

$ reana-client-go ping
==> ERROR: could not complete token refresh: Post "https://localhost:30443/keycloak/realms/reana/protocol/openid-connect/token": tls: failed to verify certificate: x509: certificate signed by unknown authority

$ reana-client ping
==> ERROR: Could not refresh authentication credentials. Please try again.
```

### Expected behaviour

Both clients should support configuring and saving connection settings at login:

```sh
reana-client login --server https://localhost:30443 --no-tls-verify
reana-client ping
```

Subsequent commands should reuse the saved settings, replacing `REANA_SERVER_URL` and `REANA_SERVER_TLS_VERIFY` that we could retire. (Making `REANA_SERVER_URL` not required would be also a good split from previous REANA 0.9 release series.)

Error messages should be consistent across both clients and give clear guidance for missing configuration, authentication failures and certificate problems.

tiborsimko:
Let `run-example` select a server with `--server` and request a TLS
bypass explicitly with `--no-tls-verify`. Probe the saved login before
starting authentication and confirm the intended destination before
running examples. Preserve the target and TLS choice in retry guidance.

Have `run-ci` forward its local server and bypass option directly to
`run-example`. Make `client-setup-environment` print a login command for
users instead of export statements, and stop evaluating its output.
External callers must also stop evaluating this output.

Use private temporary client settings files for e2e subprocesses while
retaining their browser-session token input. Update Helm post-install
instructions to use `login --server` and include the bypass when the
chart generates its own self-signed certificate.

Refresh developer and administrator login instructions, and distinguish
REANA 0.9 from 0.95 CWL authentication. Label the benchmark
prerequisites as legacy and record that its API and authentication
migration remains pending.

Closes reanahub/reana-client#792


reana/reana_dev/run.py
```diff
@@ -16,6 +16,7 @@
 import sys
 import time
 from typing import List
+from urllib.parse import urlsplit, urlunsplit
 
 import yaml
```
CameronMcClymont, line 22: The PR body says: "Make `client-setup-environment` print a login command for users instead of export statements, and stop evaluating its output.", but several docstrings still wrap it in `eval`. There are also a few other stale docstrings related to environment variables and the login process:
- Here in `run.py:389`
-`reana/reana_dev/cli.py:102`
- `scripts/create-admin-user.sh:162`
- `scripts/run_cwl_conformance_tests.py:165`
- `reana/reana_benchmark/cli.py:44`

CameronMcClymont (changes requested):
Just one comment regarding some stale docs.

reana/reana_dev/run.py
tiborsimko: Addressed all five sites in `1377c63`, with the fixes squashed into the feature commit. Replaced obsolete setup instructions with saved login guidance and distinguished legacy and 0.95 CWL workflows. The benchmark instructions are labelled legacy: its migration also requires repairing a pre-existing import of a removed client API function, so that remains separate work.

CameronMcClymont (approved):
🦭

